PR #54, opened 2026-03-30
Smart route interpretation and route editing for flights (#53)

roznet:
- Add interpret-route API endpoint that resolves known waypoints from raw
  route text and returns interpreted/skipped tokens
- Flight creation uses smart parsing: unknown tokens are skipped with a
  confirmation popup showing what was interpreted vs skipped
- Add route editing to flight detail page (PATCH endpoint accepts waypoints)
- Origin and destination are enforced immutable on route edits
- Saved briefing packs are not affected (they store their own route copy)

https://claude.ai/code/session_01X4YZ7gEteevnff8Abawxim

### DIFF
--- a/src/weatherbrief/api/flights.py
+++ b/src/weatherbrief/api/flights.py
@@ -314,6 +314,94 @@ def compute_route_distance(
     )
 
 
+class InterpretRouteRequest(BaseModel):
+    """Request body for interpreting a route string with smart waypoint filtering."""
+
+    raw_route: str = Field(..., min_length=1, max_length=2000)
+
+
+class InterpretRouteResponse(BaseModel):
+    """Result of smart route interpretation."""
+
+    original_tokens: list[str] = []  # all tokens extracted from the input
+    interpreted: list[str] = []  # waypoints that resolved successfully
+    skipped: list[str] = []  # tokens that didn't resolve
+    waypoints: list[WaypointInfo] = []  # resolved waypoint details
+
+
+@router.post("/interpret-route", response_model=InterpretRouteResponse)
+def interpret_route(
+    req: InterpretRouteRequest,
+    request: Request,
+    user_id: str = Depends(current_user_id),
+):
+    """Interpret a raw route string, resolving known waypoints and skipping unknown tokens.
+
+    Accepts free-form route text (e.g. from a flight plan paste). Extracts tokens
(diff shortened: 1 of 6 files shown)


roznet:
## Code Review

### Bug: `interpret_route` duplicates logic from `resolve_waypoints`

`interpret_route` (flights.py L341-379) manually imports `_load_airport_model` (a private, `@lru_cache`-decorated function) and creates its own `RouteResolver` to resolve points one at a time, then calls `resolve_waypoints()` a second time to build `WaypointInfo` objects. This duplicates the resolution logic already in `airports.py` and reaches into a private API.

The existing `resolve_waypoints` already raises `KeyError` listing unknown codes. A cleaner approach: attempt resolution, catch the `KeyError` to extract the unknown tokens, then resolve only the known subset. This avoids importing `_load_airport_model` and `RouteResolver` directly, keeps the resolution logic in one place, and respects the module boundary.

**Severity: Medium** — works today but couples the endpoint to an internal implementation detail; changes to `_load_airport_model` caching or `RouteResolver` API would silently break this endpoint.

### Bug: Dead code in `flight-detail-ui.ts`

```ts
const routeDisplay = flight.waypoints.length > 2
  ? flight.waypoints.join(' → ')
  : flight.waypoints.join(' → ');
```

Both branches produce the same output. This ternary is dead logic — should be a plain `flight.waypoints.join(' → ')`.

**Severity: Low** — no runtime impact, but clutters the code and reads as an incomplete thought.

### Issue: `route_name` reconstruction on edit may diverge from creation logic

When updating waypoints (flights.py ~L640):
```python
row.route_name = "_".join(w.lower() for w in new_waypoints)
```

If the original `route_name` was set by a different formula during flight creation (e.g., including profile or date), this will silently change the naming convention. Worth verifying the creation path uses the same formula, or extracting a shared helper.

### Minor: `import json as _json` inside function body

The `import json as _json` (flights.py ~L639) inside the route-change block is unusual. `json` is a stdlib module already used elsewhere in the file — a top-level import is cleaner and avoids the underscore alias.

---

No meteorological/weather-analysis code affected. No issues with the TypeScript API adapter or the confirmation popup component — `escapeHtml` is used correctly for user input, and the modal cleanup (event listener removal, backdrop removal) looks correct.

PR #57, opened 2026-04-01
Add design doc for METAR/TAF accuracy verification system

roznet:
Standalone observation archive decoupled from flights/users,
three-way verification (METAR vs NWP models vs TAF), collection
loop during flight windows, CLI for backfill/export.

https://claude.ai/code/session_018ywPXymaJF73dPYfBy3Jv1

### DIFF
--- a/alembic/versions/029_verification_tables.py
+++ b/alembic/versions/029_verification_tables.py
@@ -67,7 +67,7 @@ def upgrade() -> None:
         sa.Column("id", sa.Integer, primary_key=True, autoincrement=True),
         sa.Column(
             "observation_id", sa.Integer,
-            sa.ForeignKey("verification_observations.id"), nullable=False,
+            sa.ForeignKey("verification_observations.id", ondelete="CASCADE"), nullable=False,
         ),
         sa.Column("icao", sa.String(4), nullable=False),
         sa.Column("observation_time", sa.DateTime(timezone=True), nullable=False),
@@ -106,7 +106,7 @@ def upgrade() -> None:
         sa.Column("id", sa.Integer, primary_key=True, autoincrement=True),
         sa.Column(
             "observation_id", sa.Integer,
-            sa.ForeignKey("verification_observations.id"), nullable=False,
+            sa.ForeignKey("verification_observations.id", ondelete="CASCADE"), nullable=False,
         ),
         sa.Column("icao", sa.String(4), nullable=False),
         sa.Column("observation_time", sa.DateTime(timezone=True), nullable=False),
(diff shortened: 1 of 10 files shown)


roznet:
## Code Review — PR #57: METAR/TAF Accuracy Verification System

Design doc + implementation (migration, ORM models, collection, scoring, CLI). Reviewed against design doc intent and existing codebase patterns.

### Bug: FK violation in hourly dedup will crash collection loop

`_should_skip_for_hourly_dedup()` in `tasks/verification.py` deletes an existing `VerificationObservationRow` when a newer METAR is closer to top-of-hour. But `FlightVerificationMapRow.observation_id` has a FK to `verification_observations.id` with **no `ondelete` clause** (defaults to RESTRICT). If a previous cycle already linked map rows to that observation, the delete will raise `IntegrityError` and crash the collection cycle.

**Fix**: Either add `ondelete="CASCADE"` on the `FlightVerificationMapRow.observation_id` FK (and matching in the migration), or have `_should_skip_for_hourly_dedup` delete linked map rows first before deleting the observation.

`scoring.py:VerificationScoreRow.observation_id` and `TafVerificationScoreRow.observation_id` have the same issue — no `ondelete` clause, so scoring rows would also block observation deletion.

### Bug: ORM models missing UniqueConstraints — dev environment won't enforce dedup

The migration creates `UNIQUE` constraints on all verification tables (`uq_verif_obs_icao_time`, `uq_verif_scores_key`, `uq_taf_verif_key`, `uq_fvm_flight_icao_obs`), but the SQLAlchemy model classes only declare `Index` entries in `__table_args__`, not `UniqueConstraint`.

In dev mode, `init_shared_db()` uses `Base.metadata.create_all()` which creates tables from the ORM definitions — **without the unique constraints**. This means the dedup logic (`_should_skip_for_hourly_dedup`, exact dedup checks) would silently insert duplicates in development, masking bugs that would only surface in production (where alembic migrations run).

**Fix**: Add matching `UniqueConstraint` entries to each model's `__table_args__`.

### Design concern: INDEX.md entry for partially-implemented system

Per prior feedback, `designs/INDEX.md` is for MCP discovery of current architecture. This PR adds real code so the entry is defensible, but the system is phased — Phase 3 (API endpoints, dashboard) doesn't exist yet, and the design doc describes it extensively. Consider whether the INDEX.md entry should note "Phase 1-2 implemented" to avoid misleading MCP consumers about available exports.

### No issues found in:

- Migration 029 structure, downgrade path, FK relationships to existing `flights` table
- Scoring logic — correctly reuses `reconcile_ceiling`, `classify_flight_category`, `compute_wind_advisory` from advisory pipeline (verified signatures match)
- Circular wind direction delta, precipitation phenomena set, convection risk thresholds
- Scheduler integration pattern (asyncio.to_thread for blocking IO)
- CLI structure and error handling
- Collection dedup flow (Phase A→B→C→E) and batch optimization

PR #58, opened 2026-04-01
Add PIREP submission & aircraft registry

roznet:
## Summary

- **User aircraft registry**: ICAO type lookup, CRUD API, flight integration, settings UI, privacy-gated tail/nickname
- **PIREP Milestone 1**: submit (single + batch with offline sync), query with filters (bounds, airport, hazard, altitude, time), web viewer with list + Leaflet map, iOS reporting card
- **Progressive rollout**: PIREPs tab gated behind `pirep_can_view` user preference; aircraft dropdown hidden when none configured with Settings link
- **Dev login**: `/auth/dev-token` endpoint + simulator Dev Login button
- **UI**: ellipsis menu replaces logout button, create flight form split into two rows

## Migrations

- `030_user_aircraft` — `user_aircraft` table + `aircraft_id` FK on flights
- `031_pireps_and_device_tokens` — `pireps` + `device_tokens` tables (idempotent guards)

## Test plan

- [x] 34 new tests covering PIREP submit (single + batch), ownership validation, permission gating, query filters, storage layer, aircraft CRUD
- [ ] Manual: create aircraft in Settings, verify it appears in flight creation dropdown
- [ ] Manual: submit PIREP from iOS, verify it appears in web viewer
- [ ] Manual: verify PIREPs tab hidden for users without `pirep_can_view`
- [ ] Manual: verify batch sync deduplication with `client_uuid`


roznet:
## Code Review — PR #58: PIREP submission & aircraft registry

### Issues

**1. Bug: `flight_id` PIREP query returns nothing when PIREPs lack `pack_id`**
`src/weatherbrief/storage/pireps.py:111-115`

```python
if flight_id is not None:
    pack_ids = (
        select(BriefingPackRow.id)
        .where(BriefingPackRow.flight_id == flight_id)
    )
    stmt = stmt.where(PirepRow.pack_id.in_(pack_ids))
```

`pack_id` is nullable and optional on submission (`SubmitPirepRequest.pack_id: int | None = None`). PIREPs submitted without a `pack_id` will never match this `.in_()` filter. The iOS `fetchPireps(flightId:)` and the web briefing PIREPs tab both use this query path — they'll show empty results for any PIREP submitted without linking to a specific briefing pack.

This is especially likely for inflight PIREPs from the iOS offline queue, where the user may not know or care about the pack ID.

Either:
- Make `pack_id` required at submission time for inflight PIREPs, or
- Add a `flight_id` FK directly to `PirepRow`, or
- Fall back to a spatial+temporal corridor query when `pack_id` is NULL

**2. Unintentional behavior change: flight plan EET duration rounding**
`src/weatherbrief/api/flights.py:503-504`

```python
duration_hours = float(math.ceil(fpl.eet_minutes / 60))
```

The old code was `round(fpl.eet_minutes / 60, 1)` which preserved fractional hours (70 min → 1.2h). The new code rounds *up* to the next whole hour (70 min → 2.0h). This means a 1h10m flight plan now produces a 2-hour flight duration, which affects the weather data time window, auto-refresh scheduling, and the flight tracking window. Unless this rounding-up was intentional (and if so, it seems unrelated to the PIREP/aircraft feature), this should be reverted to the original behavior.

**3. `UpdateAircraftRequest` can't clear nullable integer fields**
`src/weatherbrief/api/aircraft.py:178-184`

```python
updates = {k: v for k, v in req.model_dump().items() if v is not None}
```

For optional integer fields like `cruise_speed_kt` and `ceiling_ft`, once set, they can't be cleared back to `None` via the update endpoint because `None` values are filtered out. The `tail_number` and `nickname` fields have special-case handling (empty string → `None`), but `cruise_speed_kt` and `ceiling_ft` do not. Consider accepting `0` as a sentinel for "clear", or use `model_dump(exclude_unset=True)` to distinguish "not provided" from "explicitly set to null".

### Notes (not blocking)

- The `can_view_pireps` / `can_publish_pireps` functions each independently parse `app_prefs_json` from the DB. When both are called in the same request (e.g. if future code checks both), that's two JSON parses of the same row. Minor, but could be consolidated into a single helper.
- The retention exemption for packs linked to PIREPs (`retention.py:101-102`) is clean — matches the design doc's "linked packs exempt from retention cleanup" requirement.
- PIREP anonymization on account deletion (`app.py:80-82`) correctly preserves observation data while removing identity — good.

Overall this is a well-structured PR with solid test coverage (34 new tests), proper permission gating, rate limiting, and privacy controls. The flight_id query gap (issue #1) is the main concern since it affects the core PIREP viewing experience.

PR #81, opened 2026-04-22
feat(routes): filter waypoints that resolve far off the route

roznet:
## Summary
- Add a detour-based filter that drops middle waypoints whose resolved coordinates sit too far off the departure→destination route. Fixes ambiguous navaid lookups like `ABB` (Europe vs US) picking a wildly wrong candidate.
- `resolve_waypoints` now returns `(waypoints, rejected)` — all call sites updated.
- `create_flight` / `update_flight` return 422 naming the off-route waypoint(s).
- `interpret_route` surfaces rejects via the existing `skipped` list, which the frontend popup already shows for user confirmation — the normal UX path.
- `compute_route_distance` and `_build_route_config` log a warning when a rejection happens, since in the gated flow they shouldn't see any.

## UX per call site
| Site | On rejection |
| --- | --- |
| `create_flight` / `update_flight` | 422 with waypoint names — hard failure (backstop for non-UI callers) |
| `interpret_route` | goes to `skipped` → existing frontend popup asks user to confirm |
| `compute_route_distance` | dropped silently, warning logged |
| `_build_route_config` (packs) | dropped silently, warning logged with flight id |
| `_resolve_airport` (pireps), `is_known_waypoint` | unaffected (no route context) |

## Notes
- The detour gate lives in `euro_aip.RouteResolver`; this branch reads `route.rejected_waypoints` via `getattr` with an empty-list fallback, so it ships independently of a `euro_aip` release. Once a `euro_aip` version with the filter is pinned, rejections will start flowing through.
- The initial commit included a progressive per-token check inside `interpret_route` that couldn't fire (the new token was always the destination of the test route, and the gate only runs on middle waypoints). Second commit removes it — O(n) instead of O(n²), same observable behavior.

## Test plan
- [x] `tests/test_airports.py` — updated for new return shape, `rejected == []` on old euro_aip
- [x] `tests/test_packs.py`, `tests/test_flights_storage.py`, `tests/test_route_points.py` — all pass
- [ ] Manual: paste a route with a ambiguous-navaid-in-wrong-region and confirm the popup flow
- [ ] Manual: create via MCP with a deliberately off-route waypoint and confirm 422 message
- [ ] Watch for the new warnings in production logs after deploy — indicates grandfathered flights or threshold drift

### DIFF
--- a/src/weatherbrief/airports.py
+++ b/src/weatherbrief/airports.py
@@ -46,7 +46,9 @@ def is_known_waypoint(code: str, db_path: str) -> bool:
     return resolver.resolve_point(code.upper()) is not None
 
 
-def resolve_waypoints(codes: list[str], db_path: str) -> list[Waypoint]:
+def resolve_waypoints(
+    codes: list[str], db_path: str
+) -> tuple[list[Waypoint], list[str]]:
     """Resolve waypoint codes to Waypoints using euro_aip RouteResolver.
 
     Accepts ICAO airport codes (4 letters), navaid codes (2-3 letters),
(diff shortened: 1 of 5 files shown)


roznet:
Re: the move_flight review — there is no `move_flight` endpoint or `POST /{flight_id}/move` route in the codebase. `grep -n "def move\|/move" src/weatherbrief/api/*.py` returns nothing, and line 741 is inside `update_flight` (a waypoint-format validation error string, unrelated to `resolve_waypoints`).

All `resolve_waypoints` call sites after this PR:
- `create_flight:269` — 422 on rejects ✓
- `update_flight:749` — 422 on rejects ✓
- `compute_route_distance:410` — logs warning, returns truncated route ✓
- `interpret_route:501` — final resolve moves rejects into `skipped` ✓
- `packs.py:_build_route_config:380` — logs warning, briefing runs on survivors ✓
- `pireps.py:_resolve_airport:268` — single code, no middles to reject ✓

Skipping this change — the bug report appears to be a hallucination.

PR #122, opened 2026-05-07
feat(maps): airport profile panel via right-click on forecast markers (#121)

### LINKED ISSUE #121: Airport detail view on forecast map: Skew-T + time cross-section via right-click
## Goal

Let pilots inspect vertical weather conditions at any airport on the forecast map without creating a flight briefing. Right-clicking an airport marker opens a detail panel with the **same Skew-T and cross-section visualizations** the briefing page provides, scoped to a small time window around the currently selected map hour.

The motivation is pedagogical as well as operational: the cross-section is intuitive at a glance for any pilot, while the Skew-T offers more depth for those trained to read it. Showing both makes the page useful across experience levels.

## User flow

1. User is on `/maps.html`, looking at the forecast tab with day/hour/model selected.
2. User **right-clicks** an airport marker (left-click keeps current tooltip behavior).
3. A **right-side panel** opens showing, for that airport, at the selected (day, hour):
   - Time cross-section spanning the selected hour and the next 3 hours
   - Skew-T at the selected hour
4. Panel has its own model selector (independent of the map's). Closing returns to map-only.
5. Changing the map's day/hour refreshes the panel; changing the map's metric does not.

## High-level approach

**Reuse, not rebuild.** The existing `CrossSectionRenderer`, `RouteGraphRenderer`, and `SkewTRenderer` already accept generic per-point analysis data. The cross-section's X-axis is monotonic distance today, but each `VizPoint` already carries time — for an airport the spatial extent collapses to one point and the time extent becomes the X-axis. Build a **frontend adapter** that turns `(icao, hours[])` into a synthetic `VizRouteData` (flat terrain at airport elevation, all points share the airport's lat/lon, X = hours). Update the cross-section X-axis tick formatter to optionally render times instead of nautical miles. No changes to layer renderers, tooltip registry, or themes.

## Backend

Add a new **SSE endpoint** (e.g. `GET /maps/airport-profile?icao=&model=&start_hour=&window_h=3`) that streams progressive phases:

| Phase | Source | Latency |
|-------|--------|---------|
| `surface` | `airport_forecast_snapshots` cache | ~0ms |
| `levels` | Open-Meteo per-airport pressure levels | 1–2s |
| `enriched` | local GRIB index (ECMWF/ICON-EU/GFS) | up to ~10s |
| `derived` | `analyze_sounding` on combined data | ~50ms after enriched |

Reuse `OpenMeteoClient`, GRIB enrichment, and `analyze_sounding` from the existing per-route pipeline — the only change is "one point, multiple hours" instead of "multiple points, one base time." Cache aggressively per `(icao, model, init_time, hour)`.

Auth required, same as other map endpoints.

## Frontend

- **Right-click handler** in `web/ts/visualization/weather-map.ts` on each marker (Leaflet `contextmenu` event).
- **Layout**: split-screen with resizable divider — map left, detail panel right. Call `map.invalidateSize()` on resize. Highlight currently-displayed airport's marker.
- **Panel content**: cross-section on top, Skew-T below, both produced by the existing renderers via the adapter. Panel has model dropdown + close button.
- **Streaming UX**: paint axes/terrain immediately, then layers as SSE phases arrive (no full-page loading state).
- **Mobile (<900px)**: panel becomes a full-screen takeover with a back button.

## Decisions made (with rationale)

1. **3h time window, fixed** — selected hour + next 3 forward (`[h, h+1, h+2, h+3]`, 4 forecast hours). Keeps GRIB fetch to ~4 files per click; pilots can step the map's hour selector to see further.
2. **Reuse renderers via adapter, not new renderers** — the existing cross-section already encodes time implicitly; an airport view is the degenerate single-location case. ~50 lines of adapter + ~10 lines for axis formatter.
3. **SSE from day one, not a polling MVP** — multi-phase data fits SSE naturally and the codebase already has the SSE pattern (`/packs/.../refresh`).
4. **Initial layout: both views stacked, but treat this as provisional** — we expect crowding to be a real issue and plan to revisit after visual review. Build the panel so switching to a single-view-with-toggle is a small follow-up, not a rewrite. Don't invest in polish on the stacked layout until that decision is made.
5. **Right-click opens panel, left-click keeps tooltip** — preserves the existing quick-look behavior.
6. **Panel has independent model selector** — pilots want to compare models for one airport without churning the whole map.

## Open questions / revisit during implementation

- **Stacked-views crowding (likely)** — visual review after first working version will probably motivate switching to a segmented toggle inside the panel (Both / Cross-section / Skew-T, persisted to localStorage), or to showing only one view at a time by default. Structure the panel DOM and renderer wiring so either option is straightforward to adopt — don't bake "always show both" into the layout assumptions.
- **Sea / no-GRIB airports** — graceful degradation: show what data is available, label what isn't.
- **Pop-out to dedicated `/airports/{icao}` URL** — out of scope for this issue, but design URL params (`?model=&hour=`) so it's an easy follow-up.

## Out of scope

- Multi-model overlay on Skew-T (existing future-work item)
- Persisting per-airport pressure levels in DB (this issue does on-demand fetch + cache only)
- Airport detail as standalone route/page
- iOS app integration (tracked separately)

## References

- `designs/forecast-page.md` — current map architecture and cache layer
- `designs/visualization.md` — `CrossSectionRenderer`, `VizRouteData` shape
- `designs/skewt-canvas.md` — `SkewTRenderer` API and current sounding endpoint
- `web/ts/visualization/weather-map.ts:428` — marker creation site
- `src/weatherbrief/api/maps.py` — current map endpoints
- `src/weatherbrief/tasks/standalone_verification.py:251` — existing per-airport pressure-level fetch logic to reuse

roznet:
Right-clicking an airport marker on /maps.html opens a right-side panel
showing a vertical cross-section + Skew-T for that airport across the
selected hour and the next 3 forward hours, without creating a flight
briefing. Pilots can compare models (independent dropdown) and toggle
between Both / Cross-section / Skew-T views.

Closes #121.

## Backend

`GET /api/maps/airport-profile` streams phased SSE events:

| Phase | Source | Latency |
|-------|--------|---------|
| `meta` | static | ~0ms |
| `surface` | `airport_forecast_snapshots` cache | ~0ms |
| `levels` | Open-Meteo per-airport pressure-level fetch | 1–2s |
| `enriched` | local GRIB index (ECMWF/ICON-EU/GFS) via `enrich_forecasts` | up to ~10s |
| `derived` | `analyze_sounding` per hour on the (possibly-enriched) profile | ~50ms |
| `complete` / `error` | — | — |

GRIB enrichment reuses the existing per-flight `fetch.grib.enrich_forecasts()`
pipeline by wrapping our (icao, model, hours[]) into a synthetic single-point
RouteCrossSection — zero duplicate decode logic. A fast preflight skips the
phase when `ECMWF_GRIB_DIR` doesn't exist (dev environments). Pass
`?enrich=false` to bypass entirely.

## Frontend

- **`airport-profile-adapter.ts`**: SSE client + synthetic VizRouteData
  assembly. Hour offset is mapped onto `distanceNm` so the existing
  CrossSectionRenderer renders the time-axis view without a fork.
- **`airport-profile-panel.ts`**: panel UI (header, model select, view
  toggle, cross + skewt slots) with view mode persisted to localStorage.
- **`axes.ts`**: optional `timeAxisMode` on VizRouteData renders the
  X-axis as `HH:MMZ` ticks per point instead of nautical miles;
  waypoint labels suppressed in this mode.
- **`weather-map.ts`**: contextmenu handler on each forecast marker;
  `setHighlightedIcao()` ties the open panel to its source pin.
- **`maps.html`**: split-screen flex layout for the forecast tab;
  panel becomes a full-screen takeover at <900px viewport.
- **`maps-main.ts`**: panel open/close + reload on day/hour change
  (metric changes intentionally don't reload). Panel state
  (`fc.apt`, `fc.apModel`) round-trips via the URL share-link pattern.

## Tests

- 222 existing web tests pass; new `tests/test_airport_profile.py`
  exercises the hour-window helper.
- TypeScript typecheck clean.
- **Untested locally** — no Python runtime in this session. See [issue #121
  comment](https://github.com/roznet/flyfun-weather/issues/121#issuecomment-4396146395)
  for the on-device test plan.

### DIFF
--- a/src/weatherbrief/api/airport_profile.py
+++ b/src/weatherbrief/api/airport_profile.py
@@ -24,6 +24,8 @@ from __future__ import annotations
 import asyncio
 import json as json_mod
 import logging
+import threading
+from collections import defaultdict
 from collections.abc import AsyncGenerator
 from datetime import datetime, timedelta, timezone
 from pathlib import Path
@@ -42,6 +44,112 @@ router = APIRouter(prefix="/maps", tags=["maps"])
 
 _VALID_MODELS = ("gfs", "icon", "ecmwf")
 _DEFAULT_WINDOW_H = 3  # selected hour + 3 forward = 4 forecast hours
+
+
+# ---------------------------------------------------------------------------
+# Concurrency control
+#
+# Two layers of concurrency protection (see issue #121 PR #122 review):
+#
+# A) Per-user / global stream limiter — bounds the "user clicks 20 airports
+#    rapidly" DoS surface and the global thread-pool exhaustion surface.
+#    Lighter than packs._RefreshRegistry: no queue, no progress polling,
+#    no flight_id dedup — just an int counter.
+#
+# B) Same-run GRIB enrichment lock — two simultaneous panels for different
+#    airports on the same (model, run_init) currently both call
+#    enrich_forecasts(), which both download the same GFS / ICON GRIB
+#    file. Without this lock they race at the file-write level (see #123
+#    for the underlying cache-layer fix). With it, the second request
+#    waits for the first's enrichment to finish; by the time it acquires
+#    the lock, the disk cache is warm and the second request is fast.
+# ---------------------------------------------------------------------------
+
+_MAX_PER_USER = 3
+_MAX_GLOBAL = 20
+
+
+class _StreamLimiter:
+    """Bounded-counter request limiter for the airport-profile SSE.
+
+    Caps in-flight streams per-user (so one user can't monopolise the
+    thread pool with rapid right-clicks) and globally (so total in-flight
+    work is bounded). Counters are incremented synchronously at acquire
+    and decremented in the endpoint's `finally` block — `acquire()`
+    raises `HTTPException(429)` rather than blocking, so the client sees
+    the rejection immediately and can retry instead of hanging.
+    """
+
+    def __init__(self, max_per_user: int = _MAX_PER_USER, max_global: int = _MAX_GLOBAL) -> None:
+        self._lock = threading.Lock()
+        self._per_user: dict[str, int] = defaultdict(int)
+        self._global = 0
+        self._max_per_user = max_per_user
+        self._max_global = max_global
+
+    def acquire(self, user_id: str) -> None:
+        with self._lock:
+            if self._global >= self._max_global:
+                raise HTTPException(
+                    status_code=429,
+                    detail=f"Server busy ({self._global} airport profiles in flight). Try again shortly.",
+                )
+            if self._per_user[user_id] >= self._max_per_user:
+                raise HTTPException(
+                    status_code=429,
+                    detail=f"You already have {self._per_user[user_id]} airport profiles in flight (limit {self._max_per_user}).",
+                )
+            self._per_user[user_id] += 1
+            self._global += 1
+
+    def release(self, user_id: str) -> None:
+        with self._lock:
+            self._per_user[user_id] = max(0, self._per_user[user_id] - 1)
+            if self._per_user[user_id] == 0:
+                self._per_user.pop(user_id, None)
+            self._global = max(0, self._global - 1)
+
+    def snapshot(self) -> dict[str, int]:
+        """Return active counts for diagnostics / tests."""
+        with self._lock:
+            return {"global": self._global, "users": len(self._per_user)}
+
+
+_limiter = _StreamLimiter()
+
+
+class _GribRunDedup:
+    """Process-wide async-lock pool keyed by (model, init_time_unix).
+
+    Two simultaneous airport-profile streams on the same model+run
+    currently both invoke `enrich_forecasts`, which redownloads the same
+    GFS / ICON GRIB file (see #123 for the cache-layer race). This pool
+    serialises the enrichment phase so the second request sees a warm
+    cache and skips the download.
+
+    Locks are created lazily and never freed — there are at most ~6 active
+    GRIB runs at any moment (3 models × 2 cycles a day) so unbounded
+    growth is not a concern.
+    """
+
+    def __init__(self) -> None:
+        self._sync_lock = threading.Lock()
+        self._locks: dict[tuple[str, int], asyncio.Lock] = {}
+
+    def lock_for(self, model: str, run_init_unix: int | None) -> asyncio.Lock:
+        # Use a sentinel for "no init time available yet" so per-model
+        # bursts before the first GRIB resolves still serialise rather
+        # than all charging at the cold cache.
+        key = (model, run_init_unix if run_init_unix is not None else -1)
+        with self._sync_lock:
+            lock = self._locks.get(key)
+            if lock is None:
+                lock = asyncio.Lock()
+                self._locks[key] = lock
+            return lock
+
+
+_grib_dedup = _GribRunDedup()
 
 
 def _iso_utc(dt: datetime) -> str:
@@ -403,6 +511,12 @@ async def get_airport_profile(
     # SSE error mid-stream).
     surface = _surface_from_cache(db, icao, model, hours)
 
+    # Acquire the per-user / global concurrency slot before opening the
+    # stream. Released in the generator's finally block. Done synchronously
+    # (raises 429) so the client doesn't waste a round-trip on an SSE
+    # connection that would never produce a complete event.
+    _limiter.acquire(_user_id)
+
     async def event_generator() -> AsyncGenerator[str, None]:
         def _event(event_type: str, payload: dict) -> str:
             return f"event: {event_type}\ndata: {json_mod.dumps(payload, default=str)}\n\n"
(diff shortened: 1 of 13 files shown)


roznet:
Pushed `5aa44ef` addressing the review.

**Critical**
- ✅ `asyncio.get_running_loop()` — Py 3.12-correct.

**Important**
- ✅ Skew-T levels-never-found — root cause was `meta.hours[i]` carrying `+00:00` while `_hourly_to_dict` left Open-Meteo's naive datetimes unmarked. Centralised through a new `_iso_utc()` helper used by every time-key emission. Good catch — that bug would've made the Skew-T silently empty even with everything else working.
- ✅ GRIB default — kept `enrich=True` (the user explicitly asked for the GRIB phase to ship in this PR, not as a follow-up). To address the dev-environment concern, added a fast preflight that skips the heavy call when `ECMWF_GRIB_DIR` doesn't exist on disk. Production deployments are unaffected; dev gets a free no-op. PR description updated to no longer claim GRIB is deferred.
- ✅ `setHighlightedIcao` — store `{marker, weight, color}` per ICAO so restore is exact for both consensus-mode (weight=2 + agreement border) and per-model (weight=1) markers.
- ✅ Panel ICAO + model now round-trip via URL state (`fc.apt`, `fc.apModel`) following the forecast-page share-link pattern. Hydration reopens the panel on page load.

**Minor**
- ✅ `_fetch_pressure_levels` log line includes ICAO + lat + lon.
- ✅ Dropped redundant `hasattr(h.time, "tzinfo")` (always True for datetime).
- ✅ Removed the `request: Request = None  # type: ignore` workaround — the endpoint never accessed it directly, the dependencies inject their own.
- ✅ `_data_dir(...) -> Path` annotation added.
- ✅ Added a shallow `SoundingPayload` interface — the inner zone arrays stay `any[]` (server contracts vary too much to mirror without duplicating models), but the top-level keys are now type-checked so a dropped field would break compilation.

Three checks before merge that still need real-environment verification:
1. ECMWF preflight: confirm a populated `ECMWF_GRIB_DIR` is detected (the check is `dir.exists() and any(dir.iterdir())` — should be tolerant of missing/empty).
2. URL round-trip: open panel → copy URL → paste in incognito → panel reopens to same airport + model.
3. The Skew-T fix: levels phase populates the Skew-T (this was the silent bug — please confirm visually).

---

roznet:
Pushed `174c13e`. Loop 1 of 3.

**Important**
- ✅ #1 `skipped` exception type — now `{"all": "exception"}` matches the `Record<string, string>` contract.
- ✅ #2 Stale docstring — refreshed to list all 5 phases (`meta`, `surface`, `levels`, `enriched`, `derived`) and dropped the "not yet implemented" note.
- ✅ #3 `close()` / `destroy()` split — `close()` now only signals the host; `destroy()` is the single teardown path. The "close partially tears down, then destroy does it again" inversion is gone.

**Minor**
- ✅ #4 Double `exists()` — `is_dir() and any(iterdir())`.
- ✅ #5 `AirportProfileSurfaceHour` contract drift — added `precipitation_mm`, `snowfall_cm`, `cloud_cover_low_pct`. Threaded the latter two into `synthesizeVizPoint` with surface fallbacks so the ~2s window between `levels` and `derived` no longer renders zeros for those scalars.

**Skipped**
- ⏭️ #6 Skew-T hardcoded to hour 0 — this is intentional, not an oversight. Issue #121 spec, step 3: *"A right-side panel opens showing… Skew-T at the selected hour"* (singular). The cross-section spans the 4-hour window per the issue, the Skew-T snapshots only the leading edge — pilots advance the map's hour selector to inspect later hours, same flow the issue calls out for "step the map's hour selector to see further". Adding an in-panel hour picker is a feature request, not a fix; I'd file it separately if you want it.

Three checks still pending real-environment verification (unchanged from the last summary): ECMWF preflight, URL round-trip, Skew-T fix.

---

roznet:
Pushed `a090c35`. Loop 2 of 3.

Aligned the SSE error handling with `/refresh/stream` and added the same class of test that PR #107 added after a UUID field silently killed every refresh stream.

**Backend**
- ✅ Per-phase try/except. Derived-phase failure now yields a structured `event: error` payload with `{type, phase, message}` and `return`s — clean handoff vs. an EventSource connection drop.
- ✅ Enrichment phase guarded too — `iterdir()` raising on a wonky FS won't kill the generator.
- ✅ Extracted `_airports_db` + `_data_dir` to `api/deps.py`, removed duplication in `maps.py` (review #1).

**Tests** (mirroring `test_api.py::TestRefreshStreamEncoder`):
- `test_derived_event_round_trips_through_sse_encoder` — builds a real `SoundingAnalysis` and runs the exact two-step encode the route does. A future UUID/datetime/Enum on `SoundingAnalysis` (or any nested model) will fail this test, not the production stream.
- `test_meta_event_is_json_safe` — locks in `+00:00` suffix on every time string so client `===` matching works across phases.
- Structural lint: every `json_mod.dumps()` must include `default=str`, every `.model_dump()` must use `mode="json"`. Brittle by design.
- `TestGribEnrichSkippedShape` — `skipped` is always a dict on every return path. The string-vs-dict bug can't come back silently.

**Frontend**
- Adapter parses structured backend errors (`MessageEvent.data` as JSON) and falls back to the raw event for connection drops. Panel shows `Error — <phase>: <message>`.

**Skipped (review #2's deeper question)**
On your "processing may not be as fast as you think" point — agree it's worth watching, but I don't think we need the `_RefreshRegistry` + `asyncio.Queue` + `progress_callback` machinery. The briefing pipeline needs that because:
- It coordinates a multi-stage backend that emits intra-stage progress
- It dedupes concurrent refreshes by `flight_id`
- It needs queue depth limits (max 5) and per-user caps (max 2) to prevent DoS
- The pipeline survives client disconnect

Our airport-profile is per-request, ephemeral, no cross-request state, no parallelism opportunity, no intra-phase progress to emit. Sequential `await loop.run_in_executor(...)` is the right shape. What we *did* need was the structured-error-event pattern, which is now in.

If real-world testing shows phases are slower than expected, the most likely upgrade isn't the registry — it's emitting a synthetic `progress` event between phases (e.g. `{"type": "progress", "phase": "levels", "elapsed_s": 1.4}`) so the UI can show a spinner. Cheaper than the full machinery and unblocks the same UX wins.

3 checks before merge still pending real-environment verification:
1. ECMWF preflight detects a populated `ECMWF_GRIB_DIR`
2. URL panel state round-trip
3. The Skew-T fix populates after the levels phase

---

roznet:
Pushed `5ea3d97`. Loop 3 of 3.

**A — Per-user / global stream limiter**
`_StreamLimiter` caps in-flight streams at 3 per user, 20 global. Acquired synchronously before `StreamingResponse` (so the client gets a fast 429, not an SSE that never completes), released in the generator's `finally` so client disconnect / normal completion / error returns all release. Defensive release on the synchronous path covers the rare "StreamingResponse construction raises" case so slots don't leak.

**B — Same-run GRIB enrichment dedup**
`_GribRunDedup` is a process-wide `dict[(model, init_unix), asyncio.Lock]`. Two simultaneous panels for different airports on the same model+run now serialise their enrichment phase — second request enters the lock with a warm cache instead of redownloading the same GFS/ICON file. Lock keys are never freed (≤6 active runs at a time, no growth concern).

**C — Filed as #123, not in this PR**
The deeper issue is that `fetch/grib/cache.py:put_cached()` does `path.write_bytes(data)` with no atomicity. Two threads racing on the same file path produce a partial-interleaved corrupt file. That race exists today and affects briefings too — fix belongs at the cache layer (`tempfile + os.replace`), not at the airport-profile endpoint. The dedup lock above is enough to keep our endpoint correct in the meantime.

**Tests**
- `TestStreamLimiter` — acquire/release symmetry, per-user 429, global 429, double-release tolerance.
- `TestGribRunDedup` — same-key returns same lock, different keys independent, async serialisation of concurrent acquirers.

**Loop budget exhausted (3 of 3 used).** I'll stop here unless you flag something specific. Real-environment verification still pending: ECMWF preflight, URL panel state round-trip, the Skew-T fix populating after levels, and now also the limiter behavior under bursty right-clicks.

---

roznet:
Pushed `20c31c0` — cleanup pass.

**Fixed**
- `_GribRunDedup` docstring + call-site comment now honestly describe the key as `(model, forecast_start_unix)` and explicitly call out what the lock does and doesn't cover. The "doesn't cover" case (cross-hour requests on the same GRIB run) is the one #123 ultimately fixes at the cache layer.
- `_surface_from_cache` normalises already-aware DB datetimes via `astimezone(timezone.utc)` — closes the narrow bug where a driver returning `timezone(timedelta(0))` would silently miss the dict lookup. Dead `by_hour.get(h.replace(tzinfo=None))` fallback removed.
- Hoisted `from weatherbrief.api.deps import …` to the top of the file (PEP 8 E402).
- ECMWF-preflight comment spells out the proxy assumption explicitly so a future maintainer understands why we don't scan ICON/GFS dirs.
- Dropped the unused `from datetime import timezone as _tz` alias inside `_grib_enrich_levels`.
- `-> Any` return annotation on `_fetch_pressure_levels` (kept lazy to match the module's import pattern).
- `AirportContextHandler` signature stripped to `(icao: string) => void` — the `lat`/`lon` advertised but never read. If a future caller wants to short-circuit the server-side `_resolve_airport_coords` lookup, extend the signature then.
- `AIRPORT_PROFILE_WINDOW_H = 3` constant in `maps-main.ts`, used at both `openAirportPanel` / `refreshAirportPanelOnHourChange` call sites. Mirror of the Python `_DEFAULT_WINDOW_H`.

**Pushed back on (not in this PR)**
- **`cloudCoverTotalPct` additive formula** (random-overlap formula would be more correct). The existing route-briefing cross-section uses the same formula at `data-extract.ts:153`. Fixing it here without fixing it there would create inconsistency between the briefing page and the airport-profile panel. Belongs in a separate PR that touches both call sites.
- **ECMWF-only preflight check** (broaden to ICON / GFS too). Doing it correctly means re-implementing each source's cache discovery; doing it cheaply (e.g. checking `data_dir`) is worse than the current proxy. Better to document the assumption explicitly (done above) and revisit if a partial setup ever ships.
- **Skew-T hour-picker.** Per #121 spec, intentional. Filed as a feature request would be the right move if the cross-section/Skew-T linking turns out to matter in real use.

Tests + typecheck + build green. Loop budget really exhausted now (4 of 3 used 🙃) — ready to merge from my side unless something critical surfaces. Real-environment verification list unchanged.

---

PR #135, opened 2026-05-08
feat(viz): orthogonal cloud source × style axes + new square style

roznet:
## Summary

- Replaces 4 hard-coded combined cloud layer files (`cloud-bands`, `nwp-cloud-bands`, `soft-cloud-bands`, `soft-nwp-cloud-bands`) with a single `cloudLayer({source, style})` factory. Adds two new layers as a third style: `square-cloud-bands`, `square-nwp-cloud-bands`.
- Splits the Cloud Method profile setting into independent **Cloud Source** (DD / NWP) and **Cloud Style** (Soft / Layer / Square) selects, with no DB migration — `cloud_method` stays the persistence key (`soft_nwp`, `square_dd`, …) and legacy values round-trip cleanly.
- Layers panel collapses the 6-checkbox cloud row into a compound control: per-source checkboxes + a shared style dropdown. Sources can be enabled independently. When the active model lacks NWP cloud data (e.g. ECMWF on a flight without GRIB enrichment) the NWP checkbox renders disabled with the existing _Not available for this model_ tooltip — no silent fallback.

## Why now

Adding a third cloud style (square) without the refactor would have multiplied four leaf files into six, plus four new entries in `PREFERRED_METHOD_LAYER`, GRAMET preset, compact overrides, and i18n. The factory consolidates the latent (source × style) decomposition that already lived inside `renderMatchedZones` callbacks. Future styles (true per-pressure-level cube once we plumb levels into `VizPoint`) become one new entry in the style map.

## Square style notes

Square uses the same continuous color function as the hatched Layer style — `cloudFillFromDD(dd, coverage)` for DD, `nwpCloudFill(coverPct)` for NWP — but draws a solid rectangle without the hatch overlay. Same DD/cover% hue+alpha modulation, cleaner cells. Bands span the half-distance to neighbouring points via the existing `drawColumnBand` and `renderMatchedZones`.

## Test plan

- [ ] Open `/briefing.html?flight=…` cross-section, toggle the NWP and DD cloud sources independently with each of the three styles selected — verify all 6 layer combinations render.
- [ ] Switch to a model with no NWP cloud data (ECMWF on flights without GRIB enrichment) — NWP checkbox should disable + tooltip; switching to GFS/ICON re-enables it.
- [ ] Open `/settings.html`, edit a profile — Cloud Source + Cloud Style should round-trip across save/reload. Existing `cloud_method = soft_nwp / soft_dd / nwp / dd` parses cleanly into the two selects.
- [ ] GRAMET preset still flips to Soft NWP only.
- [ ] Hover over a Square cloud cell — tooltip line shows DD/CC values (parity with Soft and Layer).

## Out of scope

- iOS port — currently has its own four-layer cloud setup mirroring the pre-refactor web. A parallel iOS PR can mirror this factory once the web shape stabilises.
- True per-pressure-level cube — needs `VizPoint.levels` plumbing from the backend; tracked separately.


roznet:
Note on a small silent behavior change flagged in review:

`coverageToPct('FEW')` now returns 15 (was 35 by default-fallthrough in the deleted `nwp-cloud-bands.ts`). Two NWP styles (Layer hatched + Square) now render FEW cells at lower opacity than before. This is intentional and meteorologically correct — FEW (1–2 oktas) should be visibly thinner than SCT (3–4 oktas) — but logging it here so anyone bisecting a "FEW clouds look lighter now" report knows to land on this PR rather than chasing into theme/colour code.

The Soft NWP path was unaffected (it never used `coverageToPct`; coverage drives a `COVERAGE_ALPHA` lookup which already had a FEW=0.15 entry).

Issue #167, opened 2026-05-20
Tiered refresh gating + reusable real-time refresh task


roznet:
## Motivation

The manual refresh button gates the full pipeline on a single min-rule: `_build_data_status` (`src/weatherbrief/api/packs.py:467`) returns `fresh=False` if *any* selected model has a newer covering run, and the button endpoints (`refresh_briefing` `packs.py:1189`, `refresh_briefing_stream` `packs.py:1324`) skip the pipeline when fresh. So a user can trigger a full (token + compute + disk) refresh on every single model tick — wasteful when one model update won't move a multi-day-out picture.

Goal: make the gate **progressively stricter the further out the flight is**, and make a D-0 press **always at least cheap-useful**.

## Part A — Seam: reusable real-time refresh task

Factor the body of `refresh_observations` (`packs.py:1627`) into a reusable `run_realtime_refresh(pack_dir, db_path, ...) -> RouteObservations` in `src/weatherbrief/tasks/route_weather.py`; the endpoint becomes a thin wrapper.

This is the cheap path: re-fetch METAR/TAF + comparison from stored forecasts, patch `briefing.json`. **No** model fetch, **no** GRIB, **no** LLM.

## Part B — Tiered gate

A pure `decide_refresh(pack, flight, days_out) -> RefreshDecision`, fed by the per-model states `_build_data_status` already computes:

- `n_eligible` = selected models (`pack.model_sources`) whose latest available run covers the flight horizon
- `n_updated` = eligible models with a newer-than-pack covering run (today's `stale_models`)
- `needed = min(threshold[days_out], n_eligible)`, where **threshold = {>=2: 3, 1: 2, 0: 1}**
- mode:
  - `full` if `n_updated >= needed`
  - `realtime` if `days_out == 0` and `n_updated < needed` (and observations are refreshable)
  - `none` otherwise
- `reason` string + `eta_useful` = the `(needed - n_updated)`-th soonest marker `next_expected` among not-yet-updated eligible models

The `min(..., n_eligible)` cap handles small selections: 2 models selected -> D-2 and D-1 both need both; 1 model -> every press works regardless of lead.

Note: the default selection is exactly the 3 main models (ECMWF/GFS/ICON, the only `default=True` in `fetch/variables.py`), so "count >= 3" == "all three mains" for nearly every user. A user who opts into a 4th model makes D-2 mean "3 of 4" — acceptable per design; one-line switch to "all selected" if we later want strictness.

### Wiring

[...]

PR #170, opened 2026-05-21
feat(sigmet): route SIGMET integration mirroring METAR/TAF (D-0 real-time)

roznet:
Surface SIGMETs affecting the route corridor as D-0 real-time info, built
on the #167 run_realtime_refresh seam. SIGMETs are FIR-polygon area hazards
(no model-comparison analog), so they live in a sibling RouteSigmets model
that retains polygon coords, the enroute span and the vertical band to
enable a future cross-section/map overlay without re-fetching.

- model: SigmetAlongRoute / RouteSigmets / RealtimeRefreshResult; route_sigmets on ForecastSnapshot
- fetch: run_route_sigmets via euro_aip RouteSigmetService (surface->cruise+buffer band, departure-day window)
- pipeline: step 3.5 fetch + sigmet usage flags; sigmet_corridor_nm option (default 50nm)
- realtime refresh: run_realtime_refresh refreshes SIGMETs alongside METAR/TAF and returns both
- digest/report/web: text + LLM-prompt sections, HTML report table, web SIGMET table with detail popup, i18n (en/fr/de/es)
- tests: helpers, fetch mapping, snapshot roundtrip, digest, realtime refresh

euro_aip SIGMET support is on the rzflight GitHub HEAD (not yet in the
published PyPI euro-aip 0.9.2) — develop against the git version.


roznet:
## Review feedback addressed (commit 9332a20)

Thanks for the review. Resolution below.

### Fixed
1. **Listener accumulation** — `renderRouteSigmets` now uses `el.onclick =` instead of `addEventListener`, so repeated store-subscription renders replace rather than stack the click handler.
2. **`top_ft = 0` → `FL000`** — the Jinja report now renders `SFC` for `top_ft <= 0`, matching the TS `sigmetLevel` helper.
3. **`count` desync** — `count`, `hazards` and `has_severe` are now `@computed_field` properties derived from `sigmets`; the manual bookkeeping in `run_route_sigmets` is removed. They remain serialized, so the templates/frontend are unaffected.
5. **Duplicated `_lvl`** — extracted `format_flight_level()` into `digest/format_utils.py`; both `text.py` and `prompt_builder.py` use it.
8. **Integration-test mock** — `test_refresh_gate_realtime_seam_integration` now returns a real `RouteSigmets` and asserts the `route_sigmets` patch path (on disk and in the response), exercising the `if new_sigmets is not None` branch.
9. **Window collapse** — `_departure_day_window` now logs a `debug` line when the window collapses (late-evening fetch past the departure-day end).

### Not changed — premise didn't hold on inspection
- **4 (front-end SIGMET propagation):** the realtime gate path already propagates SIGMETs. The streaming refresh runs `loadPacks()` → `fetchSnapshot()` on completion, which re-reads the `briefing.json` the seam patched (so `route_sigmets` is present), and the standalone `observations/refresh` button applies both `observations` and `sigmets` directly in `briefing-store.refreshObservations`. The non-streaming `/packs/refresh` (`refreshBriefing`) is not used by the web client. No change needed.
- **6 (`toLocaleTimeString`):** `renderRouteObservations` uses the same `toLocaleTimeString(...)` pattern (not `toISOString`), so the SIGMET code is already consistent with its sibling; changing only this path would introduce an inconsistency. Left as-is.
- **7 (`route_icaos` may contain `None`):** `Waypoint.icao` is a required `str` (non-optional) in `models/analysis.py`, so the list never contains `None`, and `run_route_weather` uses the identical line. euro_aip's `RouteSigmetService` also skips unresolvable ICAOs gracefully. No change needed.

---

roznet:
## Round-2 review addressed (commit 4118d3e)

All five items were valid and are fixed:

1. **Store blanking SIGMETs on failure** — `refreshObservations` now only overwrites `route_sigmets` when the server returns them (`result.sigmets != null`); a `null` (server-side SIGMET fetch failed) keeps the existing in-memory SIGMETs instead of silently hiding the section.
2. **Impossible `None` mock** — `test_uses_stored_corridor_nm` now uses `side_effect=Exception(...)` to exercise the real failure path that yields `new_sigmets = None`.
3. **Swallowed ImportError** — the `run_route_sigmets` import is hoisted outside the `try` in pipeline step 3.5, so a genuine `ImportError` surfaces rather than being logged as a SIGMET fetch failure.
4. **Round vs truncate** — TS `sigmetLevel` now uses `Math.floor(ft / 100)` to match Python's `ft // 100`, so the same SIGMET shows the same flight level in the web UI and the PDF report.
5. **Escaped `&mdash;`** — the empty raw-text fallback now uses `{% if ... %}{% else %}&mdash;{% endif %}` for both the SIGMET cell and the pre-existing METAR cell (line 211), so it renders an em-dash rather than the literal `&mdash;`.

Tests: full `test_route_weather` + `test_api` + `test_pipeline` + `test_report` green; TS typecheck + briefing build clean; 258 web unit tests pass.

---

roznet:
## Round-3 review — disposition (commit f972336)

**Fixed (security):**
- **XSS in `renderSigmetPopup`** — `s.direction` is now escaped via `escapeHtml` before injection into the `<pre>`, matching the table-row renderer. (Low practical risk since the data is the AWC `dir` field, but it's a real inconsistency and a one-line fix.)

**Deferred to main (small / non-blocking, by maintainer's call):**
- **`observations/refresh` response shape** — informational; no external callers today, already noted in `architecture.md`. No action.
- **`fetchLabel` shows local time labelled `Z`** — genuine but cosmetic, and *pre-existing* in `renderRouteObservations` too; should be fixed in both places together (`timeZone: 'UTC'`) in a follow-up rather than diverging here.
- **`_departure_day_window` test could assert `win_from == win_to`** — test-tightening nicety; the behavior is already correct and logged.

These three are minor and non-blocking, so they'll be addressed on `main` post-merge rather than spinning another review cycle.

---

Issue #171, opened 2026-05-21
Priority-aware GRIB decode dispatcher with fault-tolerant rescheduling


roznet:
## Problem

All GRIB decode work in the uvicorn process funnels through a single global `ProcessPoolExecutor` (`src/weatherbrief/fetch/grib/__init__.py:229`, default 2 workers). Three workloads share it with **no ordering guarantee** — submission is plain `pool.submit()`, strictly FIFO with races between concurrent submitter threads:

| Workload | Path | Priority |
|---|---|---|
| User refresh (`triggered_by="user"`) | `api/packs.py:1467` `_refresh_executor` → `execute_briefing` → `enrich_forecasts` | **INTERACTIVE (highest)** |
| Interactive airport profile | `api/airport_profile.py:478` `enrich_forecasts` | **INTERACTIVE** |
| Scheduler auto-refresh (`triggered_by="scheduler"`) | `scheduler.py:118` `to_thread` → `execute_briefing` | **SCHEDULED** |
| Standalone forecast cycle | `scheduler.py:736/811` → `run_standalone_cycle` → `_dispatch_decode` (`standalone_verification.py:663/694`) | **BACKGROUND (lowest)** |
| GRIB precache loop | `scheduler.py:1090` | **BACKGROUND** |

The standalone cycle fans out across its own `ThreadPoolExecutor` (`standalone_verification.py:413`) and can saturate both workers, so a user refresh routinely queues behind background work. `concurrent.futures` executors have **no priority support** (FIFO `_call_queue` fed in submission order), so priority must be an admission layer in front of the pool.

## Goal

A process-wide **`PriorityDecodeDispatcher`** that owns submission to the existing decode pool:

1. Orders pending jobs by priority (INTERACTIVE > SCHEDULED > BACKGROUND), FIFO within a level.
2. **Non-preemptive**: never kills a running job; priority only decides which *pending* job takes the next freed worker slot.
3. Bounds in-flight jobs to the worker count (reuses `GRIB_DECODE_WORKERS`).
4. Survives pool faults by **keeping completed work, rescheduling interrupted work, and dead-lettering poison jobs** (Recovery, below).

## Invariants

[...]

PR #172, opened 2026-05-22
feat(grib): priority-aware, fault-tolerant decode dispatcher (#171)

roznet:
Add a process-wide PriorityDecodeDispatcher in front of the GRIB decode
pool. It orders pending jobs by priority (INTERACTIVE > SCHEDULED >
BACKGROUND, FIFO within a level), bounds in-flight jobs to the worker
count, and on a pool fault keeps completed work, transparently reschedules
interrupted work, and dead-letters poison jobs (timeout victim / retry cap
/ sliding-window retry-rate budget). Each job now gets its own
GRIB_DECODE_TIMEOUT_S instead of one shared batch deadline.

Priority propagates via a _DECODE_PRIORITY ContextVar set at the entry
points: user refresh / airport profile = INTERACTIVE, auto-refresh =
SCHEDULED, standalone cycle = BACKGROUND. The dispatcher wraps and reuses
the existing pool lifecycle and hang-diagnostics rather than replacing
them; crash recovery runs off the executor-manager thread to avoid a
self-join wedge.

Bypass paths preserved: GRIB_DECODE_WORKERS=0 (in-process) and
GRIB_DECODE_PRIORITY_ENABLED=0 (legacy FIFO, rollback switch). App shutdown
drains the dispatcher so blocked callers fail fast. New tests in
test_grib_dispatcher.py (fake pool + injected workers, plus a real-pool
crash-recovery regression); test_grib_pool.py pinned to the legacy path.
Adds designs/grib-decode-dispatcher.md.

https://claude.ai/code/session_01Namje5Z84nNErYho5xf5eq

### DIFF
--- a/designs/INDEX.md
+++ b/designs/INDEX.md
@@ -75,7 +75,7 @@ GRIB2 enrichment engine: GFS S3 (CLWMR/ICMR/cloud diagnostics), ICON-EU DWD (QC/
 
 ### grib-decode-dispatcher [project]
 Priority-aware, fault-tolerant admission layer in front of the GRIB decode process pool. `DecodePriority` (INTERACTIVE/SCHEDULED/BACKGROUND, lower=higher) propagated via a `_DECODE_PRIORITY` ContextVar; `PriorityDecodeDispatcher` orders pending jobs by priority (FIFO within a level), bounds in-flight to the worker count, and on a pool fault keeps completed work, reschedules interrupted work (crash=jittered backoff, timeout=immediate), and dead-letters poison jobs (timeout victim / retry cap / retry-rate budget). Idempotency-only invariant. Bypass via `GRIB_DECODE_WORKERS=0` or `GRIB_DECODE_PRIORITY_ENABLED=0`.
-Key exports: `DecodePriority`, `PriorityDecodeDispatcher`, `enrich_forecasts(priority=...)`, `_dispatch_decode`, `_dispatch_decode_parallel`, `decode_dead_letter_counts`
+Key exports: `DecodePriority`, `PriorityDecodeDispatcher`, `set_decode_priority`, `enrich_forecasts(priority=...)`, `_dispatch_decode`, `_dispatch_decode_parallel`, `decode_dead_letter_counts`
 → Full doc: grib-decode-dispatcher.md
 
 ### multi-user-deployment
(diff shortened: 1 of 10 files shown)


roznet:
Thanks for the review — all five points addressed in 5cf78e9:

- **Important — `submit_one`/`submit_batch` drain race:** added a `_closed` guard inside the lock in both; a submit that loses the race to `drain()` now fails its caller future immediately with `DecodeDispatchError("dispatcher_shutdown")` instead of hanging. Added `test_submit_after_drain_fails_fast` to lock it in.
- **`_run_inline` swallowing `BaseException`:** now catches `Exception` (mirrored to the future) and re-raises `BaseException`, so `KeyboardInterrupt`/`SystemExit` propagate.
- **Unsynchronised `_DECODE_POOL` read in `_handle_fault`:** now read under `_DECODE_POOL_LOCK` for consistency with every other access.
- **Production code importing private `_DECODE_PRIORITY`:** added a public `set_decode_priority(p)` helper; `packs.py` and `scheduler.py` use it now.
- **Fragile `time.sleep` in the ordering test:** replaced both sleeps with a poll on the pending heap until both jobs are queued — ordering is decided by the heap, not submission timing, so it's now deterministic.

Full GRIB suite green (152 passed: pool + dispatcher + grib + cache).

---

roznet:
Round-2 review addressed in 72888f4:

- **Important #1 — floating-handle leak on drain-during-backoff:** real bug, fixed. A crash-retry handle mid-backoff lives only in a `threading.Timer`, so `drain()` (which walked only `_pending`+`_inflight`) missed it and the timer no-op'd on `_closed`. Now backed-off handles are tracked in `_delayed_handles`; `drain()` fails them, and both `_reenqueue` branches fail the caller with `dispatcher_shutdown` on `_closed` instead of dropping it. Added `test_drain_during_crash_backoff_releases_floating_handle`. (`_JobHandle` is now `eq=False` so it's hashable for the set.)
- **Important #2 — `_resolve_worker` under the lock:** moved worker resolution to submit time (stored on the handle), so no import/resolver work runs inside the critical section. Note: I don't think the described *deadlock* is actually reachable — decode workers are separate spawn **processes** (no shared import lock with the parent), `self._cond` is mutually exclusive so two threads can't hold it, and importing `decode_worker` never re-enters the dispatcher. But resolving off the lock is a genuine improvement, so done.
- **Minor #3 — `time.sleep(0.3)` in the slot-bound test:** replaced with a poll until both slots are occupied (bounded to `workers`, so it can't over-count).
- **Minor #4 — `_get_dispatcher` unsynchronised `_closed` read:** now checked under `_DISPATCHER_LOCK`.

Full GRIB suite green (153 passed).

---

Issue #192, opened 2026-05-30
Model-update-aware auto-refresh email timing (defer to ride imminent big model run)


roznet:
## Summary

Make the auto-refresh email scheduler **model-update-aware**: when a flight's regular refresh slot is scheduled to fire shortly *before* a fresh, horizon-extending model run lands, optionally defer the send so the briefing rides the newer run instead of being stale-at-birth. Strictly bounded, and **never** applied on the day of the flight (timeliness wins there).

This is a *when-within-the-day* refinement, not a *whether-to-refresh* change — the freshness gate (`decide_refresh`) already handles the latter.

## Today's behaviour

All in `src/weatherbrief/scheduler.py`. The timing decision is `_next_due_at` (scheduler.py:172):

```
next_due = min(next_regular, flight_start − 2h)
```

- `next_regular` = next occurrence of the user's preferred hour (`auto_refresh_hour`, default `departure_hour − 1`), one slot per calendar day, fired at `HH:00` UTC.
- `flight_start − 2h` = the preflight slot — a guaranteed final refresh near departure.

The scheduler is purely **clock-driven** with zero awareness of model run times.

## The problem: "just-missed"

The horizon-extending ECMWF runs (00Z/12Z, full 168h — see `fetch/freshness/registry.py:148`) land at **~06:40 and ~18:40 UTC** (`_ECMWF_OFFSET = 6h40m`). A refresh that fires at `HH:00` just before one of those deliveries builds the brief on a much older full run when a fresh one is minutes away. Example: a 06:00 UTC slot uses the previous 12Z run (~11 h old) when the 00Z run is 40 min out.

The opposite case — firing 30 min *before* a delivery — is exactly what we want to avoid, **except on the day of the flight**, where the pilot needs the brief in hand regardless of freshness.

## Production data — sizing the opportunity

(Queried prod MySQL, 2026-05-30. Sample is thin and mostly historical: 64 auto-on flight-rows ever, 30 users, only 3 currently future — treat as directional.)

- **78% (50/64)** use the default (`auto_refresh_hour` NULL → `departure_hour − 1`); 14 set an explicit hour.
- Effective refresh hour (UTC) clusters hard in the morning — **07–09 UTC = 39/64 (61%)**, peak 08:00 (25 = 39%):

  | Hour UTC | 4 | 5 | 6 | 7 | 8 | 9 | 10 | 11 | 12 | 13 | 16 | 18 | 21 |
[...]

PR #193, opened 2026-05-30
Defer auto-refresh emails for imminent model updates (issue #192)

roznet:
## Summary

Implements model-update-aware email timing so auto-refresh briefings ride imminent horizon-extending model runs instead of being stale-at-birth. When a regular auto-refresh slot is scheduled to fire shortly before a fresh ECMWF 00/12Z delivery lands (~06:40/18:40 UTC), the email is deferred a bounded amount to include the newer data.

The feature has two levers:
1. **Lever 1 (opt-in)**: Users with explicit `auto_refresh_hour` can enable "defer for model update" in account preferences
2. **Lever 2 (silent)**: Users with NULL `auto_refresh_hour` (default) automatically snap out of the pre-delivery dead-zone to ride fresh runs

Deferral is strictly bounded and never applied within 2 calendar days of the flight — timeliness wins on/near departure.

## Key Changes

**Scheduler logic** (`src/weatherbrief/scheduler.py`):
- Added `_defer_regular_for_model_update()` to defer regular slots when a full-horizon run lands within a 2-hour window after the slot
- Modified `_next_due_at()` to optionally apply deferral to the regular term (preflight is untouched)
- Added `_user_defers_for_model_update()` to read per-user opt-in preference with caching
- Updated `_find_due_flights()` to apply model-update-aware timing based on user preference and NULL-default status
- Added configuration constants for deferral bounds, margins, and day-of gates

**Registry enhancements** (`src/weatherbrief/fetch/freshness/registry.py`):
- Added `max_horizon()` to find the largest forecast horizon any cycle reaches (e.g., ECMWF 00/12Z = 168h vs 06/18Z = 90h)
- Added `next_full_horizon_run()` to identify the next full-horizon run after a given time, excluding medium-only cycles that would give shorter horizons

**Preferences API** (`src/weatherbrief/api/preferences.py`):
- Added `defer_email_for_model_update` boolean field to `PreferencesResponse` and `PreferencesUpdate`
- Added `load_defer_email_for_model_update()` to read the preference from the database
- Updated preference parsing and update logic to handle the new toggle

**Frontend** (`web/settings.html`, `web/ts/settings-main.ts`, `web/ts/adapters/preferences-adapter.ts`):
- Added checkbox UI for "Wait for an imminent model update before sending" in account settings
- Wired preference loading/saving to the new API field

**Tests**:
- `TestDeferRegularForModelUpdate`: 13 test cases covering deferral logic (window bounds, day-of gates, stale markers, slip handling)
- `TestNextDueAtModelUpdate`: 4 test cases for integration with `_next_due_at()` (NULL-default snap, opt-in toggle, preflight untouched)
- `TestMaxHorizon` / `TestNextFullHorizonRun`: 8 test cases for registry functions (cycle selection, delivery timing, uniform vs per-cycle horizons)
- `test_defer_email_for_model_update_round_trip`: Preference persistence test

## Implementation Details

- **Deferral window**: Only defers if a full-horizon run lands within 2 hours after the regular slot (slots earlier already ride fresh data)
- **Day-of gate**: Never defers within 2 calendar days of the flight — the preflight slot (2h before departure) owns this window
- **Slip handling**: Respects late-running model cycles up to a hard cap of 2h30m wait (mirrors forecast-fetch's bounded-wait pattern)
- **Marker validation**: Requires a healthy heartbeat (non-stale marker) to confirm an imminent delivery; missing/stale markers disable deferral
- **Cycle selection**: Uses `next_full_horizon_run()` to skip medium-only cycles (e.g., ECMWF 06/18Z

https://claude.ai/code/session_01BVKWn6kUvnytVkjoYmumJy

### DIFF
--- a/designs/freshness-markers.md
+++ b/designs/freshness-markers.md
@@ -170,7 +170,7 @@ When tuning registry offsets:
 
 ## References
 
-- Issue #108 (design + acceptance criteria); issue #167 (tiered refresh gate)
+- Issue #108 (design + acceptance criteria); issue #167 (tiered refresh gate); issue #192 (model-update-aware auto-refresh email timing — `scheduler._defer_regular_for_model_update` consumes the store + `registry.next_full_horizon_run`/`max_horizon`)
 - Existing helpers wrapped: `fetch/model_status.py`, `fetch/grib/{grib_fetch,icon_eu_fetch,ecmwf_watcher}.py`
 - Pack-side: `api/packs.py:_build_data_status`, `_backfill_sources`, `_finalize_refresh` (records `model_sources`), `_provider_label` (reads `registry.SOURCE_REGISTRY.provider_label`)
 - Tiered gate: `api/packs.py:decide_refresh` + `_refresh_threshold`/`_days_out_now`/`RefreshDecision`, `ModelStatus.covers_horizon`; wired into `refresh_briefing`, `refresh_briefing_stream`, `scheduler._auto_refresh_one`; realtime seam `tasks/route_weather.run_realtime_refresh` (see [metar-taf-route-weather.md](metar-taf-route-weather.md))
--- a/designs/multi-user-deployment.md
+++ b/designs/multi-user-deployment.md
@@ -342,6 +342,8 @@ Background scheduler (`scheduler.py`) polls every 10 minutes for flights with `a
 
 **Scheduling formula**: `next_due = min(next_regular, flight_start − 2h)` where `next_regular` is the next occurrence of the user's preferred hour (`auto_refresh_hour`, defaulting to `departure_time.hour − 1`) after the last refresh.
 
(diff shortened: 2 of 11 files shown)


roznet:
Reviewed and merged (rebased onto main). The implementation is correct and well-built — I verified the one subtle correctness question myself: `marker.is_stale(LOOP_INTERVAL)` does **not** false-disable deferral, because `_run_freshness_check_once` calls `store.mark_check()` on every 5-min tick (scheduler.py) so `last_check` stays within the 10-min threshold during normal operation. The staleness guard only trips if the freshness loop itself is dead — a genuine fail-safe. Tests green (111).

Addressed the last review's 4 findings directly on main in 1eee55a8:
- Added `defer_email_for_model_update` to the `load_service_toggles` no-row fallback dict (sync with `_parse_service_toggles`).
- `load_defer_email_for_model_update` now reuses `load_service_toggles` instead of duplicating db.get/json.loads/default (also removes the migration-skip divergence).
- Dropped the unreachable `else regular` guard in `_defer_regular_for_model_update`.
- Preference-read failure now logs at `warning`, not `debug`.

**My own review comment (one addressed, one deferred):**
- ✅ *Deferral was silent.* Added an `info` log when a regular slot is actually deferred, mirroring the `_wait_for_marker_freshness` pattern — important because this shifts send-time for the NULL-default majority and we'll want to confirm it firing in prod logs.
- ⏭️ *No kill-switch for the silent Lever-2 behaviour.* `apply_mu` is always-on for NULL-default users, with no env off-switch (cf. `DISABLE_STANDALONE_VERIFICATION`). I left this **unbuilt on purpose** — the feature already fails safe (stale/missing marker → no defer) and the deferral is capped at 2.5h and never within 2 days of the flight, so a misfire's blast radius is bearable. Worth adding only if the ECMWF delivery schedule changes and we see deferrals misbehave. Noted here so it's not lost.

Issue #249, opened 2026-06-13
Regulatory alternate assessment (FAA 91.169 + EASA Part-NCO): computed-requirement bands with Likely/Marginal/Unlikely confidence


roznet:
## Summary

Add a **regulatory "alternate required?" assessment** to the briefing, computed two ways — **FAA (14 CFR 91.169)** and **EASA Part-NCO (non-commercial GA)** — for the destination, and a **per-alternate qualification** for each weather-computed divert candidate we already surface.

**Ceiling and visibility are two independent criteria, each evaluated from real forecast data** (TAF or NWP). The forecast values are never proxied. What we lack is the **published plate minima** (DA/MDA and RVR/visibility) — we only know the approach *type*. The key design idea (this issue):

> Don't hide the missing minima behind a single conservative number. Instead **compute the actual requirement transparently** (e.g. EASA needs `ceiling ≥ DH + 200`), express the unknown plate value as a **plausible range** for the approach class, and report a **confidence band**: **Likely / Marginal / Unlikely**. "Likely" = the forecast clears even the worst-case plate minima; "Unlikely" = it fails even the best-case; "Marginal" = inside the band, genuinely undeterminable without the plate.

- **FAA** thresholds are **fixed regulatory numbers** (2000/3 trigger; 600-2 / 800-2 alternate) — no plate uncertainty, so FAA stays **binary** (Yes/No). The only FAA proxy is the precision-vs-non-precision *classification*, kept conservative (ILS-only = precision).
- **EASA** derives its thresholds from the (unknown) plate minima + margins, so EASA uses the **3-state band**.
- Where a hard yes/no is still forced (does *any* qualifying alternate exist? is an alternate required?), **Marginal counts conservatively** (marginal alternate → not counted as qualifying; marginal trigger → treat as required). Conservatism lives at the decision points; nuance lives in the display.

This is **planning-grade and advisory**, consistent with the attention-director framing of the briefing — never a go/no-go verdict.

## Scope (decided)

- **Alternates assessed = the existing weather-computed divert candidates** (`RouteAlternates` / `AlternateAirport`, issue #210). We are **not** adding a user-entered Field-16 alternate list here — possible follow-up.
- Two outputs:
  1. **Destination trigger** — "is an alternate required?" (FAA: Yes/No; EASA: No / Marginal / Required).
[...]

PR #250, opened 2026-06-13
Add regulatory alternate-requirement assessment (FAA 91.169 + EASA Part-NCO)

roznet:
Implements issue #249: a planning-grade advisory layer that answers "is a filed alternate required?" and "does each divert candidate meet alternate minima?" for both FAA (14 CFR 91.169) and EASA Part-NCO regimes.

## Summary

The feature computes regulatory alternate requirements transparently from real forecast data (TAF when available, NWP consensus fallback) and estimated plate-minima ranges per approach class. Rather than hide the missing published minima behind a single conservative number, it expresses the unknown as a confidence band:

- **FAA**: Binary verdicts (Yes/No) based on fixed regulatory thresholds (2000 ft / 3 SM trigger; 600-2 / 800-2 alternate minima)
- **EASA**: Three-state verdicts (Likely / Marginal / Unlikely) derived from estimated plate minima ± margins

## Key Changes

**Core logic** (`analysis/alternate_requirement.py`):
- Pure functions for verdict computation (no I/O, fully unit-testable)
- `ApproachProxy` dataclass with estimated plate-minima ranges per approach class (ILS, RNP/RNAV, VOR/NDB/LOC)
- `TrendView` / `CeilingVisWindow` for TAF trend aggregation and worst-case windowing over the ETA window
- `band_qualification()` / `band_trigger()` verdict primitives; `combine_qual()` / `combine_trigger()` for worst-of-criteria
- `compute_faa_trigger()` / `compute_easa_trigger()` for destination alternate requirement
- `compute_faa_qual()` / `compute_easa_qual()` for per-candidate qualification

**Wiring** (`tasks/alternate_requirement.py`):
- Post-pipeline step (after `run_alternates` + `run_route_weather`) that builds the destination ceiling/visibility window from TAF (via `euro_aip.briefing.weather`) or NWP fallback
- Looks up destination approach class from `nav.db`
- Computes `AlternateRequirement` (FAA + EASA triggers) and per-candidate `AlternateQual` verdicts
- Writes results to `snapshot.alternates.alternate_requirement`

**Models** (`models/alternate_requirement.py`):
- `BandVerdict` enum (LIKELY / MARGINAL / UNLIKELY)
- `TriggerVerdict` enum (NOT_REQUIRED / MARGINAL / REQUIRED)
- `CriterionAssessment` (one criterion vs its requirement band)
- `RegAlternateTrigger` (destination trigger for one regime)
- `AlternateQual` (per-candidate qualification for one regime)
- `AlternateRequirement` (full assessment: destination + per-candidate)

**UI** (`web/ts/managers/briefing-ui.ts`, `web/css/style.css`):
- `faaQualBadge()` / `easaQualBadge()` for per-candidate qualification display
- `altRequirementBanner()` for destination trigger display
- CSS classes for verdict styling (green=ok, amber=marginal, grey=fails)

**Integration**:
- Pipeline wiring in `pipeline.py` (post-step after alternates)
- Text digest output in `digest/text.py`
- Model exports in `models/__init__.py` and `models/alternates.py`
- TypeScript types in `web/ts/store/types.ts`

- **Conservative rules**: Missing ceiling in NWP = clear (good); missing visibility = fail. Unknown approach class degrades to non-precision (most demanding).
- **TAF windowing**: Samples at ETA±60/±30/0/+30/+60 min; TEMPO/PROB groups can only worsen the prevailing condition; tracks whether the binding value came from a temporary group.
- **Approach proxy tuning surface**: `APPROACH_CLASS_PROXY` dict maps `procedures.approach_type` strings to estimated DH/MDH and visibility ranges.

https://claude.ai/code/session_01Dwj1Ek17b5h7zdN2pjF7Fe

### DIFF
--- a/designs/INDEX.md
+++ b/designs/INDEX.md
@@ -73,6 +73,11 @@ Weather-based alternate airports (D-2 inward, gated by `compute_alternates` pref
 Key exports: `run_alternates`, `RouteAlternates`, `AlternateAirport`, `AlternateAxisPick`, `best_ceiling`, `flight_category`, `enrich_wind`, `consensus`, `compute_route_distances`
 → Full doc: alternates.md
 
+### alternate-requirement
+Regulatory "is a filed alternate required?" for the destination, computed two ways — FAA (14 CFR 91.169, binary) and EASA Part-NCO (Likely/Marginal/Unlikely band) — plus per-candidate alternate-minima qualification for each #210 divert candidate. Forecast ceiling/visibility are real (TAF at D-0, NWP consensus otherwise); the unknown plate minima are estimated as a per-approach-class range that sets the Marginal band width. Pure logic in `analysis/alternate_requirement.py`, wired as a pipeline post-step.
+Key exports: `run_alternate_requirement`, `compute_faa_trigger`, `compute_easa_trigger`, `compute_faa_qual`, `compute_easa_qual`, `build_window`, `APPROACH_CLASS_PROXY`, `AlternateRequirement`, `AlternateQual`, `BandVerdict`, `TriggerVerdict`
+→ Full doc: alternate-requirement.md
+
 ### time-alignment
 Time and spatial alignment in the data pipeline: aware-UTC datetime convention, per-hour GRIB enrichment across flight windows, spatial index consistency, hour-matching merge logic, old pack backward compatibility.
 Key exports: `compute_flight_window_hours`, `compute_icon_eu_flight_window_hours`, `_forecast_hour_to_utc`, `_merge_cloud_water_into_sections`
--- a/designs/alternate-requirement.md
+++ b/designs/alternate-requirement.md
@@ -1,0 +1,161 @@
+# Regulatory Alternate Requirement (FAA 91.169 + EASA Part-NCO)
+
+> For a briefing's destination, answer **"is a filed alternate required?"** two
+> ways — **FAA (14 CFR 91.169)** and **EASA Part-NCO** — and, for each
+> weather-computed divert candidate (#210), **"does it meet alternate minima?"**.
+> Planning-grade and advisory, never a go/no-go verdict.
+
+**Status: Shipped (#249).** Computed as a pipeline post-step whenever the
+alternates stage ran (D-2 inward + `compute_alternates` opt-in).
+
+## The core idea: transparent requirement + confidence band
+
+Ceiling and visibility are two independent criteria, each evaluated from **real
+forecast data** (TAF or NWP). The forecast values are never proxied. What we
+lack is the **published plate minima** (DA/MDA and RVR/visibility) — we only know
+the approach *type*.
+
+Rather than hide the missing minima behind one conservative number, we **compute
+the actual requirement** (e.g. EASA needs `ceiling ≥ DH + 200`), express the
+unknown plate value as a **plausible range** per approach class, and report a
+**confidence band**:
+
+- **Likely** — the forecast clears even the worst-case plate minima.
+- **Unlikely** — it fails even the best-case.
+- **Marginal** — inside the band; genuinely undeterminable without the plate.
+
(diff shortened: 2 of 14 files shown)


roznet:
Thanks for the review — pushed `4f36557` addressing it.

**Fixed (both Important + all Minor, batched into one push):**
- **`_is_temporary` base-group hazard** — `_trend_to_view` now zeroes out `probability` for the base group (`is_base=True`), so a stray `probability` attribute on the parent `WeatherReport` can no longer misclassify the prevailing line as temporary. Added a regression check: a base group with `probability=20` now keeps its real ceiling/visibility (not optimistically-clear / pessimistically-fail) and sets no TEMPO tag.
- **EASA trigger coverage** — added `TestEasaTrigger` (above-band → NOT_REQUIRED, inside-band → MARGINAL, below-band → REQUIRED, vis-forces-marginal, VFR-proxy collapse, no-forecast → REQUIRED) and the wiring smoke test now asserts `req.easa.status`.
- **Minors** — dropped the unreachable `else "nwp"` branch in `_build_trigger` (Pydantic `Literal` validation now keeps a broken invariant loud); annotated `nwp_ceiling`/`nwp_vis` as `float | None`; narrowed `CriterionAssessment.verdict` from `str` to a `Literal` of the five verdict values.

**Deferred:** nothing — all findings addressed.

Re: the meteorological-validity confirmation (thresholds, window, conservative bias) — appreciated, that matched intent.

---

roznet:
Pushed `58282aa` for round 2.

**Fixed:**
- **Month-rollover (Important)** — confirmed real. `_coerce_dt` now anchors validity markers to the ETA month (trying ±1 month, picking the calendar date nearest the ETA) instead of a fixed Jan-2000 epoch. A base valid from day 29 and an FM from day 01 now order correctly (FM supersedes base), so a worse FM can't be silently dropped → no false NOT_REQUIRED. Returns naive datetimes so the ordering key never mixes aware/naive with the `datetime.min` sentinel. Added a month-straddling supersession test (base 3000 ft / FM 800 ft at an ETA on day 01 → window resolves to 800 ft).
- **FAA marginal direction (Minor)** — digest *and* the TS banner now key off the safe direction: only an explicit `not_required` reads "Not required"; anything else (incl. a stray marginal) → Required.
- **TS `verdict` (Minor)** — narrowed `string` → the five-value `Literal` union, matching the Python model.
- **Base-stray-probability test (Minor)** — added the targeted regression test you suggested (the previous fix was verified locally but I hadn't committed the test; now it's in).

**Deliberately not changed:**
- **`nwp_window(None, None)` → both Required** — when *both* ceiling and visibility are absent we have no usable NWP data, so `has_forecast=False` + "no forecast available" + Required is the correct conservative reading. Treating the absent ceiling as "clear" there would mislabel a genuine data gap as known-clear; and the trigger status is Required either way (missing vis). The "None ceiling = clear" rule only applies when a forecast is actually present (ceiling absent *with* a visibility), which `nwp_window(1500, None)` etc. already handle correctly. Happy to revisit if you'd prefer the per-criterion ceiling shown as "no ceiling" in that degenerate case.

---

roznet:
Round-3 review addressed on `main` post-merge in `7ad96058`:

- **Important — TEMPO badge stale-persist**: `build_window` now derives `triggered_by_tempo` per criterion by comparing the window's overall worst vs. the worst *prevailing-only* line, instead of latching a per-instant flag. A later equally-bad prevailing instant clears the tag. (Values/verdicts were already correct — display label only.) Two regression tests added.
- **Minor** — `_build_destination_window` return type; `digest/text.py` direct `easa_map` lookup; `altRequirementBanner` labels the reason with its regime.

Deliberately kept: `nwp_window(None, None)` → source `none` (both axes absent = genuine no-data, verdict Required either way), and `_is_temporary`'s bare-PROB branch (defensive for a PROB line with empty trend_type, not dead). 53/53 tests pass; tsc clean.

PR #251, opened 2026-06-14
Add voluntary donation system with Stripe integration

roznet:
## Summary

Implements a complete voluntary donation system for Flyfun Weather, allowing users and anonymous visitors to support the service. Donations are framed around impact (coverage in user-months) rather than transaction value, and are web-only by design per App Store IAP rules. The system integrates with Stripe Checkout, records donations idempotently via webhook, and displays community/personal impact using margin-excluded cost economics.

## Key Changes

**Backend (Python)**
- New `donations.py` router with three endpoints:
  - `POST /donations/checkout` — creates a Stripe Checkout Session (supports anonymous donations)
  - `POST /donations/webhook` — idempotent webhook handler for `checkout.session.completed` and `charge.refunded` events
  - `GET /donations/me` — viewer's donation total + personal impact (auth required)
  - `GET /donations/summary` — public this-year community total + coverage framing (no auth)
- New `impact.py` module with pure donation impact math:
  - `ProgramEconomics` — margin-excluded run-cost denominators derived from cost reports
  - `DonationImpact` / `YearlyImpact` — frozen dataclasses for coverage calculations
  - Natural-language phrasing layer (`format_user_coverage`, `format_yearly_coverage`)
- Updated `credits.py` to remove retired `spending_limit` auto-reload logic and add `fx` block to cost responses
- Updated `preferences.py` to add `display_currency` field (ISO 4217 or "auto")
- Alembic migration 067 creates `donation_ledger` table and drops `users.spending_limit` column

**Frontend (TypeScript/HTML)**
- New `donate.html` page with:
  - Community impact card (this year's total + coverage phrasing)
  - Personal impact card (logged-in users only, hidden if no donations)
  - Currency selector + preset amount buttons
  - Stripe Checkout integration
- New `donate-thanks.html` and `donate-cancel.html` redirect pages
- New `donations-adapter.ts` with API client functions and `FxBlock` interface
- New `donate-main.ts` with form logic, impact rendering, and currency handling
- Updated `settings.html` to add display-currency picker
- Updated `settings-main.ts` to handle currency preference persistence

**Tests**
- `test_donations_api.py` — comprehensive tests for checkout, webhook (idempotency, fee handling), `/me`, `/summary` endpoints
- `test_impact.py` — pure impact math tests (economics derivation, coverage calculations, phrasing)
- `test_credits_changes.py` — validates removal of `spending_limit` and addition of `fx` block

- **Webhook is the source of truth**: donations are only recorded via the Stripe webhook, never from the client redirect. Idempotent on `provider_ref` to handle retries.
- **USD-canonical with FX display**: all amounts stored in USD; responses carry an `fx` block so the frontend renders the viewer's currency without additional API calls.
- **Margin-excluded impact**: impact framing uses `ProgramEconomics` derived from cost reports with margin intentionally dropped — donations offset the operator's real cost, not the buffer.
- **Anonymous donations allowed**: `viewer_id` is optional throughout; donations can be attributed to `None`.
- **Web-only by design**: no iOS app integration; the donate UI lives only in the web frontend per App Store IAP rules.
- **Idempotent cost recording**: `charge_briefing` no longer decrements a per-user balance; it simply appends the real USD cost to the shared ledger (the old `spending_limit` + `$5` auto-reload was write-only churn).
- **Stripe fee handling**: when available, `retrieve_net_ratio` is called to compute `net_usd` (amount after Stripe fees) for impact calculations.

## Dependencies

- Updated `flyfun-common` to `>=0.5.0` (provides shared payment plumbing: `create_checkout_session`, `verify_webhook_event`, `record_donation`, `retrieve_net_ratio`, `

https://claude.ai/code/session_012inf2KmXSsRtWV4cbfR9r1


roznet:
Thanks — addressed all of review round 2 in `9107ec6`.

**Important (both data-loss, fixed):**
1. **Unguarded `fx.to_usd()` in webhook** — now wrapped; on FX failure it records a 1:1 USD fallback (logged) instead of 500→retry→eventual loss. USD always converts exactly; the fallback only mis-rates a non-USD donation during a *total* outage with no cache, which beats dropping it.
2. **`recurring:true` accepted by the API** — now rejected with `422` at `/checkout` (not just hidden in the UI), so a direct POST can't open a subscription whose renewals the webhook ignores.

**Minor (batched into the same push):**
3. `fx: dict` → typed `FxBlock` pydantic model across all four cost/donation response models (real OpenAPI schema).
4. `field_validator` normalizes `PreferencesUpdate.display_currency`.
5. Hoisted the deferred `date` / `flyfun_common.fx` imports in `preferences.py` to module top.

Added tests for the two blockers (recurring-rejected, FX-failure-falls-back) — 36 passing locally. Nothing deferred from this round.

Note: CI will stay red until **flyfun-common 0.5.0** is published (the pin is `>=0.5.0`; PyPI only has 0.4.2) — that's the dependency-rollout step, not a code issue.

---

PR #260, opened 2026-06-16
Long-range early outlook digest + forecast-horizon booking gate

roznet:
## What

Two related changes for handling flights past the reliable forecast horizon.

### 1. Long-range early outlook digest (haiku)
Beyond the ~7-day ECMWF **GRIB** horizon we lose direct soundings + ICON; only the global models (ECMWF/GFS via Open-Meteo) reach the date. Running the full Sonnet briefing there is expensive and falsely precise. So:

- **Regime detection** — `is_long_range()` (days_out > ECMWF GRIB horizon, computed from the registry) switches to a trimmed, **model-agreement-led** context, a cheaper model (**haiku**), and a new `LongRangeDigest` with a soft scale: **TRENDING_SETTLED / MIXED_SIGNALS / TRENDING_UNSETTLED** — never GREEN/AMBER/RED.
- **Coarse context** — broad signals only; precise sounding indices (CAPE/CIN/K-index/shear…) dropped because they aren't skilful this far out and invite false precision. All-model agreement is promoted to the headline.
- **"More detail from \<date\>"** — the milestone (first full-horizon ECMWF GRIB run that reaches the flight) is computed in code from the registry and injected as a fact; the LLM phrases it (no LLM date math).
- **Text suppression** — DWD/NWS text is included only when a block *strictly covers* the flight day (new `strict` mode). DWD's Mittelfrist tops out at ~7 days, so at D-8/D-9 nothing is fed rather than a stale, non-covering narrative.
- **Persistence + UI** — `outlook`/`outlook_reason` persisted (`briefing_packs`, migration **068**), served via the pack API, rendered as a distinct dashed **"Early outlook"** badge (web + email, all 4 locales). Trend continuity survives the long↔short-range boundary.

### 2. Forecast-horizon booking gate
Creating **or moving** a flight past the dual-model horizon — `min(ECMWF, GFS)` Open-Meteo availability = **9 days** (computed, not hardcoded) — is rejected (422) with *"choose a departure date on or before \<date\>"*. The web date picker caps its `max`. MCP `create_flight` inherits the gate (routes through the same endpoint).

## Validation
- **haiku vs sonnet** on 12 real D-8 prod packs: **91% outlook agreement**; haiku's prose is tighter and arguably better-suited (sonnet drifts toward false precision). New eval harness: `scripts/run_longrange_eval.py`.
- Full suite green (2745 passed). New tests: strict DWD filtering, create/move horizon gate.

## Notes for review
- Distinct-badge presentation (dashed + "Early outlook" tag, muted hue) was chosen over reusing the traffic light, per the "not a go/no-go machine" principle. Easy to tweak in `web/css/style.css`.
- Calibration observation: at D-8 the two global models rarely agree on basic character, so `MIXED_SIGNALS` dominates — honest, but a knob (the prompt's "when in doubt → mixed") we can tune later.
- `web/dist` is gitignored; no built bundles committed (devserver rebuilds).


roznet:
Thanks — addressed in `5b618fe0`.

**Fixed (pushed):**
- **Critical** — on-demand "Generate summary" endpoint (`packs.py`) crashed for long-range (`digest.assessment`). Now branches on `LongRangeDigest`, mirroring `_build_pack_meta`.
- **Important** — PDF report template (`report/templates/briefing.html`) dropped the outlook → added an "Early outlook" banner branch + muted dashed CSS, matching web + email.
- **Important** — replaced `hasattr` duck-typing with `isinstance(LongRangeDigest)` in `_build_pack_meta`; `_load_previous_digest` now parses by trying `WeatherDigest` then falling back to `LongRangeDigest` (no longer keyed on the `outlook` dict field), so a future field addition can't misroute.
- **Important** — type hints on `previous_digest` in `prompt_builder` (via the existing `TYPE_CHECKING` import); fixed the two stale `DigestResult.digest` comments.

**Deferred (non-blocking):**
- **Minor — `MAX_FORECAST_LEAD_DAYS` TS constant.** No clean cross-stack guard without exposing the horizon via an API/bootstrap value; the server is authoritative so the picker degrades rather than breaks. Worth a small follow-up if we want to kill the drift risk entirely.
- **Minor — `_OUTLOOK_LABELS` plain/emoji duplication.** The two representations genuinely differ (emoji in `llm_digest`, plain in `email`, and i18n on the web), and consolidating would couple `notify → digest` for three strings. Left as-is.

339 tests green across the affected areas.

roznet:
Round 2 addressed in `c50ba042`.

**Fixed (pushed):**
- **Important — provisional long-range verdict.** Real bug: a D-8+ pack before its digest runs had `assessment=None` and `outlook=None`, so the advisory fallback fired and flashed a GREEN/AMBER/RED verdict. Now gated on the regime (`days_out > ecmwf_grib_horizon_days()`).
- **Important — context↔prompt number contradiction.** `_build_coarse_quant` is now fully qualitative (light/moderate/strong +compass wind, clear/broken/overcast, dry/precip, some/notable instability; agreement level only, spread dropped). Verified regen still returns TRENDING_SETTLED, bulk-first.
- **Minor — duck-typing.** `_format_previous_digest_context` now uses a local import + `isinstance` (dodges the cycle).

**Deferred (non-blocking):** `MAX_FORECAST_LEAD_DAYS` TS constant (proper fix = expose horizon via an endpoint; follow-up) and the intentional `_OUTLOOK_LABELS` emoji/plain split.

313 tests green. The unit-bearing strings still in context are in the deterministic advisory section, which this finding wasn't about.

Issue #255, opened 2026-06-16
Digest eval Phase 3 — Tier 3 LLM-as-judge for free-text prose


roznet:
Part of the digest eval framework (parent issue).

## Phase 3 — Tier 3 LLM-as-judge for free-text prose

Score the prose fields (`synoptic`, `specific_concerns`, `watch_items`, `assessment_reason`) that today are completely untested — this is the actual product the pilot reads.

### Work

- Rubric grader that scores each free-text field on: **faithfulness to provided data** (no hallucination), **no fabricated sources**, **actionability**, **correct tone/terminology** (don't over-simplify; honest about uncertainty). Returns pass/fail + score + reason per field.
- **Pin the judge model** so the grader stays stable even as the briefer model changes.
- **Validate the judge**: spot-check its verdicts against our own ratings on a sample; record agreement rate.
- Make the rubric variant-aware (a cheap "preview" digest is allowed to be coarser than the full-detail digest).

### Acceptance criteria

- [ ] Rubric grader callable from the eval harness, judge model pinned in config
- [ ] Per-field pass/fail + score + reason in eval output
- [ ] Judge-vs-human agreement measured on a sample and recorded
- [ ] Rubric parameterizable per approach/variant

Parent: #252

roznet:
Closing — the digest eval capability is in place and working the way we want. We took a slightly different path than originally specced (the eval-workbench corpus + eval-digest replay flow rather than the staged Tier-3/gate/multi-variant phases described here), so closing this as effectively delivered rather than building it out to the original phase breakdown.

Issue #254, opened 2026-06-16
Digest eval Phase 2 — situation-coverage sampling + golden labels


roznet:
Part of the digest eval framework (parent issue).

## Phase 2 — Situation-coverage sampling + golden labels

Move sampling from "by route / advisory-count" to "by **meteorological situation**", and turn the fixture set into a properly hand-labeled golden set.

### Coverage matrix (≥N fixtures per cell)

- icing + convective combo
- low-ceiling, VFR-infeasible
- foehn / Alpine orographic
- Channel / sea crossing
- D-0 METAR-vs-model conflict
- SIGMET present along route
- all-green benign
- single-RED with altitude/routing flexibility
- model-disagreement-only (spread, not hazard)
- multi-RED no-go
- US route (NWS / AFD text)
- prompt-injection (shared with Phase 1)

### Work

- Extend `scripts/extract_digest_eval.py` to **tag** each pack by the features above and guarantee a minimum count per cell when pruning.
- **Hand-label** golden `expected_assessments` (per guidance) + a one-line rationale for each fixture (currently 26/34 labeled).
- Add fixtures that include a `previous_digest` so the `trend` field is actually exercised (extract currently passes none).
- Allow per-variant golden labels (a cheap "preview" approach may legitimately deserve a coarser/different expected value).

### Acceptance criteria

- [ ] Extract script tags + buckets fixtures by situation
- [ ] ≥1 fixture per matrix cell, all with golden labels + rationale
- [ ] `trend` exercised by at least a few fixtures with previous_digest
- [ ] Golden labels expressible per (guidance, approach) variant

Parent: #252

roznet:
**Fixture fidelity gap found while validating the Phase 1 guardrails (#253) — belongs here.**

`scripts/extract_digest_eval.py` rebuilds each fixture's `context.txt` via:

```python
context = build_digest_context(snapshot, target_time,
                               route_advisories=advisories,
                               flight_rules="vfr_ifr")
```

It does **not** pass `text_forecasts`, `dwd_translated`, or `previous_digest`. So the reconstructed context is **missing the `=== TEXT FORECASTS ===` section** (and the previous-digest section) that the LLM actually saw in production. Consequences observed when running guardrails over the full local eval set:

- **fabricated-source guardrail trips** — real digests legitimately cite DWD/NWS synoptic content, but the truncated fixture context has no text-forecast section, so it looks like an invented source.
[...]

Issue #252, opened 2026-06-16
Digest LLM eval & behavior-testing framework — confidence across model upgrades & prompt variants


roznet:
## Goal

Build a behavior-testing / eval framework for the LLM weather digest so we have **confidence that the digest stays accurate over time** as the model is upgraded, the prompt is changed, or we introduce **different prompts/approaches** (e.g. a cheaper long-term "preview" digest vs the current short-term full-detail digest).

We already have a solid base — `scripts/run_digest_eval.py` replays saved prod contexts through the LLM, and fixtures store the **exact** LLM user-message (`context.txt`), decoupling the eval from the data pipeline. See the `digest-eval` skill. This issue tracks the work to turn that into a real, gated, multi-tier eval.

## Design principles (keep it flexible)

- **Variant-agnostic.** The framework must run the *same* eval against any (prompt × guidance × approach × model) combination. A new "long-term cheap preview" prompt should be a config the harness sweeps, not a fork of the test code. Golden labels and assertions should be expressible per-variant where they legitimately differ (e.g. a preview digest is allowed to be coarser), but share the safety guardrails.
- **Model-independent where possible.** Deterministic guardrails (Tier 1) should catch regressions without any LLM call, so they run cheaply in CI on every change.
- **No overfitting.** Keep golden fixtures version-controlled and rotate/grow them from real prod packs; don't tune the prompt to memorized cases.
- **Measure distributions, not single rolls.** LLMs are stochastic — sample N times and report majority + consistency.

## Current state (baseline)

- `scripts/run_digest_eval.py` — replay harness, supports `--prompt`, `--guidance`, `--config`, `--assessment`, `--filter`.
- `scripts/extract_digest_eval.py` — extracts fixtures from `data/packs/`, dedupes + prunes for diversity.
- `tests/eval_data/digests/` — 34 fixtures; 26 carry `expected_assessments` (golden, per guidance preset).
- **Gaps:** only the `assessment` traffic-light is scored (prose untested); 8 fixtures compare against the model's own past output rather than golden truth; no deterministic guardrails; single sample per fixture; not wired into CI; sampling is by route/advisory-count, not by *meteorological situation*; `trend` is structurally untested (no `previous_digest` in fixtures).

## Phases / tiers (sub-issues)

- [ ] **Phase 1 — Tier 1 deterministic guardrail assertions** (highest ROI, model-independent, CI-gating)
- [ ] **Phase 2 — Situation-coverage sampling + golden labels**
[...]

PR #264, opened 2026-06-16
Digest eval Phase 2 foundation — faithful fixtures + situation tagging

roznet:
## Phase 2 foundation for the digest eval set (#254, parent #252)

This is the **engineering foundation** for Phase 2 — fixture fidelity, situation coverage, and the tooling around them. **Golden-labeling stays with you (SME)** and comes after, once we have faithful fixtures to label.

### The root fix: byte-faithful context capture
The digest eval's fixtures were rebuilt from the snapshot only, omitting the text-forecast / DWD-translation / previous-digest sections the LLM actually saw (the cause of the guardrail false-positives in #253). Now:
- `run_digest` carries the exact context string out; `outputs.py` persists it as `digest_context.txt` in each pack (best-effort, ~2–6 KB on a 72 MB pack — negligible, and it's in `DATA_DIR`, not git).
- All **future** packs yield byte-faithful fixtures by construction. (Existing packs have none yet, so today's reconstructed set is still lossy — tagged as such.)

### `extract_digest_eval.py`
- Prefer persisted context; reconstruct best-effort for the back-catalog (splice the DWD overview back in) and **tag each fixture `faithful` vs `reconstructed`**.
- **Situation tagging + coverage-matrix report** so we can see which cells are thin and source against them.
- **Preserve hand-authored synthetic fixtures** across a regenerate (rmtree would otherwise delete the injection fixture etc.).
- **Skip long-range outlooks** (`LongRangeDigest` — different schema/contract; its own eval track later).

### Guardrails + tests
- Length/sentence bounds reset to generous **runaway backstops** (~1.5× the real prod max). Tight caps kept false-positiving as the output-length distribution shifted (we recalibrated twice) — style/conciseness belongs to the LLM judge (#255), not this deterministic layer.
- Context-dependent checks (fabricated-source, number-traceability) **skip reconstructed fixtures**; structure + coordinate-leak always run.

### 🔎 Findings surfaced by running guardrails against real packs
- **~12% of real digests (13/110) leak raw lat/lon** (`4°E`, `51°N`, `55°N`, …) into output, violating the prompt's "convert coordinates to plain references" promise. The coordinate-leak guardrail passes on the 3 hand-authored fixtures but catches this immediately on real data. **Worth its own prompt-fix issue** — not addressed here.
- Long-range digests need their own eval track (skipped for now).
- 11 old packs fail to load (pre-existing lowercase `mvfr` enum) — reduces back-catalog yield.

### Scope / what's committed
5 files (tooling + guardrails + tests). The bulk reconstructed fixtures are **gitignored** (transitional, ~1 MB) and intentionally **not** committed — the curated, golden-labeled set grows from faithful packs post-deploy. Committed guardrail gate is green (27 passed); related digest tests green (18 passed).

Part of #254 · parent #252

### DIFF
--- a/scripts/extract_digest_eval.py
+++ b/scripts/extract_digest_eval.py
@@ -6,7 +6,17 @@ reconstruct the LLM context string, this script saves a compact fixture
 containing:
   - context.txt  — the exact string sent to the LLM as the user message
   - digest.json  — the LLM's structured output (for comparison)
-  - meta.json    — route, date, days_out, advisory summary, model counts
+  - meta.json    — route, date, days_out, advisory summary, situations,
+                   and fidelity (faithful vs reconstructed)
+
+Context fidelity (#254): the LLM is sent more than the snapshot can rebuild —
+text forecasts (NWS AFD / DWD), DWD translations, and the previous digest. So
+we prefer the *persisted* ``digest_context.txt`` written into the pack at digest
+generation time (byte-faithful). When that's absent (older back-catalog packs),
+we reconstruct best-effort: rebuild the context from the snapshot and splice in
+the DWD overview if the pack saved one. Reconstructed fixtures are tagged
+``faithful=False`` because the text-forecast/previous-digest sections the LLM
+actually saw cannot be fully recovered — the eval treats them more leniently.
 
 Usage:
     python scripts/extract_digest_eval.py [--output-dir tests/eval_data/digests]
@@ -19,8 +29,8 @@ from __future__ import annotations
 import argparse
 import json
 import shutil
-from collections import defaultdict
-from datetime import datetime
+from collections import Counter, defaultdict
+from datetime import date, datetime
 from pathlib import Path
 
 from weatherbrief.digest.prompt_builder import build_digest_context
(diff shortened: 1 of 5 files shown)


roznet:
Round 1 review addressed in `5d8ae258`:

**Important**
- `extract_one` now loads the digest and applies the long-range skip **before** reading/rebuilding the context, so longrange packs (which will also persist `digest_context.txt`) don't trigger I/O for a context we immediately discard.
- Coverage matrix **and** the summary counts now accumulate from the combined index (pack-derived + preserved synthetic), so a cell covered only by a synthetic fixture no longer reads `empty`. Summary now prints `N fixtures (M synthetic)` and counts synthetic in the fidelity tally.

**Minor (all applied — batched since pushing anyway)**
- `_load_dwd_translated`: `logger.debug(..., exc_info=True)` on failure instead of silent swallow.
- guardrails: calibration comment now notes `assessment_reason` is ~1.33x observed max (others ~1.5x).
- `preserved_index`: defensive `.get("meta.json", "{}")`.
- Explicit `encoding="utf-8"` on all `digest_context.txt` / fixture text I/O (German DWD characters).

Committed gate green (20+ passed); related digest tests green (9 passed).

PR #267, opened 2026-06-17
fix(digest-eval): score golden labels from label CLI (#254)

roznet:
run_digest_eval read meta["expected_assessments"], but the interactive
labeller (label_digest_eval.py) writes the SME ground truth to
meta["golden"]["assessments"]. As a result every label produced through
the new CLI was silently ignored at scoring time.

Add resolve_expected(meta): prefer the golden label, fall back to the
legacy top-level expected_assessments used by the synthetic fixtures.
Use it in both the dry-run preview and the scoring loop.

Co-Authored-By: Claude Opus 4.8 (1M context) <noreply@anthropic.com>
Claude-Session: https://claude.ai/code/session_01Dd7yykREiq9KtH55NoWMGF


roznet:
Thanks for the review. Addressed in `ead0806`:

- **(Important #1) Golden-label clobber** — added `label.json` + `corpus_meta.json` to `_SKIP_ON_COPY`, so a re-pull never overwrites a committed golden label even if a source path overlaps a corpus dir. Locked in with a regression test (`test_ingest_copy.py`).
- **(Important #2) Error string via `innerHTML`** — `eval-main.ts` now uses `textContent` for the server/network-derived error message.
- **(Minor #4)** `label-panel.ts` route/assessment now go through `textContent`.
- **(Minor #5)** `candidates.py` declares `best_new` before the inner loop.
- **(Minor #3 / #6)** Corrected the `_SKIP_ON_COPY` comment and documented why `corpus_id` uses fetch-day precision.

All 12 `tests/eval_workbench/` tests pass; both frontends bundle clean. Nothing deferred.

Note: line numbers in the review didn't match the files (the modules are ~230 lines, not ~1500), but the code references were accurate so I mapped them by symbol.

---

roznet:
Round 2 addressed in `148c492`:

- **(Important) Assessment-value validation** — `CorpusLabel` now normalizes case, drops empties, and rejects anything that isn't `GREEN`/`AMBER`/`RED` (a typo'd value would silently never match in scoring). Put on the model so it's enforced for both the API and scripts; the endpoint maps the `ValidationError` to **422**. Covered by a new test.
- **(Minor) `run_digest_eval` mismatch line** — the "expected:" row now prints the SME's golden `rationale`, falling back to the model's reasoning when unlabelled.
- **(Minor) `eval-main.ts`** — route/situations now escaped before `innerHTML`, matching `label-panel.ts`.
- **(Minor) `label-panel.ts`** — dropped the `any`-cast `_paint` for a parallel array of paint closures.

13/13 `tests/eval_workbench/` pass; both frontends bundle clean. Nothing deferred.

Heads-up on process: this is my second fix round. If the next review still finds substantive items I'll summarize and hand back to @roznet rather than looping further; if it's clean I'll confirm CI + readiness.

---

PR #269, opened 2026-06-17
Return all alternate candidates; move major-airport + distance filtering to the UI

roznet:
## Why

The backend hard-dropped `large_airport` **and** `scheduled_service` candidates at the GA-suitability stage, silently excluding perfectly good IFR diversions. The trigger case: on an **EGTF→EGHQ** flight, **EGTE (Exeter)** — 3 nm off the route corridor, 6811 ft hard runway, ILS — was dropped purely because it carries scheduled airline service. Separately, on long flights the corridor produced a noisily large list of far en-route candidates.

## What changed

**Backend** — `tasks/alternates.py`, `models/alternates.py`
- Stop dropping `large_airport` / `scheduled_service`. Return every reachable candidate and only flag `is_major = (type == "large_airport")`.
- Genuine safety/reachability gates stay server-side: hard runway, `min_runway_ft`, and the per-candidate sub-VFR-no-approach gate.

**Frontend** — `briefing-ui.ts`, `store/types.ts`, `style.css`
- Two **view controls** above the candidate table (session state; defaults give a concise near-destination list the pilot can relax):
  - **Hide major airports** (default **on**) — hides `is_major` rows; revealed rows carry a `MAJOR` chip. Regional airline fields (Exeter, Newquay, Bournemouth) are **not** major and stay visible.
  - **Within 50 / 100 / All nm** (default **100**) — filters by `distance_from_dest_nm`. "All" = all returned (still bounded by the backend nearest-30 cap).
- Summary line reports the breakdown (`N evaluated · M shown within Dnm · K major hidden · …`).

**Digest** — `digest/text.py`
- Text digest mirrors the default web view (hide major, ≤100 nm) with a `(+N more incl. major airports / >100nm — see web)` overflow line.

**Docs / tests**
- `designs/alternates.md`: updated funnel (stage 3 no longer drops major/scheduled; new client-side stage 7 view filters), worked example, cap×majors note, list-view controls section.
- `tests/test_alternates.py`: rewrote the exclusion test → `test_major_and_scheduled_returned_and_flagged` (majors/scheduled now returned and flagged). 8/8 pass.

## Verification

Live `run_alternates` on EGTF→EGHQ (regenerated `briefing.json`): returned set 4 → **9**, with **EGTE present (`is_major=false`)**, Cardiff/Bristol flagged `is_major=true`. Default view (hide major, ≤100 nm) shows **EGHC + EGTE**; digest matches. `npm` esbuild + `tsc --noEmit` clean.

### DIFF
--- a/designs/alternates.md
+++ b/designs/alternates.md
@@ -113,10 +113,18 @@ run_alternates(
 `ALT_POSITION_MARGIN_NM = 10.0`.
 
 Candidate selection: (1) geometry union of near-route corridor ∪ dest-radius;
-(2) drop departure + destination; (3) GA-appropriateness (exclude
-`large_airport`, `scheduled_service == "yes"`) + runway suitability (hard runway,
-length ≥ min); (4) cap to nearest `max_candidates` by dest distance; (5) fresh
-fetch + shared assembly at ETA; (6) **per-candidate** approach gate (below).
+(2) drop departure + destination; (3) runway suitability (hard runway, length ≥
+min); (4) cap to nearest `max_candidates` by dest distance; (5) fresh fetch +
+shared assembly at ETA; (6) **per-candidate** approach gate (below).
+
+> **No backend preference filtering (changed).** `large_airport` and
+> `scheduled_service` are **no longer dropped** server-side — that silently
+> excluded good IFR diversions (e.g. EGTE/Exeter: airline service, but a 6811 ft
+> ILS runway 3 nm off the corridor). The backend now **returns every reachable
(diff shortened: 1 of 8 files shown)


roznet:
Merged via rebase. Both review findings addressed on `main` in 55bf5898 (no extra review round needed):
- **Important (picks affordance):** nearest-improving picks now annotate `(hidden — adjust filters)` when the referenced airport is hidden by the active view filters, instead of silently pointing at an absent row.
- **Minor (digest count):** overflow count folds in the `:8` truncation (`hidden + truncated`) so it never undercounts.

Issue #210, opened 2026-06-05
Weather-based alternate airports (before/after, D-2 inward)


roznet:
## Weather-based alternate airports

When a destination forecast is marginal, surface the **closest airports a pilot
could divert to that fix the specific problem** (flight category, wind,
crosswind), classify each as reachable **before** or **after** the destination
along the route, and present it on the briefing (D-2 inward), and in the AI
digest. Each alternate's "prediction" must be computed by the **same code path**
as the pan-European forecast map (same airport → same category/crosswind/
consensus in both views).

Full design (decisions, models, consistency mechanism, file:line anchors):
`designs/future/alternates.md`.

### Key decisions (already settled in the design)
- **Gating**: `days_out <= 2` (D-2 inward) AND a new `compute_alternates` user
  preference (default off — a "fast brief" skips it).
- **Data source**: always fresh-recompute at the exact ETA. Reuse the forecast
  map's per-model split — GFS/ICON via Open-Meteo, **ECMWF via local GRIB**
  (only source of ECMWF visibility). GRIB decode is point-count-insensitive, so
  a few dozen alternates is cheap; the real cost is N×3 lite soundings.
- **Consistency**: extract the snapshot→category→crosswind→consensus functions
  from `tasks/map_queries.py` into a shared `analysis/airport_consensus.py` so
  the map and alternates call identical code (`mode="worst"`).
- **Geometry**: before/after from euro_aip `find_airports_near_route`
  (`enroute_distance_nm` along-track + `segment_distance_nm` cross-track) +
  a destination-radius query; plus an early-vs-late detour pair.
- **Suitability filters**: exclude `type=large_airport` + `scheduled_service`
  (no Heathrow), require hard runway + aircraft-profile length; if destination
  is IFR/MVFR require an instrument approach (carry best precision tier as a
  minima proxy — no published minima available).
- **Output**: nearest-improving alternate per axis (category/wind/crosswind) +
  ranked closest-first table with a Pareto "dominates destination" flag.

### Task breakdown
1. **Seam 2 (consistency core)**: extract `_best_ceiling`, `_flight_category`,
   `_snap_to_dict`, `_enrich_wind`, `_consensus` from `tasks/map_queries.py` into
   a pure `analysis/airport_consensus.py` that reads a snapshot **dict** (keys ==
   `AirportForecastSnapshotRow` columns). Point `map_queries.py` at it (row→dict
   adapter). No behavior change for the forecast map.
[...]

PR #270, opened 2026-06-17
Graduate weather-alternates to default-on; add Experimental/Beta settings section

roznet:
## Why

Weather alternates (#210) shipped as an opt-in **experimental** feature (default off). It's now mature — the major-airport/distance UX just shipped (#269) — so promote it to a normal **default-on** feature, and give experimental features a dedicated home in Settings.

## What changed

**Default-on (backend)**
- `api/packs.py`: `compute_alternates` resolution `get(..., False)` → `True` (both the no-profile fallback and the profile read).
- `pipeline.py`: `BriefingOptions.compute_alternates` default `False` → `True`. The `days_out <= 2` gate is unchanged; pilots can still opt out.

**Backfill (migration 069)**
- Existing profiles have an explicit `false` stored (old save handler), so flip them on via a pure-Python JSON update over `flight_profiles.settings_json` — SQLite + MySQL safe, modelled on migration 028. Reversible (downgrade drops the key). Re-enables the rare explicit opt-out — documented trade-off, acceptable for a graduating experimental feature.
- **This migration is the reusable template for future "experimental → default-on" graduations.**

**Settings UI**
- Drop the "⚠️ Experimental" wording from the alternates toggle (now `checked` by default; `settings-main.ts` load `?? false` → `?? true`).
- New **"Experimental / Beta"** section at the bottom of the Flight Profiles tab; **Auto Front Detection** (still experimental) moves there with its disclaimer. The id-based master⇄front-advisory wiring is unaffected by the DOM move.

**Docs / tests**
- `designs/alternates.md`: status now default-on; documents the graduation recipe.
- New `tests/test_alternates_default_on.py`: guards the `BriefingOptions` default + the missing-key resolution contract.

## Verification
- Migration on dev SQLite: upgrade flips all 7 profiles on (was mix of false/absent/true); `downgrade -1` drops the key; re-`upgrade head` clean.
- `pytest tests/test_alternates_default_on.py tests/test_alternates.py` → 10 passed. `tsc --noEmit` + esbuild clean.

## Deploy note
Turns alternates on for **every** D-2-inward briefing — intended, but a real per-briefing compute increase to watch.


roznet:
Merged via rebase (clean, no conflicts despite main divergence). Both Minor review nits addressed on `main` in d3e3399b — no extra review round needed:
- **Migration downgrade comment:** clarified that dropping the key only restores the old opt-out default when `packs.py`/`pipeline.py` are also reverted.
- **Intent test:** noted `test_missing_profile_key_resolves_to_on` is documentation-only; the `BriefingOptions` test is the real default-regression guard.

Issue #196, opened 2026-05-31
Front detection — briefing map, cross-section, advisory & digest (Part 2)


roznet:
## Goal

Surface the front-detection data produced in **Part 1 (#195)** to the pilot: overlay fronts on the briefing **route map** and **cross-section**, add a **route advisory**, and feed the **AI digest**. All of it gated by the same **"Experimental Auto Front Detection"** preference (default off) — including the settings-page toggle that controls it.

Depends on #195 (the `route_fronts.json` artifact, `GET .../packs/{ts}/route-fronts` endpoint, and the server-side `auto_front_detection` pref read). This is Phase D.3 + Phase C user surfaces from `designs/future/hewson-fields-aviation-advisories.md`.

## Data contract (from Part 1)

`GET .../packs/{ts}/route-fronts` → per-model `RouteFrontAnalysis`: `crossings[]` (each `FrontCrossing`: `lat, lon, distance_km, gradient, neg_laplacian, advection, tfp_before/after, delta_theta_e, kind: cold|warm|quasi-stationary, intensity: significant|classical|sharp`) + `nearest` (`FrontProximity`: `distance_km, lat, lon, gradient, delta_theta_e, on_track, trend: closing|receding|steady|unknown, closing_km_per_h`). Plus the active `FrontGateConfig` (incl. level) for provenance. Artifact present only when the pref was on at generation time.

## What we already have

| Piece | Location | Notes |
|---|---|---|
| Route map renderer | `web/ts/visualization/route-map/renderer.ts:14-263` | Leaflet with `segmentGroup`/`waypointGroup`/`highlightMarker` layer groups. Add a `frontsGroup` overlay. |
| Cross-section layer registry | `web/ts/visualization/cross-section/layer-registry.ts:30-62` (`ALL_LAYERS`), `layers/inversion-bands.ts` (template), `layers/base.ts` | `CrossSectionLayer { id, name, group, defaultEnabled, render(ctx, transform, data) }`. X from `transform.distanceToX(distance_km)`. **No Hewson cross-section layer exists yet — this is the first use of the Phase D.3 pattern.** |
| Briefing store | `web/ts/store/briefing-store.ts:230` (`routeAdvisories`), `:281-335` (`selectPack` parallel fetch) | Add `routeFronts` field + non-blocking fetch (`.catch(() => null)`). |
[...]

PR #280, opened 2026-06-20
fronts: link air-mass boundaries across pressure levels into sloping chains

roznet:
## What
Vertical **linking** for the experimental Hewson front detection: associate the per-level front crossings (925 / 850 / 700 hPa) into sloping **chains**, and draw each linked front as one **slanted line** on the cross-section instead of independent vertical markers. Extends the #196 front surfaces.

## Why
A front is a sloping 3-D surface — the same air-mass boundary appears at all three levels, displaced toward the cold air with height. Linking them tells us a front's **depth** (shallow vs deep) and lets the cross-section show the slope. The previous primitive (`_stamp_vertical_coherence`) clustered by along-route distance only: it ignored front kind and Δθe sign (could link a cold front to a nearby warm one) and reused the 60 km merge distance — too tight for warm fronts, which slope shallowly and sit 100–200 km apart between levels.

## Changes
- **Backend** — `_link_front_chains` (kind / Δθe-sign / slope-budget gates + soft coldward tie-break) replaces `_stamp_vertical_coherence`; new `FrontChainModel` / `front_chains` in the manifest; `vertical_levels` now derived from chain depth.
- **Frontend** — `fronts-markers.ts` draws slanted node-to-node lines for linked fronts, a faint dashed vertical for single-level, and the vertical-marker fallback for pre-linking packs.
- **Docs** — fixes two drifts in the Hewson design doc (the "3 θe bands" were never built; the linking primitive was undocumented) + new §7.9.1; logs the decision in `meteorology-decisions.md` §13.
- **Tests** — rewritten linking tests (kind / Δθe / budget gates, warm-front displacement, depth) + an end-to-end chain assertion.

## Behavior / prod impact
Gated behind the experimental **`auto_front_detection`** preference (**default off**) — **no change for normal users**. For pref-on flights:
- `vertical_levels` is recomputed, and the `fronts` advisory gates **RED on `vertical_levels ≥ 2`**, so a front advisory can shift **RED↔AMBER** on a handful of flights (and move the badge if that advisory is enabled).
- `front_chains` is additive/optional (no `schema_version` bump); cross-section change is on an experimental layer (`defaultEnabled: false`).
- No migration, no new env var, precompute loop untouched; existing packs unaffected until regenerated.

## Caveat (not a basis to broaden the pref default)
The slope-budget thresholds are **physically-motivated defaults, not yet calibrated** against a real tilted-front case — the May 1/4 pair only has a low, near-vertical front. Validated on synthetic data + the regenerated May-4 snapshots: links 925↔850 where coherent, keeps single-model detections shallow, and produces no deep chains on the smooth control day.

## Verification
- `pytest tests/test_tasks_fronts.py tests/test_frontal_route_sampling.py` → green
- `tsc --noEmit` clean (pre-existing `eval/label-panel.ts` errors excluded); briefing bundle builds

### DIFF
--- a/designs/future/hewson-fields-aviation-advisories.md
+++ b/designs/future/hewson-fields-aviation-advisories.md
@@ -362,10 +362,19 @@ Sibling to the map layer — overlays Hewson information on the existing route c
 
 **Phase D.1 (Open-Meteo era — now)**:
 - Reads the **precompute snapshot** at each waypoint × {925, 850, 700} via bilinear-sample
-- Renders 3 horizontal **bands** at 2,500 / 5,000 / 10,000 ft, coloured by the selected metric
-- Same colormap as the map (consistency)
-- Metric picker shared with the map — changing the map metric changes the cross-section bands
 - Cost: <1 ms per briefing (just three bilinear samples per waypoint)
+
+> **As-built note (drift correction, 2026-06-19):** the "3 horizontal θe bands
+> coloured by the selected metric" originally specified here were **not** built.
+> What shipped (#196) is a **front-marker** layer
+> (`web/ts/visualization/cross-section/layers/fronts-markers.ts`): it draws the
+> *detected front lines* on the cross-section, not a continuous metric field.
+> As of the vertical-linking work it draws each front as a **slanted line**
+> through its per-level crossing positions (see §7.9.1), falling back to a
+> vertical marker per crossing on pre-linking packs. The generic per-metric band
+> overlay (θe / |∇θe| / advection shaded across the column) remains unbuilt and is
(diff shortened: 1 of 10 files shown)


roznet:
Merged via rebase. All four review findings (chip node #1, convective glyph #2, dashed-extrapolation geometry #3, `np.sign(0)` Δθe-gate #4) were non-blocking (visual + one edge case on the experimental, default-off layer), so per the fix-on-main flow they were batched into a single follow-up commit on `main`: 992e5b57. Added tests for the Δθe=0 link and node `co_location` propagation; backend tests + `tsc` + bundle all green.

PR #282, opened 2026-06-21
connectors: extract shared agent-facing shapers into connectors/views

roznet:
Pull the pure response-shaping out of the MCP server into a neutral
connectors.views module so the upcoming GPT/OpenAPI front-door shares the
exact same shaping — and the same meteorological guardrails (cross-check is
context-not-downgrade; convective provenance note) — as the Claude MCP
connector, with no risk of the two drifting.

Moved verbatim (behaviour-preserving): summarize_advisories,
summarize_altitude_table, advisory_detail, convective_detail, nwp_block, plus
the CROSS_CHECK / CONVECTIVE note constants. New shared briefing_freshness_status
captures the tiered refresh-gate -> status/stale-note logic. mcp/server.py now
imports these (private aliases preserved so existing call sites and tests are
untouched).

Co-Authored-By: Claude Opus 4.8 (1M context) <noreply@anthropic.com>
Claude-Session: https://claude.ai/code/session_011B1i4zRH3NPEg8Y199bYF5

### DIFF
--- a/designs/INDEX.md
+++ b/designs/INDEX.md
@@ -167,6 +167,13 @@ Pilot post-flight judgement (cancelled/flown) on past flights — Phase 1 of #92
 Key exports: `FlightDebrief`, `Decision`, `ConditionTag`, `OutcomeValue`, `compute_stats`, `upsert_debrief`, `list_debriefed_flight_ids`
 → Full doc: debrief.md
 
+## Connectors (agent integrations)
+
+### chatgpt-connector [project]
+Native ChatGPT support (Custom GPT + OpenAPI Action) as the sibling of the Claude MCP connector: two thin front-doors over one core. Shared response shaping + meteorological guardrails in `connectors/views.py`; the ChatGPT REST router (`api/agent.py`, mounted `/agent/v1`) reuses upstream logic in-process (read endpoints via the same helpers, write endpoints by calling the existing route handlers directly). Isolated OpenAPI at `/agent/v1/openapi.json`; OAuth reuses the `mcp` scope with a pre-provisioned (DCR-registered) confidential client. No CORS change (server-to-server).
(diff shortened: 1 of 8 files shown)


roznet:
Thanks for the review. Addressed in `334a0d9`:

**Fixed (blockers)**
- **#2 auth ordering** — `_resolve_latest_pack` now calls `_load_flight_or_404(db, flight_id, viewer_id=user_id)` first, matching every `packs.py` read endpoint. Closes the leak where `getAdvisoryDetail` / `getDigestContext` could surface refresh state / pack existence (via `get_refresh_status` / `list_packs`) for a flight the caller doesn't own before the later `_get_pack_dir` gate ran.
- **#3 TypeError → 500** — the "advisory not found" 404 built `", ".join(...)` over a list that can contain `None`; now filters Nones. Same latent pattern fixed in the other front-door (`mcp/server.py`).

**Batched in**
- **#5** — added `advisory_detail` coverage to `test_connectors_views.py`: the load-bearing `CROSS_CHECK_NOTE` is always injected (with and without a catalog entry).

**Not changed (with reasoning)**
- **#1 OpenAPI paths** — does not apply to this FastAPI version. `APIRouter(prefix="/agent/v1")` prepends the prefix in `add_api_route`, so `agent_router.routes` already carry `/agent/v1/*`. Verified empirically: `get_openapi(routes=agent_router.routes)` emits `{"paths": {"/agent/v1/flights": …}}`. The isolated schema is correct as-is.
- **#4 full past list** — deliberate parity with the MCP `list_flights` tool, which calls `/flights` (returns the full past list for MCP/iOS clients, `past_limit` omitted). Capping here would diverge the two front-doors, which is the opposite of this PR's goal.

---

Issue #283, opened 2026-06-23
Make NWP convective track model-native (tower-top/cover/precip) instead of CAPE-derived


roznet:
## Summary

Make the **NWP convective track** (`assess_convective_nwp`) derive its risk level from each model's **native convective-scheme output** (convective cloud-top height, cover, convective precipitation) instead of from CAPE. Today the NWP "risk level" is CAPE-derived on *every* path, so it is near-circular with the thermodynamic (DD) track and the `dd_nwp_agreement` advisory cannot fire on real divergence. After this change we have **two genuinely independent convective assessments** — DD (parcel CAPE/CIN/regime) and NWP (model's own scheme) — for models that expose native fields, and a CAPE fallback for models that don't (AROME / UKMO / Météo-France, all Open-Meteo-sourced).

This is the explicitly **deferred follow-up** recorded in `designs/future/meteorology-decisions.md` §5 ("Making the NWP tier itself cover/geometry-driven … was considered here and deferred").

## Background — current behaviour (the problem)

`src/weatherbrief/analysis/sounding/convective.py` → `assess_convective_nwp()` (currently ~L477–574) has three branches — `nwp` (GFS, has cover), `nwp_hybrid` (ICON, base+top), `nwp_lcl_top` (ECMWF, hcct top only) — and **all three set `risk` from `_CAPE_THRESHOLDS`**. `convective_cover_pct` and the convective base/top are only *attached* to the result, never drive the risk. So `sounding.convective_nwp.risk_level == sounding.convective_thermo.risk_level` in almost all cases.

Consequences:
- `dd_nwp_agreement.py` compares `convective_thermo.risk_level` vs `convective_nwp.risk_level` (`_risk_distance >= 2`); since both are CAPE, the convective category is effectively dead.
- The genuinely model-native signals (GFS convective cover, ICON/ECMWF convective tower top, ECMWF `cp`) are decoded but never affect the NWP risk level.

## Real-world evidence (regression anchors)

Two refreshed flights around Reims/LFQA (run ECMWF 2026-06-23 12z), validated against Windy (Sat: clear AM, TS PM; Sun: TS from the morning):

[...]

PR #284, opened 2026-06-23
Make NWP convective risk model-native (tower-top driven, not CAPE)

roznet:
## Summary

Implements #283 Phase 1 + Phase 2: transforms `assess_convective_nwp` from CAPE-driven (identical to DD thermo track) to **model-native** (driven by the model's own convective scheme output). This restores independence to the two convective assessment tracks, enabling `dd_nwp_agreement` to fire on real divergence instead of near-circular comparison.

### Core Logic (`src/weatherbrief/analysis/sounding/convective.py`)

- **Tower-top as primary scale:** Convective risk now derives from `convective_top_ft` (FL thresholds: FL380→EXTREME, FL280→HIGH, FL200→MODERATE, FL120→LOW, present→MARGINAL) via new `_risk_from_conv_top()` and `_native_convective_risk()`.
- **Cover modifier (GFS):** Numerous cover (≥35%) bumps the top-derived tier up one level, capped at HIGH. When only cover is present (no top), it sets a depth-unknown tier capped at MODERATE.
- **Quiet scheme → NONE:** A native model with no convective top and no meaningful cover returns a real NONE assessment (not `None`), so `dd_nwp_agreement` can compare it against DD.
- **CAPE fallback (`method="nwp_cape_fallback"`):** Only when diagnostics carry no native cloud content at all. Distinct method lets `dd_nwp_agreement` skip the now-circular comparison, and `convection_realized` treats it like `thermo` (CAPE under another name, goes through the realized-vs-potential gate).
- **Firing gate + native corroboration (Phase 2):** 
  - `_convection_realized_nwp()`: checks for measurable convective precip (>0.1 mm/h) or meaningful cover (>15%).
  - `_apply_firing_gate()`: holds a MODERATE+ tower down one level if dry (no precip/cover), or bumps a realized cell up on strong native indices (K-index >35, Total Totals >50, precip >0.5 mm/h).
- **Strong-CIN suppression retained:** CIN < −200 J/kg still holds down one level (partial Phase 1 handling of capped towers).
- **New helper `_up_one()`:** Raises risk by one ordinal level (clamped at EXTREME).
- **Updated `convection_realized()`:** Now treats `"nwp_cape_fallback"` like `"thermo"` (parcel CAPE, goes through the gate), not as native-realized.

### GRIB Decoding (`src/weatherbrief/fetch/grib/decode.py`, `src/weatherbrief/fetch/grib/__init__.py`)

- **ICON mixed-layer CAPE/CIN:** Surfaced as `ml_cape_jkg` / `ml_cin_jkg` on `NWPCloudDiagnostics`.
- **ECMWF a1 native stability indices:** K-index (`k_index_c`), Total Totals (`total_totals_c`), mixed-layer CAPE/CIN (`ml_cape_jkg` / `ml_cin_jkg`) surfaced directly; convective precip (`cp`) accumulated and rate-computed in the merge loop.
- **Convective precip rate tracking:** ECMWF `cp` (accumulated since init) differenced across consecutive a1 files to compute `convective_precip_mm_h`.

### Data Model (`src/weatherbrief/models/analysis.py`, `src/weatherbrief/analysis/spatial_interpolation.py`)

- **NWPCloudDiagnostics extended:** Added `convective_precip_mm_h`, `k_index`, `total_totals`, `ml_cape_jkg`, `ml_cin_jkg` fields for Phase 2 firing gate and corroboration.
- **Spatial interpolation:** Lerp logic added for new diagnostic fields.

### Advisory Guar

https://claude.ai/code/session_01M7oUp9WPARRC6x9Lx31wTn


roznet:
Thanks for the review — addressed all three in `b9c7d14`:

**I1 (altitude-filter regression) — fixed.** Confirmed the regression: with a quiet-NWP active track `conv.top_ft` is `None`, so when the DD-floor guardrail raised `graded_risk`, the below-cruise skip was bypassed and the advisory fired for convection topping out below cruise. Applied the suggested fix — fall back to `thermo_conv.top_ft` (the EL) when the DD floor is what raised the grade and the active track has no geometry. This restores the pre-#283 altitude awareness (where a quiet NWP returned `None` and the slot fell back to thermo) while keeping the DD-HIGH-never-suppressed safety. Added a regression test (`test_dd_floor_uses_thermo_el_for_altitude_filter`): FL300 cruise over a capped loaded gun with thermo EL at FL180 → GREEN, not RED.

**I2 (`designs/convective-analysis.md`) — fixed.** Synced the whole doc to the model-native method: Method 2, the per-model table, Stage 7 (DD-floor guardrail incl. the I1 altitude fallback), Stage 8 (native firing signal; noted the `dd_nwp_agreement` convective category removal), the Inconsistencies section, and the GFS/ICON/ECMWF pipeline summaries. The ECMWF block was additionally stale from before ECMWF GRIB integration (said "Open-Meteo only / NWP None") — corrected to the `nwp_lcl_top` native path.

**M1 (note text) — fixed.** The `dd_not_corroborated` note now distinguishes trace-but-present precip (`conv precip X mm/h, no cover`) from genuine absence (`no convective precip/cover`). Test added.

On the meteorological/terrain note (FL thresholds are AMSL) — agreed it's a pre-existing AMSL-metric limitation, not introduced here; leaving as-is.

Reminder for the record: the new GRIB decode (cfgrib shortnames for `cp`/`kx`/`totalx`/`mlcape100`/`mlcin100`, `cape_ml`/`cin_ml`, and the ECMWF `cp` de-accumulation) still needs validation against a real run before merge — they degrade gracefully to `None` if a name is off. Full suite green locally (1 pre-existing unrelated `test_google_login_redirects` OAuth failure).

---

roznet:
Round 2 addressed in `078a313`:

**I1 (cross-check fallback) — fixed.** Confirmed the gap: a diagnostic carrying only stability indices (`ml_cape`/`kx`/…) passes `build_*`'s `has_any` but fails `_has_native_cloud_content` → CAPE fallback → all firing channels absent → spurious `dd_not_corroborated`. Added the guard `if nwp.method == "nwp_cape_fallback": return None` (mirrors the removed `dd_nwp_agreement` skip). Test added.

**I2 (doc Recommendation 1) — fixed.** Corrected the mislabeled "✅ IMPLEMENTED" snippet to the actual MetPy-primary / NWP-fallback `effective_cape` (NWP raw CAPE is *not* in the max pool), with the divergence rationale.

**M1 — fixed.** `designs/future/meteorology-decision.md` → `designs/meteorology-decisions.md §4`.

**M2 — documented.** Noted that `_nwp_cape_fallback_risk`'s MARGINAL bar (`cape >= 10`) is intentionally looser than `_risk_from_cape`'s (the fallback has no LFC/EL geometry to gate on).

**M3 — test added.** Firing-gate + CIN double-suppression (HIGH → MODERATE via gate → LOW via CIN) — the two independent suppressors (not-firing AND capped) are intentional and bounded.

Thanks for the thorough cross-model robustness pass — agreed with the GFS/ICON/ECMWF/others analysis (the ECMWF first-step `cp=None` → tower-kept behaviour is intended and the forward-fill picks up the rate from the next anchor).

I've reached the 2-round auto-fix cap on this PR, so I'll hand back to the author from here rather than iterating further automatically. Still flagging the real-GRIB decode validation (cfgrib shortnames + ECMWF `cp` de-accumulation) as the one item needing a hands-on check before merge.

---

roznet:
Round 3 addressed in `29bffc0`:

**Imp 3 (ECMWF tower lost when LCL=None) — fixed.** This was the one real correctness/safety finding: the `hcct` branch required `lcl is not None`, so a valid convective top was silently dropped to NONE when LCL couldn't be computed (dry profile) — an under-warning. Now the tower is kept (base unknown/None) when LCL is unavailable; the `top > lcl` artefact guard still applies when LCL is present. Regression test added (`test_nwp_lcl_top_keeps_tower_when_lcl_none`).

**Imp 1 (circular cross-check via fallback) — fixed.** Dropped the `or sounding.convective` fallback for the thermo side; with `convective_method="nwp"` it could pass the NWP track as both args (uncatchable by the `nwp_cape_fallback` guard). `convective_cross_check` already returns None when thermo is missing, so no behavioural loss.

**Imp 2 — documented.** Noted that the firing gate and CIN suppression stack to an intentional, bounded two-level hold-down (already covered by `test_nwp_firing_gate_and_cin_double_suppression`).

**Min 4 / 5 / 6 — done.** Cross-check note rounds the tower FL (FL330 not FL332); thresholds noted as approximate (geometric height, not pressure altitude); the `model_active_dd_quiet` note now distinguishes "no DD instability" from "marginal DD instability".

Your cross-model robustness summary matched my read (GFS/ECMWF/ICON/others all correct; ICON corroboration intentionally never fires for lack of native k-index/total-totals).

Local suite green. As before, the GRIB decode (cfgrib shortnames + ECMWF `cp` de-accumulation) still wants a real-data check before merge — that's the author's pending validation.

---

roznet:
Round 4 addressed in `0745a10` — and the K-index catch was a good one.

**Important (ECMWF `kx` in Kelvin) — fixed.** Confirmed the physics: K-index = (T₈₅₀+Td₈₅₀+Td₇₀₀) − T₅₀₀ − T₇₀₀ has 3 positive / 2 negative temperature terms, so a Kelvin computation carries a +273.15 offset — `k_index > 35` would be always-true and fire the corroboration bump for every realized ECMWF cell. Rather than guess the unit, I added a defensive normalization in the decoder that's correct either way: a value >100 is unambiguously Kelvin (a real K-index in °C never exceeds ~45) → subtract 273.15; °C passes through. Total Totals is offset-immune (terms cancel), left as-is. Test added (`test_ecmwf_k_index_kelvin_normalized`). This also removes the need to special-case it during your real-data validation.

**Min (altitude-filter fallback too narrow) — fixed.** Broadened to use the *deeper* of the active and thermo tops when the DD floor raised the grade — so a shallow non-None NWP top (your FL150-vs-FL350 case) no longer filters out a point graded HIGH by a deep DD EL. Test added.

**Min (string-based note routing) — fixed.** `_apply_firing_gate` now returns `(risk, note, is_driver)`; the caller routes by flag, not by matching "corroborated".

**Min (duplicate `_num`/`_raw` closures) — fixed.** Extracted a shared module-level `_opt_float` helper.

Thanks also for the explicit "checked and found correct" list (cp de-accumulation math, prev_cp step-0 tracking, CIN sign convention, etc.) — useful confirmation. Local suite green.

I think the substantive review surface is now exhausted across four rounds; the only remaining merge gate is the real-ECMWF-GRIB validation (now just confirming the shortnames decode — the K-index unit is handled either way).

---

PR #306, opened 2026-06-26
ECMWF cp fires NWP convective track without a tower; character prefers native cp

roznet:
## Why

Investigating EGTF→BIG→LFAT→LFQA 2026-06-27 (D-1): Windy showed vigorous convection over the Channel on ECMWF, but our briefing read it as benign (NWP convective track NONE, convective-character GREEN) while flagging Reims. Root cause — ECMWF's own scheme **is** precipitating convectively over the Channel (`cp` peaks **4.26 mm/h @ ~52 nm**, with 0.4–1.2 mm/h either side), but two §14 Phase-1/2 assumptions broke:

1. **"No tower top ⇒ quiet scheme ⇒ NONE."** ECMWF's `hcct` (convective top) is sentinel/absent across the whole Channel *even where `cp` fires*, and ECMWF has no convective-cover field. So `assess_convective_nwp` fell through to the quiet branch → NONE, with the `cp` we already decode sitting unused. The Phase-2 firing gate only ever *holds a tower down* — no symmetric path to *lift NONE up*. This is the marine/**elevated** convection case (surface parcel has zero CAPE, so the DD/parcel track is also quiet).
2. **Character "realized" read Open-Meteo `showers`, which is structurally 0.0 for ECMWF IFS** (verified: 0.00 at every point while total precip was 4.38 mm/h). So no ECMWF point could ever be "realized" by precip → GREEN.

## What

- **`cp` is now a first-class native firing signal** (`convective.py`). No tower top + no cover + `cp > 0.1` → tier from a convective-precip-rate ladder (`method="nwp_precip"`): ≥2.0 → MODERATE, ≥0.5 → LOW, ≥0.1 → MARGINAL. **Tower top stays primary whenever present** — ladder is the geometry-absent fallback only. Capped at MODERATE (depth unknown from rate); skips the firing-gate hold-down + precip-corroboration (would double-count `cp`). Rate is resolution-dependent → thresholds are a synoptic-scale v1.
- **Convective-character "realized" prefers GRIB-native `convective_precip_mm_h` over Open-Meteo `showers`** for any model carrying NWP diagnostics; Open-Meteo fallback only for non-GRIB models (AROME/UKMO/MF). The "use the GRIB variable, not Open-Meteo" principle.

## Effect (validated on the real pack)

- **NWP track over the Channel: NONE → fires** — 52 nm `cp` 4.26 → **MODERATE**, 72 nm → LOW, 42/62/82 nm → MARGINAL; tower-top points (10–22 nm) still `nwp_lcl_top` (geometry primary).
- **ECMWF convective character: GREEN → AMBER "Scattered cells" (20% of route)** under the shipped default method.
- **Graded severity colour over the Channel is unchanged** — §14's DD-floor (`max(native, thermo)`) already graded it MODERATE off elevated MU-CAPE. This is a narrative / character / track-independence fix, plus it lets the inline cross-check reflect the firing the model actually forecasts.
- **Safety asymmetry preserved:** `cp` can only *add* a firing signal, never downgrade a DD red.

## Tests / docs

- 3 new regression tests in `tests/test_convective.py`: precip-rate tiering, the ECMWF Channel case (`cp` fires when `hcct` absent → MODERATE), tower-top-stays-primary guard.
- Full suite: **2984 passed, 9 skipped, 0 failed**.
- Logged as meteorology-decisions **§14 Phase 3** (reasoning + the revisited Phase-1 "precip is yes/no only" stance).

> Note: visible effect reaches users only after a refresh recomputes packs — existing packs hold the pre-computed soundings.

### DIFF
--- a/designs/meteorology-decisions.md
+++ b/designs/meteorology-decisions.md
@@ -1384,6 +1384,79 @@ a yes/no firing gate.
 - `tests/test_convective.py` — native-top tiering, cover modifier + cap,
   cover-only scale, CIN suppression, quiet-native NONE, CAPE-fallback path, and
   the Reims regression anchors.
+
+### Phase 3 — `cp` fires the NWP track without a tower; character prefers native `cp`
+
+**Date:** 2026-06-26
+**Status:** Implemented.
+**Context:** EGTF→BIG→LFAT→LFQA 2026-06-27 (D-1). Over the Channel, ECMWF's own
+scheme is precipitating convectively — `cp` peaks **4.26 mm/h at ~52 nm**, with
+0.4–1.2 mm/h either side — matching what Windy shows. But the NWP track read
+**NONE** the whole way across, and the convective-character advisory read ECMWF
+**GREEN**. Two Phase-1/2 assumptions broke here:
+
+1. **"No tower top ⇒ quiet scheme ⇒ NONE."** Phase 1 made `convective_top_ft`
+   (`hcct`) the primary native scale on the premise that a firing scheme emits a
+   tower top. ECMWF violates it: `hcct` is sentinel/absent across the entire
+   Channel *even where `cp` = 4 mm/h*, and ECMWF has no convective-cover field.
+   So `assess_convective_nwp` fell through to the quiet branch → NONE, with the
+   `cp` we already decode sitting unused on the same diagnostics object. The
+   Phase-2 firing gate only ever *holds a tower down*; there was no symmetric
+   path to *lift NONE up* when `cp` fires without geometry. This is the marine /
+   **elevated** convection case — the surface parcel has zero CAPE (DD quiet),
+   so neither track saw it.
+2. **Character "realized" read Open-Meteo `showers`, which is structurally 0.0
+   for ECMWF IFS.** `showers_at_point()` returns Open-Meteo's convective-only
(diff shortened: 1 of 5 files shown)


roznet:
Thanks — addressed in 14fd7ed8.

**Important (CIN suppression on `nwp_precip`) — fixed.** Guarded the strong-CIN block with `if method != "nwp_precip":`. `cp` proves the scheme fired, so penalising it on surface/ML CIN is circular (the parameterisation already overcame the elevated inhibition, which a negative surface/ML CIN doesn't describe). Added `test_nwp_precip_not_suppressed_by_strong_cin` covering both a strong negative ML-CIN (−300, the preferred source) and a strong negative surface CIN — both now leave the precip-derived MODERATE tier intact. Good catch: the prior test sidestepped it via `cin_surface = 0`.

**Minor (mm/h vs mm at `showers = native_cp`) — noted in code.** Added a comment recording the mm/h ≈ mm equivalence for the hourly step and a flag to revisit if `showers_mm` is ever recalibrated. No behavioural change (numerically equal for hourly data).

**Minor (threshold comment at 0.1) — reworded.** `# MARGINAL lower bound (entry gate requires precip > 0.1)`.

Full `tests/test_convective.py` green (126 passed). Nothing deferred.

Issue #311, opened 2026-06-28
Sync (i) help/info content to the iOS/iPad app from the web app


roznet:
## Motivation

The web app has rich "(i)" info popups explaining metrics, advisories, and layers. We want the same help content in the iOS/iPad app **without maintaining a copy** of the text on the client — the web app stays the single source of truth.

Help must render **offline** (it's reference text; no point requiring a live connection per popup). Approach: the app downloads the latest help catalog when it connects, caches it to disk, and renders popups from that cache. A bundled baseline covers first-run-before-network.

## Current state (web app — for reference)

Help content today lives in three places:

1. **Metrics (63 entries, ~100 KB)** — `web/ts/data/metrics-catalog.json`, keyed by metric id. Entry shape: `name, unit, vibe, primary_goal, best_used_for, limitations, theory, wikipedia, thresholds[], llm_prompt`. **Currently bundled into the JS bundle only — NOT served by any API** (imported directly in `web/ts/helpers/metrics-helper.ts`).
2. **Advisories (~22)** — derived in `src/weatherbrief/analysis/advisories/*.py` via each evaluator's `catalog_entry()`, registered in `registry.py:get_catalog()`. **Already served**: `GET /user/preferences/advisories/catalog` (`src/weatherbrief/api/preferences.py`). Shape: `id, name, short_description, description, category, altitude_dependent, parameters[]`.
3. **Layers / fronts / Hewson** — in code + i18n strings (`layer-legends.ts`, `panel.ts`, `fronts-info.ts`, `hewson-metrics-catalog.ts`). Messier; **out of scope for v1**.

## i18n decision

The web app is localized into `en/fr/de/es` (`web/ts/i18n/locales/*.json`). Advisory and layer help go through the `t()` translation system. **The metrics catalog help (`metrics-catalog.json`) is English-only** — it has no language dimension, so metric (i) popups are already shown in English even to non-English users. This is a known existing gap, not introduced by this work.

Decision for this issue:

- The endpoint **accepts a `lang=` param and localizes advisory content** (mirrors current web behavior).
- The **metric catalog is returned in English regardless of `lang`** (matches current web behavior).
- The `lang` contract is **wired now** so that when the metrics catalog is eventually translated, no client or endpoint change is needed — the localized strings just start flowing.
- Localizing `metrics-catalog.json` into fr/de/es is a **separate future follow-up** (benefits web and iOS at once; out of scope here).

### Backend (`src/weatherbrief/`)

[...]

Issue #285, opened 2026-06-23
iOS app modernisation (tracking epic)


roznet:
Tracking epic for the iOS/iPad app modernisation.

Design of record: `designs/future/ios-app-modernisation.md` — §8 is the sequencing.

**Model:** supervised autonomy in vertical chunks — each phase is a runnable slice ending in a human checkpoint (build + simulator + screenshots + `/code-review`). The web app is the porting reference.

## Phases (dependency-ordered)
- [x] Phase 0 — Data plumbing (iOS-only) · #286
- [x] Phase 1 — Cross-section render parity · #287
- [x] Phase 2 — Backbone (nav + state + chrome + skin) · #288
- [x] Phase 3 — Cross-section interaction · #289
- [x] Phase 4 — Skew-T feature pass · #290 (pkg: roznet/rzskewt#1)
- [x] Phase 5 — Brief + advisory detail (+ backend endpoint) · #291
- [x] Phase 6 — Map tab · #292
- [x] Phase 7 — Flight management · #293

Only backend touch in the whole plan: Phase 5's one small REST endpoint. Phase 4 package work: roznet/rzskewt#1.

PR #317, opened 2026-06-28
ios(tips): native TipKit contextual coachmarks for the briefing detail views (#312)

### LINKED ISSUE #312: iOS: native TipKit contextual tips for the briefing detail views (download/refresh/cross-section)
## Motivation

The web app has a guided **tour** (driver.js) that spotlights buttons and explains the UI. We want a native equivalent on iOS/iPad — but rather than port a forced linear walkthrough, use **TipKit** (Apple's first-party contextual-tip framework, iOS 17+; we target iOS 26.2). TipKit fits our cockpit "non-blocking" constraint: tips surface at the point of use, the user dismisses them, and the framework remembers — no modal, no scripted sequence required.

Goal of *this* issue: establish the **pattern + scaffolding** and ship a **handful of example tips**, then add more incrementally as we go.

This is **not** the same as #311 (sync the deep "(i)" reference help catalog from the web app). These tips are short "what does this button do" coachmarks. They can later *deep-link into* the #311 help catalog from a tip's action button, but the two are independent.

Part of the iOS modernisation epic #285.

## Concept: how TipKit works

Four pieces:

1. **Configure once at launch** (in the `@main` App init or `AppState` setup):
   ```swift
   try? Tips.configure([
       .displayFrequency(.immediate),         // revisit .daily later if tips feel noisy
       .datastoreLocation(.applicationDefault)
   ])
   ```
2. **Each tip is a `Tip` struct** — title, message, optional icon, optional action buttons and eligibility `rules`.
3. **Attach to a view** with one of two presentations:
   - `.popoverTip(_:)` — a floating callout that points at a specific control. Our tool for "highlight this button."
   - `TipView(_:)` — an inline banner card embedded in layout. Better for explaining a whole area.
4. **Invalidate** when the user performs the action (`tip.invalidate(reason: .actionPerformed)`), so it never nags again. Persistence is automatic — no storage code.

**Sequencing (optional):** TipKit shows any eligible tip immediately. To get a soft "do this one first, then that one" progression, gate a tip behind a `#Rule` on an event the previous tip donates. It stays non-forced — if the user ignores one, the next simply waits.

**Known limitation (acceptable here):** `.popoverTip` anchors to the view it's attached to. You **cannot** cleanly anchor a popover to a `Tab` bar item in SwiftUI. That's fine — this issue scopes tips to **buttons/controls inside the detail views**, not the tab bar. (To "explain a tab," drop a `TipView` banner inside that tab's content instead.)

## Proposed structure

New group `app/flyfun-weather/flyfun-weather/Tips/`:
- `BriefingTips.swift` — one `Tip` struct per concept (small, declarative).
- `Tips.configure(...)` call added at app launch.
- Attachment points (`.popoverTip` / `TipView`) wired into the existing views below.

Each tip's copy should be short and action-oriented. Where a deeper explanation exists, the tip can carry an action button that opens the relevant config sheet or (later) the #311 help entry.

## Example tips to ship in this issue

All view references are in the worktree branch `feat/ipad-briefing-layout`.

- Attach `.popoverTip` to the `arrow.down.circle` button (the `.notDownloaded` case).
- `invalidate(.actionPerformed)` after `downloadCurrentPack()` succeeds.

- Attach `.popoverTip` to the `arrow.clockwise` button.
- Sequence after the download tip (donate event on download-tip dismissal) so the two read as a pair — they sit side by side and the contrast (offline-save vs. fetch-new) is the whole point.
- `invalidate(.actionPerformed)` after `refresh()`.

- Attach `.popoverTip` to the `layersPill` button (Label "Layers", `slider.horizontal.3`).
- Gate eligibility on the Cross-Section tab being visible (so it doesn't fire on the Advisory tab).
- `invalidate(.actionPerformed)` when `showingConfig` is opened.

- A `TipView` banner above the canvas (or a `.popoverTip` on the canvas frame) the first time the Cross-Section tab is opened.
- `invalidate(.actionPerformed)` once the user uses the scrub gesture once.

## Implementation plan

1. Add `Tips.configure(...)` at launch (guarded `try?`).
2. Create `Tips/BriefingTips.swift` with the four `Tip` structs above + the two events for the download→refresh sequence.
3. Wire the four attachment points and their `invalidate(...)` calls into the existing views (no structural changes to those views).
4. Build + run on iPad; verify each tip appears once, dismisses, and stays gone across relaunch. TipKit ships `Tips.showAllTipsForTesting()` / `Tips.resetDatastore()` to make manual QA easy.
5. (Follow-up, not this issue) add more tips incrementally — Skew-T variable side panel, Map metric picker, Advisory cards, etc. — and consider wiring a tip action button into the #311 help catalog once that lands.

## Out of scope

- Forced/linear walkthrough (TipKit is intentionally non-modal; if a true spotlight tour is ever wanted, that's a separate custom-overlay issue).
- Anchoring popovers to the tab bar.
- The deep reference help catalog (#311).
- Localization of tip copy beyond English for v1 (mirror the existing metrics-catalog English-only stance; revisit with #311's `lang` work).

roznet:
## What

Native iOS/iPad equivalent of the web app's guided tour, using Apple's **TipKit** (iOS 17+) instead of a forced linear walkthrough. Tips surface at the point of use, the user dismisses them, and the framework persists the seen-state — non-blocking, cockpit-friendly. Establishes the pattern + scaffolding and ships **four example tips**; more are added incrementally.

Closes #312. Part of the iOS modernisation epic #285.

## Changes

- **Launch config** (`WeatherBriefApp.swift`) — `Tips.configure([.displayFrequency(.immediate), .datastoreLocation(.applicationDefault)])` in a guarded `try?`.
- **`Tips/BriefingTips.swift`** (new) — four `Tip` structs + the download→refresh sequencing event. The two cross-section tips carry a `@Parameter` visibility gate.
- **Four attachment points** wired into existing views (no structural changes):
  1. **Download button** → `.popoverTip` "Save for offline".
  2. **Refresh button** → `.popoverTip` "Get the latest forecast", **sequenced after** the download tip. A `statusUpdates` watcher donates the sequencing event the moment the download tip is dismissed (closed via × *or* downloaded), so the offline-save / fetch-new pair reads in order.
  3. **Cross-section "Layers" pill** → `.popoverTip` "Choose what to show"; gated to the Cross-Section tab; invalidated when the config sheet opens.
  4. **Cross-section canvas** → inline `TipView` "Tap or drag anywhere"; gated to the tab; invalidated on first scrub.
- **Settings** (`SettingsView.swift`) — DEBUG-only "Developer" section with **Show All Tips** / **Reset Tips Override** (`showAllTipsForTesting` / `hideAllTipsForTesting`) for visual QA, mirroring the existing debug server picker. (`resetDatastore()` can't run post-`configure()`, so the session overrides are the correct runtime tools.)

## Testing

- Builds clean for iPad Pro 11" (M5) simulator, iOS 26.
- Verified on iPad simulator: "Save for offline" → close → "Get the latest forecast" sequences correctly; Cross-Section "Layers" tip appears on that tab only.

## Out of scope (per issue)

Forced/linear tour, tab-bar popover anchoring, the #311 deep help catalog, and tip-copy localization beyond English.

### DIFF
--- a/app/flyfun-weather/flyfun-weather/App/WeatherBriefApp.swift
+++ b/app/flyfun-weather/flyfun-weather/App/WeatherBriefApp.swift
@@ -1,9 +1,20 @@
 import SwiftUI
+import TipKit
 
 @main
 struct WeatherBriefApp: App {
     @State private var appState = AppState()
     @Environment(\.scenePhase) private var scenePhase
+
+    init() {
+        // Contextual coachmarks for the briefing detail views (#312). Guarded
+        // `try?` — a tip-store failure must never block app launch. Revisit
+        // `.daily` later if the tips ever feel noisy.
+        try? Tips.configure([
+            .displayFrequency(.immediate),
+            .datastoreLocation(.applicationDefault),
+        ])
+    }
 
     var body: some Scene {
         WindowGroup {
--- a/app/flyfun-weather/flyfun-weather/Tips/BriefingTips.swift
+++ b/app/flyfun-weather/flyfun-weather/Tips/BriefingTips.swift
@@ -54,7 +54,9 @@ struct RefreshBriefingTip: Tip {
 /// so it never fires while the Advisory (or any other) tab is on screen.
 struct CrossSectionLayersTip: Tip {
     /// Set from `CrossSectionView`'s appear/disappear so the tip is eligible
-    /// only while that tab is the visible one.
+    /// only while that tab is the visible one. TipKit requires each `@Parameter`
+    /// to live on its own Tip type, so this is written in tandem with
+    /// `CrossSectionScrubTip.crossSectionVisible` — update both together.
     @Parameter static var crossSectionVisible: Bool = false
 
     var title: Text { Text("Choose what to show") }
@@ -71,6 +73,8 @@ struct CrossSectionLayersTip: Tip {
 /// the canvas the first time the tab is opened; invalidated once the user
 /// scrubs once. Shares the visibility gate with the Layers tip.
 struct CrossSectionScrubTip: Tip {
+    /// Written in tandem with `CrossSectionLayersTip.crossSectionVisible` from
+    /// `CrossSectionView`'s appear/disappear — update both together.
     @Parameter static var crossSectionVisible: Bool = false
 
     var title: Text { Text("Tap or drag anywhere") }
--- a/app/flyfun-weather/flyfun-weather/Tips/BriefingTips.swift
+++ b/app/flyfun-weather/flyfun-weather/Tips/BriefingTips.swift
@@ -1,0 +1,84 @@
+import SwiftUI
+import TipKit
+
+/// Contextual coachmarks for the briefing detail views (#312).
+///
+/// TipKit (iOS 17+) surfaces each tip at its point of use, then persists the
+/// dismissal so a tip never nags twice — no storage code on our side. This is
+/// the native, non-blocking counterpart to the web app's driver.js tour: tips
+/// appear next to the control they describe, the user dismisses them, and the
+/// framework remembers. There is no forced linear walkthrough.
+///
+/// Scope (per #312): buttons/controls *inside* the detail views, not the tab
+/// bar (`.popoverTip` cannot cleanly anchor to a SwiftUI `Tab`). Add more tips
+/// incrementally — keep each one short and action-oriented.
+
+// MARK: - Events
+
+/// Donated when the user acts on the download tip, so the refresh tip can
+/// sequence *after* it. The two controls sit side by side and the whole point
+/// is the contrast — offline-save vs. fetch-new — so they should read as a pair.
+enum BriefingTipEvents {
+    static let downloadTipSeen = Tips.Event(id: "briefing.downloadTipSeen")
+}
+
+// MARK: - Toolbar tips
+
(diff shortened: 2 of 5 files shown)


roznet:
Addressed all three minors from the review (rebased onto current `main` first):

- **Refresh tip invalidation** — now guarded on a successful refresh; a `.error` result no longer dismisses the coaching tip, so the user can retry with the coaching still present (mirrors the download-tip guard). Confirmed success states are `.completed`/`.noRefresh`.
- **Scrub `TipView`** — gated on `csVM.vizData != nil`, so it won't be consumed coaching "tap anywhere" while the canvas is still loading or in an error/no-data state.
- **Duplicated `@Parameter`** — added cross-reference doc comments on both `crossSectionVisible` parameters noting they're written in tandem.

Build green for iPad Pro 11" (M5) sim. Pushed as `480714da`.

roznet:
Round-2 fixes pushed:

- **Important (landscape tip-burn)** — `updateScrub` now invalidates the scrub tip only when `!isLandscapeFocus`, so a landscape-focus drag can't consume a tip that's never rendered there. The tip survives for the portrait `TipView` where it's actually shown.
- **Minor (per-tick invalidate)** — added a one-shot `@State scrubTipInvalidated` guard so it fires exactly once.

Build green for iPad Pro 11" (M5) sim.

Issue #318, opened 2026-06-28
iOS XCUITest: remaining journeys + briefing fixture set (follow-up to #314)


roznet:
## Goal

Finish the XCUITest journey set started in #314 (Part 2). The launch seam and the first two journeys are merged via the PR for #314; this issue covers the remaining three journeys, which all depend on a **briefing fixture set** that doesn't exist yet.

**Prerequisite already in place (from #314):**
- Launch seam in `AppState` — `FLYFUN_UITEST=1` (skip auth gate, fake JWT on the observable mirror only) and `FLYFUN_MOCK=1` (serve `FixtureBriefingRepository` instead of the network). Both `#if DEBUG`.
- `FixtureBriefingRepository` (DEBUG, app target) — serves `flights()` + a stateful `createFlight`; collection endpoints return empty; **every briefing endpoint currently throws `FixtureError.notProvided`**.
- `accessibilityIdentifier`s so far: `flightList`, `flightCard-<id>`, `addFlightButton`, `waypointsField`, `submitFlightButton`.
- UI tests live in `app/flyfun-weather/flyfun-weatherUITests/flyfun_weatherUITests.swift`; `revealFlightList(_:)` handles the iPad-portrait split-view collapse (taps "Show Sidebar").

## The core task: build the briefing fixture set

The three remaining journeys all open a flight's briefing, so `FixtureBriefingRepository` must serve a coherent pack for at least one fixture flight (`fixture-1`). Wire up these endpoints (replace the `notProvided` throws) with canned, internally-consistent data:

- [ ] `latestPack` / `packs` → one `PackMetaResponse` (a green or amber assessment, recent `fetchTimestamp`, `daysOut`)
- [ ] `advisories` → `AdvisoriesResponse` with a few advisories incl. **one RED** (e.g. convective) carrying a `per_model` split + `cross_check` note, so the drill-down has something to show
- [ ] `advisoryDetail` → `AdvisoryDetailResponse` for that advisory (the "why it's RED" data)
- [ ] `digest` → `DigestResponse`
- [ ] `snapshot` → `SnapshotResponse` (route + waypoints; reuse the shape from the RouteMapViewModel unit-test fixture)
- [ ] `routeAnalyses` → `RouteAnalysesResponse` (enough points for the cross-section to render)
- [ ] `elevation` → `ElevationResponse`

[...]

Issue #315, opened 2026-06-28
iOS testing: wire unit + XCUITest into a release gate (Test Plan + CI + archive gate)


roznet:
## Goal

Wire the iOS test suites into a single, runnable release gate so every release automatically runs unit + UI tests across iPhone and iPad and blocks on failure. This is the "make it automatic" layer on top of the tests themselves.

**Blocked by #314** — the unit + XCUITest suites (and the `MockBriefingRepository` + launch-arg/mock seam) must exist and be green first. Start this once #314's tests run locally.

## Scope

- [ ] **Single Test Plan** (`AllTests.xctestplan`) containing:
  - the unit suite — runs on every configured device, fast;
  - the UI suite — iPhone + iPad configs (incl. iPad landscape).
- [ ] Reference the Test Plan from the **shared scheme** so it's the default ⌘U action (and lives in `xcshareddata` so it's checked in, not per-user).
- [ ] **Headless CI command** verified to run clean:
  ```bash
  xcodebuild test -scheme flyfun-weather -testPlan AllTests \
    -destination 'platform=iOS Simulator,name=iPhone 16' \
    -destination 'platform=iOS Simulator,name=iPad Pro 11-inch (M4)'
  ```
  (UI configs run with `FLYFUN_UITEST=1` / `FLYFUN_MOCK=1` so no backend/OAuth is needed — deterministic + offline.)
- [ ] **Gate the archive/release flow** (the `/archive` skill / release checklist) on this command passing — release stops on red.
- [ ] Decide + document the **destination matrix** (which simulators/OS versions are the canonical gate vs. occasional broader runs) and pin OS where reproducibility matters.
- [ ] Confirm the run is **invocable from the Claude Code Xcode MCP** (`RunAllTests` / `RunSomeTests`) so the suite can be triggered from a session, not only the command line.

## Nice-to-have (fold in if cheap)

- [ ] Surface the per-device launch screenshots + accessibility-audit results as build artifacts for a quick human eyeball each release.
- [ ] Note in the release checklist that native tooling has **no image-diff** — visual regression is assertion-based + human-reviewed screenshots by design.

## Definition of done

One command (and one MCP call) runs the full unit + UI matrix on iPhone and iPad, is the scheme's default test action, and the release/archive path refuses to proceed if it fails.

roznet:
Closing — the practical release-gate goal is covered by what we have:

[...]

Issue #314, opened 2026-06-28
iOS test suite buildout: unit (Swift Testing) + XCUITest (iPhone & iPad)


roznet:
## Goal

Build an automated regression net for the iOS/iPadOS app so we can release with confidence that nothing silently broke. This issue covers **implementation of the tests themselves** — both the unit suite and the XCUITest suite. CI wiring / release-gating is split into a follow-up (see the linked issue) so this one can land and prove out the test code first.

Native Apple tooling only (Swift Testing + XCTest + XCUITest). No third-party deps.

## Where we are today

- **Unit tests (`flyfun-weatherTests`)** — real, well-written, but cover only **JSON decode/plumbing**: `Phase0DataPlumbingTests` (extended sounding fields, optional-field tolerance, cross-check/parameters) and `Phase4SkewTWiringTests` (thermo indices + altitude→pressure interpolation). They validate the wire format only. None of the app's actual behaviour — view models, caching, the color/geometry math that draws every chart — is tested.
- **UITests (`flyfun-weatherUITests`)** — pure Xcode boilerplate (`testExample` asserts nothing). Two real blockers:
  - **Zero `accessibilityIdentifier` in the entire view layer** → any XCUITest keys off visible/localized text and is brittle.
  - **Whole app is behind a hard auth gate** (`isAuthenticated ? FlightListView : LoginView`, no guest mode). DEBUG "Dev Login" hits `/auth/dev-token`, but there's **no launch-argument seam** to inject a token or mock data, so a UI test can't get past login without a live backend + OAuth.

## Key asset we already have

Every ViewModel takes `repository: any BriefingRepository` — a `Sendable` **protocol**, not a concrete client. That's the seam: a `MockBriefingRepository` returning canned fixtures lets us unit-test all six ViewModels with **no network and no refactor**. (`APIClient` itself has no protocol and wraps `URLSession` directly — leave it at the integration edge, don't unit-test it.)

---

# Part 1 — Unit tests (Swift Testing, fast, no simulator UI)

Use `@Test`/`#expect` to match the existing style. Test boundary values, not just happy paths.

### Tier A — pure functions (highest value-per-minute; these draw every chart, regressions here = wrong colors/positions with no crash)
[...]

PR #319, opened 2026-06-28
iOS test suite: unit (Tier A/B/C) + XCUITest seam & first journeys (#314)

roznet:
## Summary

Builds out the iOS app's automated test suite (#314): a full unit-test layer plus the XCUITest launch seam and the first two UI journeys. ~51 unit tests + 2 UI journeys, all green on iPhone 17 and iPad Pro 11.

**Part 1 (unit) — complete. Part 2 (XCUITest) — seam + journeys 1–2.** The remaining three journeys (briefing drill-down, cross-section, offline), which all need a coherent briefing fixture set, are split into **#318** so this PR stays reviewable.

## Part 1 — unit tests (`flyfun-weatherTests`)

- **Tier A — pure logic:** `CoordTransform` (axis mapping, inverted-Y, round-trip), `Assessment`/`severityRank`, `Extensions`, `MapColors` (ceiling/SFIP/risk threshold buckets + continuous-scale clamps), `MapMetrics` (`linearWidth`, segment aggregation, and the private altitude helpers via the public registry + a `VizPoint` stub).
- **Tier B — ViewModels** via a new `MockBriefingRepository`: `AddFlightViewModel` (waypoint parse, `canSubmit`, change detection, ICAO regex), `FlightListViewModel` (state machine), `RouteMapViewModel` (fit-region math), `BriefingViewModel` (pack labels).
- **Tier C — cache/offline:** `CachedPackEntry` rule, `BriefingCacheStore` (index/persistence/data round-trips), `PirepOfflineStore` (enqueue/sync/persist).

**Colour thresholds are tested by bucket identity** (same-bucket equal, cross-boundary unequal) so they survive palette tweaks and only fail if a threshold moves.

## Part 2 — XCUITest

- **Launch seam** (`AppState`, `#if DEBUG`): `FLYFUN_UITEST=1` skips the auth gate (fake JWT on the observable mirror only — never written to the keychain); `FLYFUN_MOCK=1` serves `FixtureBriefingRepository` instead of the network — deterministic, offline, no backend/OAuth.
- **`FixtureBriefingRepository`** (DEBUG, app target): serves `flights()` + a stateful `createFlight`; briefing endpoints throw `notProvided` until #318 wires them up.
- **`accessibilityIdentifier`s:** `flightList`, `flightCard-<id>`, `addFlightButton`, `waypointsField`, `submitFlightButton`.
- **Journey 1** — flight list renders seeded fixtures.
- **Journey 2** — add-flight: submit gated until ≥2 waypoints, create, new flight appears.
- Both journeys pass on **iPhone + iPad**; `revealFlightList` handles the iPad-portrait split-view collapse (taps "Show Sidebar").

## Minimal production changes (all test-motivated)
- `BriefingCacheStore` / `PirepOfflineStore`: added designated `init(cacheDir:)` / `init(fileURL:)` overloads (production `init()` delegates) so the on-disk actors test against a temp dir.
- `AppState`: the `#if DEBUG` UITest seam.
- `FixtureBriefingRepository`: new DEBUG-only file.
- `accessibilityIdentifier`s on a few controls. No behaviour change.

## How files are added
The project uses Xcode 16 synchronized folder groups, so new test/source files compile automatically — no `project.pbxproj` edits.

## Verification
- Unit suite: `** TEST SUCCEEDED **` on iPhone 17.
- UI journeys: green on iPhone 17 **and** iPad Pro 11.
- Rebased onto latest `origin/main` (app-icon + #311 help-content commits); re-ran unit + UI post-rebase — still green.

## Deferred (tracked in #318)
- Briefing/cross-section/offline journeys + the briefing fixture set.
- `CrossSectionViewModel.extractVizData` unit test.
- `CachingBriefingRepository` fallback test (needs a protocol-injection refactor of the concrete `OnlineBriefingRepository`).

Part of #314. Remaining journeys: #318. CI/release-gate wiring: #315.

### DIFF
--- a/app/flyfun-weather/flyfun-weather/App/AppState.swift
+++ b/app/flyfun-weather/flyfun-weather/App/AppState.swift
@@ -122,7 +122,9 @@ final class AppState {
         if Self.isUITestMockMode {
             repository = FixtureBriefingRepository()
         } else {
-            applyToken("uitest-token")
+            // Live-dev mode: don't write to the keychain — reuse whatever real
+            // token the developer already has, so we never leave a fake token
+            // behind that would 401 the real app on next launch.
             setupClient()
         }
     }
--- a/app/flyfun-weather/flyfun-weather/App/AppState.swift
+++ b/app/flyfun-weather/flyfun-weather/App/AppState.swift
@@ -76,6 +76,16 @@ final class AppState {
 
     var isAuthenticated: Bool { jwt != nil }
 
+    #if DEBUG
+    /// Launched by a UI test (`FLYFUN_UITEST=1`) — skip the auth gate so the app
+    /// renders past `LoginView` without a real sign-in.
+    static var isUITesting: Bool { ProcessInfo.processInfo.environment["FLYFUN_UITEST"] == "1" }
+    /// Serve bundled fixtures instead of the network (`FLYFUN_MOCK=1`) so UI tests
+    /// are deterministic and offline. Stub at the app boundary — XCUITest can't
+    /// intercept network the way Playwright can.
+    static var isUITestMockMode: Bool { ProcessInfo.processInfo.environment["FLYFUN_MOCK"] == "1" }
+    #endif
+
     // MARK: - Lifecycle
 
     init() {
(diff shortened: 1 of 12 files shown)


roznet:
Addressed the review (commit efac81d0). All five findings fixed:

**Important**
- `revealFlightList` now keys the early-exit off the `flightList` accessibility identifier instead of the fixture route text — renaming a fixture no longer breaks iPad split-view handling, and the iPhone fast-path is preserved (verified: list journey still ~9.7s, not the +5s a broken early-exit would add).
- The non-mock UITest path no longer calls `applyToken` — it never writes the fake token to the keychain, so the docstring is accurate and the real app won't 401 on next launch.
- `FixtureBriefingRepository.decodeFlights` uses `try!` so a model change that breaks the fixture fails loudly instead of silently emptying the flight list.

**Minor**
- `makePirepRequest` sets `remarks` on the decoded value instead of interpolating caller content into raw JSON.
- `PirepOfflineStoreTests` — each test owns a temp dir with `defer` cleanup; no more orphaned `wbtests-<UUID>` dirs.

Verified: unit + UI suites green on iPhone 17 and iPad Pro 11. Branch was also rebased onto current `main` (TipKit #312) before these fixes — clean, no conflicts. Nothing deferred.

Issue #320, opened 2026-06-28
iOS: cross-section color Theme system (separate from layer Presets)


roznet:
## Context

The iOS cross-section currently has **no color-theme system** — every color is a static constant in `app/flyfun-weather/flyfun-weather/Views/CrossSection/ColorScales.swift` (sky, clouds, icing, CAT, convective, terrain, temperature/stability/reference lines, inversion). The on-device "Preset" picker (GRAMET/Windy/ForeFlight) only swaps *cloud style*; it changes no colors.

The web app already distinguishes two orthogonal concepts (see `designs/visualization.md` → "Cross-Section Theme System" and "Preset System"):

- **Theme** = colors. `web/ts/visualization/cross-section/theme.ts` defines a `CrossSectionTheme` interface and a `THEMES` map with three entries: `standard` (light blue sky #7395DB), `high-contrast` (dark navy #1B3060), `gramet` (deep blue #2B5DA8, CloudPath-inspired). Switched via `getActiveTheme()`/`setActiveTheme(id)`; renderers listen for a `theme-changed` event to re-render. Legends are theme-aware (`layer-legends.ts`).
- **Preset** = layer config. `PRESETS` in `layer-registry.ts` are `{ id, label, themeId, enabledLayers }` — each preset also selects a theme.

This issue covers ONLY the Theme (colors) half. The layer-Preset + advisory-preset work (and items 1,2,4,5,7,8) are handled in the "ios ui tweaks" branch and do NOT depend on this — they leave a single hardcoded color scheme in place.

## Goal

Introduce a cross-section color-theme system on iOS with the same three themes as the web (`standard`, `high-contrast`, `gramet`), a Theme selector in `CrossSectionConfigSheet`, and have the layer-Presets carry a `themeId` so selecting GRAMET/Windy/ForeFlight also sets the matching theme (web mapping: gramet→gramet, windy→light/standard, foreflight→high-contrast).

## Source of truth: inherit from web where possible

There is no shared runtime config — the web themes are bundled TypeScript. **Port the color values verbatim from `web/ts/visualization/cross-section/theme.ts`** into a Swift `CrossSectionTheme` struct + `THEMES` registry, mirroring how `CrossSectionLayer.swift` already mirrors the web layer registry (keep the "mirrors web X" comments and matching IDs so the two can be diffed). Do not invent new palettes; match the web so a route looks identical on both platforms.

## Implementation sketch

[...]

PR #321, opened 2026-06-28
iOS: briefing & cross-section UI tweaks

roznet:
Nine UI tweaks to the iOS briefing viewer + cross-section, ported to match web behaviour where relevant. Color themes for the cross-section are split out into #320 for an independent agent.

## Changes
1. **Pack picker** — chip shows the run time (`D-0 · 14:05`) when another pack shares the same UTC day, so same-day runs are distinguishable (menu rows already showed full time).
2. **Advisory model badges** — `FlowLayout` instead of an evenly-spaced `LazyVGrid`; badges pack tightly and wrap to a second row instead of stretching one row.
3. **Scroll-spy nav** — tap-to-scroll was a silent no-op: `.id()` wasn't the outermost modifier on the anchor, so `ScrollViewReader.scrollTo` couldn't find it (scroll-position highlight still worked, masking it). `.id()` now applied last.
4. **Watch** — moved to the end of the Advisory tab, rendered as **markdown** (was an unreadable one-line chip row); dropped the cross-section deep-link (watch items are just a list to keep an eye on).
5. **Discussion** — decode `specific_concerns` and render **Synoptic / Specific Concerns / Trend / Watch** as markdown (was synoptic-only).
6. **Cross-section presets** — layer presets now carry real per-preset `enabledLayers` ported from the web (GRAMET/Windy/ForeFlight); added the advisory-lens family (Basic/Icing/Clouds/Convective/Turbulence/VFR/IFR) ported from the web, selectable in the config sheet and via an "On cross-section" chip on advisory cards.
7. **Cloud controls** — split the single 6-option dropdown into two independent controls: **Source** (DD/NWP) and **Style** (Soft/Natural/Square).
8. **Terrain** — always renders (force-on in the renderer); removed its config toggle, mirroring the web panel.
9. **iPhone landscape** — cross-section collapsed to a vertical sliver because `.aspectRatio(nil, .fit)` uses a Canvas's (nonexistent) intrinsic ratio. Portrait keeps 2:1; landscape fills.

## Sync / duplication note (item 6)
There is no shared runtime config for presets — the web presets are bundled TypeScript. The tables are **ported into Swift** (`CrossSectionPresets.swift`), the same pattern already used for the layer registry, translated to iOS layer IDs. Both sides carry reciprocal `SYNC:` comments:
- `CrossSectionPresets.swift` ↔ `web/ts/visualization/cross-section/layer-registry.ts` (PRESETS), `advisory-presets.ts` (ADVISORY_PRESETS / mapping), `layers/cloud-bands-factory.ts` (cloud source×style).

## Verification
- `xcodebuild` (Debug, iOS Simulator) **succeeds**, no warnings.
- ⚠️ **Items 3 and 9 are code-fixed but not live-tested**: the fixture repository doesn't provide digest/advisories/routeAnalyses/elevation, so mock mode can't render the advisory scroll-spy or the cross-section. Both fixes are canonical SwiftUI patterns; please confirm on device (scroll-spy pill tap jumps to section; iPhone landscape cross-section fills width).

### DIFF
--- a/app/flyfun-weather/flyfun-weather/Models/API/DigestResponse.swift
+++ b/app/flyfun-weather/flyfun-weather/Models/API/DigestResponse.swift
@@ -6,6 +6,9 @@ struct DigestResponse: Codable, Sendable {
     let assessment: String?
     let assessmentReason: String?
     let synoptic: String?
+    /// Hazard-by-hazard concerns (digest `specific_concerns`). Decoded via the
+    /// client's `.convertFromSnakeCase` strategy. Markdown.
+    let specificConcerns: String?
     let winds: String?
     let icing: String?
     let turbulence: String?
@@ -54,6 +57,14 @@ struct DigestResponse: Codable, Sendable {
     /// Watch items as an array of strings.
     var watchItemsList: [String] {
         watchItems?.values ?? []
+    }
+
+    /// Watch items reconstructed as a single markdown string (the server sends
+    /// `watch_items` as one markdown block; `FlexibleStringArray` splits it on
+    /// newlines, so re-join to render the original list). Empty → nil.
+    var watchItemsMarkdown: String? {
+        let joined = watchItemsList.joined(separator: "\n")
+        return joined.isEmpty ? nil : joined
     }
 }
 
--- a/app/flyfun-weather/flyfun-weather/ViewModels/BriefingViewModel.swift
+++ b/app/flyfun-weather/flyfun-weather/ViewModels/BriefingViewModel.swift
@@ -50,6 +50,9 @@ struct FocusIntent: Equatable {
     var layerId: String?
     /// Map metric id to select.
     var mapMetricId: String?
+    /// Advisory lens id to apply on the cross-section (e.g. "icing"); see
+    /// `CrossSectionPresets`.
+    var advisoryPresetId: String?
     var pointIndex: Int?
     var distanceNm: Double?
     var altitudeFt: Double?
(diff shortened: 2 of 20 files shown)


roznet:
## Code review — PR #321 (iOS briefing & cross-section UI tweaks)

Single complete pass. Scope is iOS-heavy; the web changes are reciprocal `SYNC:` comments only (no CSP/runtime impact). Reviewed against the iOS-app-code and cross-platform-parity criteria.

**Verified clean:**
- **DTO decode safety** — `DigestResponse.specificConcerns` is `Optional`, there's no hand-written `CodingKeys` (synthesized), and `JSONDecoder.weatherBrief` uses `.convertFromSnakeCase`, so `specific_concerns` maps correctly and an omitting server can't break decode. Server emits the field (`digest/llm_digest.py`). 👍
- **Ported tables match web** — `ADVISORY_TO_PRESET` (9 keys), every lens's `groups`/`lines`, the cloud source×style → layer-id map, and the three preset `themeId`s all match the web sources exactly. The only dropped `lines` (`surface-obscuration-bands` on `vfr`, `sld-bands` on icing) are the documented iOS-missing layers — correct.
- **Concurrency/memory** — `CrossSectionViewModel` stays `@MainActor`; the new `static let` `DateFormatter`/`ISO8601DateFormatter`s are immutable and fine to share; button closures capture the reference-type view model but are owned by the view tree (no retain cycle).
- **#9 Canvas aspect** and **#3 scroll-spy** (`.id()` outermost + reader-wraps-only-the-ScrollView with an `onChange` trigger) are the right SwiftUI patterns.

### Important

**1. `CrossSectionPresets.preset(forAdvisory:)` drops the web's per-advisory overrides** — `CrossSectionPresets.swift`
The web resolves a card chip via `getPresetForAdvisory`, which **unions** `ADVISORY_OVERRIDES` onto the base lens. Today that map adds, for `fiki_icing`, `lines: ['minus-10c', 'minus-20c', 'sld-bands']`. The Swift port returns the bare `icing` lens with no merge, so a **FIKI-icing advisory's "On cross-section" chip won't enable the −10/−20 °C warm-nose isotherms** that web shows (iOS lacks `sld-bands`, but it *does* have `minus-10c`/`minus-20c`).

This is a silent divergence: the file's SYNC header lists `getPresetForAdvisory` as mirrored, and the known-gaps note covers missing *layers* but not this missing *override mechanism*. Fix either way:
- Port the override: add an `advisoryOverrides: [String: (groups:[LayerGroup], lines:[String])]` map and union it in `preset(forAdvisory:)` (mirrors web's union semantics, drops unknown ids as already done); **or**
- If intentionally out of scope, add `fiki_icing` warm-nose emphasis to the documented known-gaps comment so it's not a surprise.

### Minor

**2. `DigestMarkdownText.normalize` list-break heuristic can fire on sentence-final integers** — `DigestMarkdownText.swift`
`(?<=\S)[ \t]+(?=\d+\.\s)` correctly ignores decimals like `10.5`, but it still matches a normal sentence that ends in a bare integer (e.g. `"gusts to 25. Expect…"` → line-break before `"25. Expect"`). Low impact since digest watch/concerns are usually already-structured lists, but if you want to tighten it, require the marker at/near line start or anchor on a preceding sentence terminator. Fine to accept as-is.

**3. Parity process note — run `/sync-ios-web` before merge.** The layer-preset `enabledLayers` (GRAMET/Windy/ForeFlight) and the reset-group set are hand-ported with translated IDs. I verified themes, advisory lenses, the advisory→lens map, and the cloud-axis map by hand; the exhaustive per-layer `enabledLayers` diff against the web `*_ENABLED` constants is exactly what the new `/sync-ios-web` skill automates — worth a run to confirm no translated-ID drift (and it'll catch item 1 above too).

### Verdict
No blockers. Item 1 is a real but narrow behavioural gap (one advisory, two isotherm lines) — fix or document before merge; items 2–3 are optional/process. Nicely structured port with good SYNC discipline on both sides.

🤖 via `/code-review`

roznet:
Addressed the review (commit 90744a8d):

- **Blocker — advisory lens "None" no-op**: fixed. Added `clearAdvisoryPreset()` on the VM and the picker setter now handles the empty-string tag (`id.isEmpty → clearAdvisoryPreset()`), so "None" deselects the lens instead of snapping back.
- **Dead `defaultEnabled`**: removed (both uses already moved to `CrossSectionPresets.gramet`).
- **Empty CTA `HStack`**: extracted to a guarded `ctaRow` `@ViewBuilder` that only renders when a button is present.

Note on the flagged GRAMET style change (Soft NWP → Natural NWP): intentional — it's web-preset parity (the web GRAMET preset enables `nwp-cloud-bands`). Will call it out in the release note. Build passes.

roznet:
Addressed the iOS-focused review (commit 2426d874):

- **Important #1 — advisory overrides dropped**: ported web `ADVISORY_OVERRIDES` and union them in `preset(forAdvisory:)`, so FIKI-icing's "On cross-section" chip now enables the −10/−20 °C warm-nose isotherms like web. (`sld-bands` kept in the map for parity, dropped at apply since iOS lacks it.) SYNC header updated to list `ADVISORY_OVERRIDES`.
- **Minor #2 — list-break heuristic**: `normalize` now anchors on a clause terminator (`[.,:;)]`), so it won't break before a sentence-final integer ("gusts to 25. Expect…") or a range ("5 - 10 kt").
- **Minor #3 — parity**: ran `/sync-ios-web` — metrics-catalog identical, preset/lens/cloud-axis tables aligned (incl. the new override), DTOs aligned, SYNC reciprocal. No drift.

Build passes. The GRAMET Soft→Natural NWP default-style change remains intentional (web parity) and will be noted in the release notes.

PR #299, opened 2026-06-25
Implement iOS modernisation epic (#285)

downle:
## Summary
- Implements the iOS modernisation epic tracked by #285 across the phase commits for #286 through #293.
- Adds the modern iOS briefing/data backbone, cross-section parity, touch interaction, Skew-T overlays, advisory drilldown, map metric overlays, and flight management polish.
- Adds a follow-up iOS fix so dev users can add aircraft from the new-flight flow, create flights, and see precise API/SSE authorization errors.

## Verification
- Built the iOS app on the Mac with Xcode 26.3:
  `xcodebuild -project flyfun-weather/flyfun-weather.xcodeproj -scheme flyfun-weather -configuration Debug -destination "platform=iOS Simulator,id=A2381A35-36C7-490E-AADE-EEFE8A003588" -derivedDataPath /Users/qianzhao/DerivedData/flyfun-weather-epic-285-build build`
- Result: `BUILD SUCCEEDED`.

## Notes
- PR branch contains the committed epic/fix work only. Local uncommitted Xcode project/scheme edits in the NFS worktree were left out.

Closes #285.

PR #297, opened 2026-06-24
iOS app modernisation (epic #285) — Phases 0–7

roznet:
Implements the iOS/iPad app modernisation tracking epic **#285** end-to-end — one commit per phase, in dependency order. Design of record: `designs/future/ios-app-modernisation.md`.

Each commit compiles (`xcodebuild`, iPhone 17 sim) and tests pass. Only one backend touch in the whole series (Phase 5's REST endpoint).

## Phases
- **#286 Phase 0 — Data plumbing**: decode the 14 dropped `SoundingProfileLevel` fields (RH, θe, lapse, Ri, ω, CC, CLW, ICE, icing indices) + advisory `cross_check`; fixture decode tests.
- **#287 Phase 1 — Cross-section render parity**: improved **natural** cloud style (bit-faithful `hash01` port) + new **square** style + LCL/LFC/EL lines; shared `NaturalCloudConfig` mirrors the web `DEFAULT_NATURAL_CONFIG`.
- **#288 Phase 2 — Backbone**: 4-tab restructure (Brief·Cross-section·Skew-T·Map), shared state (`activePoint`/`focusIntent`), connective header, `Theme` skin tokens (web light+dark palette, both modes).
- **#289 Phase 3 — Cross-section interaction**: touch scrub + unified readout strip + "Sounding ›" deep-link, config sheet (preset/compact/legend), route-graph shared cursor, landscape immersive focus.
- **#290 Phase 4 — Skew-T app wiring**: feed LCL/LFC/EL pressures + CIN + lifted index to the existing RZSkewT marker/index renderer (alt→pressure fallback). _Package rendering (overlay bands, interactivity, side panel) deferred to roznet/rzskewt#1._
- **#291 Phase 5 — Brief + advisory detail (+ backend)**: the one backend touch — `GET …/advisories/{id}/detail` reusing `connectors/views.py` (advisory_detail + convective_detail), with tests. iOS "why it's RED" ladder (cross-check rendered in neutral chrome per the cardinal rule) + digest decomposed by role (synopsis last).
- **#292 Phase 6 — Map tab**: per-segment metric overlay (`MapMetrics` registry, colour+width), altitude slider, legend, waypoint conditions; native MapKit dark mode.
- **#293 Phase 7 — Flight management**: Future/Recent/Past logbook + native edit-flight (closes the web-only gap) with a re-briefing cost confirm.

## Verification
- Per-phase `xcodebuild` build on iPhone 17 simulator — all succeed.
[...]

PR #322, opened 2026-06-28
ios(cross-section): split static scene from scrub cursor overlay (#303)

### LINKED ISSUE #303: iOS: cross-section scrub redraws full layer stack per drag tick
From the PR #297 code-review (Important, deferred to a follow-up as it's a real refactor rather than a tweak).

**Problem:** `CrossSectionView.crossSectionCanvas` re-renders whenever `scrubDistanceNm` changes (every drag tick, ~60fps). The natural cloud style draws O(n_slots × subBlobsPerSlot) radial-gradient fills per band per segment — ~400+ gradient fills per tick on a typical route. The scrub cursor is a thin dynamic overlay over an otherwise static scene, but the single `Canvas` redraws everything each tick → jank on older A-series chips.

**Fix direction:** separate the static layer (cloud bands etc., rebuilt only on data/model/size change) from the dynamic cursor overlay. Options: render the static scene into a cached `ImageRenderer`/`UIImage` and overlay a lightweight SwiftUI cursor rule + aircraft marker; or at minimum `.drawingGroup()` the static canvas. The cursor (currently drawn inside the same Canvas via `selectedDistanceNm`) needs to move to its own overlay first.

Note: this pattern is shared with the alternative #299 implementation too — both drew the cursor in the layer Canvas.

roznet:
Fixes the cross-section scrub jank from #303 (deferred from the PR #297 review).

## Problem
`CrossSectionView`'s single `Canvas` re-rendered the **whole** layer stack whenever `scrubDistanceNm` changed — i.e. every drag tick (~60fps). The natural cloud style draws O(n_slots × subBlobsPerSlot) radial-gradient fills per band per segment (~400+ gradient fills/tick on a typical route). The scrub cursor is a thin dynamic overlay over an otherwise static scene, but everything repainted each tick → jank on older A-series chips.

## Fix
Separate the static layer from the dynamic cursor overlay, as the issue suggested:

- **`StaticCrossSectionScene`** (`View & Equatable`) — sky + all enabled data layers + axes/grid. The expensive pass. Its `Equatable` gate keys on a new monotonic `dataVersion` (bumped only in `CrossSectionViewModel.update`, i.e. model/route/elevation change) plus the layer dict, so SwiftUI skips re-running the `Canvas` closure on a scrub tick. `VizRouteData` is a deep value type with no `Equatable`, hence the cheap version int instead of diffing the struct each tick. A frame/size change still re-runs the closure (the `Canvas`'s own `size` input changes) — exactly the "rebuild on resize" behaviour we want.
- **`CrossSectionCursorOverlay`** — the scrub cursor rule + live aircraft marker. O(1), `allowsHitTesting(false)` so the scrub gesture on the static scene beneath still works. Redraws every tick — but it's now cheap.

`CrossSectionRenderer.render` is split into `renderStatic` / `renderCursor`; the combined `render` stays for any single-pass caller (previews/snapshots).

Note (from the issue): this cursor-in-the-layer-Canvas pattern was shared with the alternative #299 implementation too.

## Verification
- `xcodebuild` (generic iOS Simulator, Debug) — **BUILD SUCCEEDED**.
- No data-path/logic change beyond the `dataVersion` bump; no test references the renderer or the counter.
- Visual parity (static scene identical, cursor/aircraft unchanged) wants a device/simulator eyeball — flagged as a human checkpoint since mock mode can't render the cross-section.

Closes #303.

### DIFF
--- a/app/flyfun-weather/flyfun-weather/ViewModels/CrossSectionViewModel.swift
+++ b/app/flyfun-weather/flyfun-weather/ViewModels/CrossSectionViewModel.swift
@@ -9,9 +9,16 @@ final class CrossSectionViewModel {
     private(set) var enabledLayers: [String: Bool] = CrossSectionPresets.gramet
     /// Currently-applied advisory lens id (e.g. "icing"), or nil when none/Custom.
     private(set) var activeAdvisoryPreset: String?
+    /// Monotonic identity for `vizData`: bumped only when the data is rebuilt
+    /// (model/route/elevation change). `VizRouteData` is a deep value type with no
+    /// `Equatable` conformance, so the static cross-section scene keys its
+    /// `Equatable` redraw gate on this counter instead of diffing the whole struct
+    /// each scrub tick (#303).
+    private(set) var dataVersion: Int = 0
 
     func update(routeAnalyses: RouteAnalysesResponse, elevation: ElevationResponse?, model: String) {
         vizData = Self.extractVizData(from: routeAnalyses, model: model, elevation: elevation)
+        dataVersion += 1
     }
 
     func toggleLayer(_ id: String) {
--- a/app/flyfun-weather/flyfun-weather/Views/CrossSection/CrossSectionRenderer.swift
+++ b/app/flyfun-weather/flyfun-weather/Views/CrossSection/CrossSectionRenderer.swift
@@ -14,7 +14,20 @@ struct CrossSectionRenderer {
         let opacity: Double
     }
 
+    /// Convenience: render the whole scene in one pass (static scene + dynamic
+    /// cursor/aircraft overlay). The live view splits these into two separate
+    /// Canvases so a scrub tick redraws only the cheap overlay (#303); this stays
+    /// for any single-pass caller (previews, snapshots).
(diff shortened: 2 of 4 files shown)


roznet:
Acted on the review (round 1):

**Important — `.equatable()` missing → fixed** (commit fdf2a9a1). Good catch: the `Equatable` conformance alone wasn't gating anything, so the static scene would have redrawn every scrub tick and the split would've been a no-op. Added `.equatable()` at the call site and documented it as load-bearing in both the call site and the `StaticCrossSectionScene` doc comment so it isn't silently removed later. `xcodebuild` (generic iOS Simulator, Debug) still **BUILD SUCCEEDED**.

**Minor — multi-line comments "violate CLAUDE.md" → not actioned.** The quoted rule (*"Never write multi-paragraph docstrings or multi-line comment blocks — one short line max"*) isn't present in any CLAUDE.md in this repo or the global config (grepped all of them). The verbose "why" comments match the established idiom of the surrounding cross-section code, and they explain the genuinely subtle Equatable-gate mechanism — trimming would reduce consistency, not improve it. Deferring/declining rather than churning.

PR #324, opened 2026-06-28
ios: revamp flight-create flow (loading fix, autocomplete, TZ, profiles, autorouter, interpret)

roznet:
Brings the iOS flight-create flow to parity with the web flight-create page and fixes the stuck **"Loading Briefing"** screen. Almost entirely iOS client work — the web backend already had the endpoints; one new server endpoint was added for the autocomplete DB.

## What's in it

| # | Change |
|---|--------|
| **1** | **Stuck loading fixed.** Creating a flight only saved it (no briefing). `BriefingViewModel.loadBriefing()` now detects the no-pack `404` and routes into the existing streamed `refresh()`, so the progress banner (stage/detail/%) drives first-briefing generation instead of a dead spinner. |
| **6** | **Dynamic ICAO autocomplete** on the route field (debounced, last-token, never blocks typing), backed by RZFlight `KnownAirports.rankedSearch`. The airports DB is **downloaded**, not bundled: new `GET /api/nav/airports-db` serves a slim ~3.5 MB SQLite (airports+runways from `nav.db`), cached server-side, ETag-refreshed. iOS caches it in Application Support and refreshes on launch; with no copy (first run / offline) autocomplete is simply empty. **The DB is gitignored — never committed** (it churns every AIRAC cycle). |
| **3** | **Profile selector** (create + edit). `GET /api/user/profiles`; applies the profile's cruise altitude; sends `profile_id`. |
| **2** | **Import from Autorouter** (`/autorouter-routes`, reuses the FPL parse path; `409` → "link on the web app") + **recent-routes menu** (client-side from flight history). |
| **4** | **TZ-aware departure time**: date + hour/minute + **timezone dropdown** populated from route waypoints, DST-correct via `TimeZone.secondsFromGMT(for:)`. Also fixes a latent bug where a device-local `DatePicker` value was force-stamped `Z`. |
| **5** | **"Interpret route" sheet**: understood / skipped / off-route chips + a MapKit route polyline; confirm-on-create when tokens were dropped. Shares the one `interpret-route` call with #4. |

Edit parity applied throughout (same `AddFlightView`).

## Tests
- Swift: `DepartureTimeModelTests` (summer/winter DST instant math, tz-switch-preserves-instant, offset formatting) + `RouteAutocompleteTests` (last-token parsing) — pass.
- Python: `test_nav_airports_db.py` (slim-DB shape + ETag/304) — pass; existing flights-API tests still green.
- 6 clean `xcodebuild` builds. On-device visual verification done by the author.

## Notes / follow-ups
- Sharing: all new iOS code is app-local, reusing RZFlight's existing primitives. Promoting the pure TZ helpers + a `KnownAirports.bundled` loader into RZFlight (so flyfun-forms can drop its DST-buggy `TimeEntryView`) is a deliberate follow-up — both apps track rzflight `main`.
- Design notes: `designs/future/ios-flight-create-overhaul.md`.

### DIFF
--- a/app/flyfun-weather/.gitignore
+++ b/app/flyfun-weather/.gitignore
@@ -1,0 +1,4 @@
+
+# Slim airports DB is downloaded from the server at runtime, never committed (churns per AIRAC cycle)
+airports.db
+*.airportsdb
--- a/app/flyfun-weather/flyfun-weather/App/AppState.swift
+++ b/app/flyfun-weather/flyfun-weather/App/AppState.swift
@@ -237,5 +237,9 @@ final class AppState {
         repository = CachingBriefingRepository(client: client, online: online, cache: cache)
         Task { await userPreferences.refresh(using: client) }
         Task { await helpCatalog.refresh(using: client) }
+        // Open any cached airports DB immediately (offline-safe), then refresh it
+        // from the server if the AIRAC copy changed (ETag → usually a 304 no-op).
+        AirportDatabase.shared.loadCached()
+        Task { await AirportDatabase.shared.refresh(using: client) }
     }
 }
--- a/app/flyfun-weather/flyfun-weather/Models/API/AutorouterRoute.swift
+++ b/app/flyfun-weather/flyfun-weather/Models/API/AutorouterRoute.swift
@@ -1,0 +1,46 @@
+import Foundation
+
+/// One recent route from the user's linked Autorouter account
+/// (`GET /api/flights/autorouter-routes`). Selecting one feeds its ICAO flight
+/// plan through the same parse→fill path as "Paste Flight Plan".
+struct AutorouterRoute: Codable, Identifiable, Hashable, Sendable {
+    var id: String { routeid }
+
+    let routeid: String
+    let departure: String
+    let destination: String
+    let departureName: String?
+    let destinationName: String?
+    let departureTime: String?
+    let fplan: String
+    let routeDistanceNm: Int?
+    let aircraftDescription: String?
+    let callsign: String?
+
+    /// "EGTF → LSGS" headline for the picker row.
+    var title: String { "\(departure) \u{2192} \(destination)" }
+
+    /// Secondary line: date · distance · aircraft, omitting any missing parts.
+    var subtitle: String {
+        var parts: [String] = []
+        if let departureTime, let date = Self.isoParser.date(from: departureTime) {
+            parts.append(Self.dayFormatter.string(from: date))
+        }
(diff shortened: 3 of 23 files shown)


roznet:
Thanks — addressed all five findings in `1f5d65c2`:

**Important**
1. **`AirportDatabase` data race** — made the class `@MainActor`, so `isLoaded`/`knownAirports`/`db`/`loadTask` reads and writes are actor-isolated; the only off-main work (file IO + opening the SQLite) stays inside `Task.detached` blocks that hop back via `MainActor.run`. `cacheURL` is `nonisolated` so those detached blocks can still read it. Dropped `@unchecked Sendable`.
2. **Unauthenticated endpoint** — gated `GET /api/nav/airports-db` behind `Depends(current_user_id)` (the iOS client is always authenticated when it fetches). Test now overrides `current_user_id`.
3. **Orphaned `Task`** — removed the discarded `Task { await loadTask?.value }`.

**Minor (batched in since I was pushing anyway)**
4. **Profile picker `0`-sentinel** — switched to `Optional<Int>` tags + binding; `applyProfile` already handled `nil`.
5. **TZ picker blank row** — `setRouteTimeZones` now resets the selected zone to `preferred`/`UTC` when a route change drops it from the options. Added a unit test.

All green: iOS build, Swift tests (incl. new TZ-reset test), and the nav endpoint test pass.

roznet:
Round-2 review follow-up — merged this PR with `--rebase`, then folded the worthwhile findings into a direct commit on `main` (`1a671a1f`) to avoid another review round:

- **#1** (non-Sendable cross-actor transfer) — boxed the `FMDatabase`+`KnownAirports` hand-off in an explicit `@unchecked Sendable` struct, so only a `Sendable` value crosses into `MainActor.run`. Keeps the DB build off-main and is Swift-6 strict-concurrency ready. *(Note: not a current blocker — the target is `SWIFT_VERSION = 5.0`, so it already compiled cleanly; this is future-proofing.)*
- **#3** — `refresh()` now reads `fileExists`/`UserDefaults` off the main actor.
- **#5** — autorouter 409 catch tightened to require **both** `code == 409` **and** the `autorouter_not_linked` detail.
- **#6** — dropped `raise_server_exceptions=False` in the nav test.

Skipped **#2** (false positive — `AddFlightViewModel` is already `@MainActor`) and **#4** (`ISO8601DateFormatter` *formatting* is thread-safe and the type is main-thread-only; per-render locals would cost allocations for no real gain).

Verified on main: iOS build green, nav endpoint test passes.

PR #325, opened 2026-06-28
Expose weather alternates via MCP + ChatGPT connectors (#210/#249)

roznet:
## What

Surfaces the snapshot's **weather-based divert alternates** (`RouteAlternates`) to both agent front-doors — the Claude **MCP** server and the ChatGPT **`/agent/v1`** OpenAPI connector. Until now neither could see them: the LLM digest deliberately excludes alternates (`prompt_builder.py:269`), and the plain-text digest that *does* render them is never persisted or served.

This was driven by wanting Claude to take the divert candidates and cross-reference them with airport/AIP tools (euro_aip / airfield-directory) to review the **operational** side of an alternate before advising a pilot.

## Design — dedicated tool + a lightweight hook

Mirrors the existing `get_advisory_detail` / `cross_check_present` pattern:

- **`get_alternates`** (MCP tool) / **`getAlternates`** (`GET /agent/v1/flights/{id}/alternates`) — the full divert picture: the advisory FAA/EASA "alternate required?" trigger, the nearest airport improving each deficient axis (category/wind/crosswind), and ranked candidates with consensus weather **plus suitability** (approach type, runway, point-of-entry) for the operational cross-check.
- **`alternates` hook** in `get_briefing` / `getBriefing` — a compact signal only (required-flag + candidate count + nearest picks + `detail_tool`), so the briefing hot path stays lean while remaining discoverable.

Keeping the briefing lean + a distinct drill-in tool matched the use case (operational review is a deliberate step you'll enrich with other tools) and the server's established idiom.

### Weather-vs-operational guardrail
These are **weather-improvement** candidates, **not** operational alternates. Operational suitability (hours, customs/PPR, fuel, NOTAMs, approach currency) is **not** evaluated. That caveat ships in `ALTERNATES_NOTE` (in every payload's `note`), the tool descriptions, and the MCP server instructions, so an agent combines them with airport/AIP data before recommending a divert.

## Layout
- `connectors/views.py` — shared `summarize_alternates` + `alternates_hook` shapers (one source of truth across web/iOS/MCP/ChatGPT).
- `api/packs.py` — `GET /packs/{ts}/alternates` extracts the block from `briefing.json` (404 when not computed).
- `mcp/` — `get_alternates` tool, `get_briefing` hook, `client.get_alternates`, server-instructions note.
- `api/agent.py` — `getAlternates` route + `getBriefing` hook (description trimmed to the connector's sub-300-char convention).

## Tests (128 pass)
- Shaper units built from the **real** pydantic models (`mode="json"` dump → matches on-snapshot serialization).
- The `packs.py` endpoint **and** the ChatGPT connector exercised through the **real FastAPI app** (routing, DI, ownership gate, 404 paths).
- The MCP `get_alternates` tool path (ready / none / processing / no-pack).

## Verification note
Tested end-to-end against the real FastAPI app via TestClient; only the thin fastmcp transport layer (unchanged here) is unverified locally. Per discussion, the live MCP tool will be smoke-tested on prod after deploy — what remains after this glue is just prompt/tool-hint wording, cheap to tweak.

### DIFF
--- a/src/weatherbrief/api/agent.py
+++ b/src/weatherbrief/api/agent.py
@@ -318,6 +318,13 @@ def get_briefing(
     if alt_table:
         result["altitude_table"] = views.summarize_altitude_table(alt_table)
 
+    # Weather-based divert alternates: a compact hook only (required-flag +
+    # candidate count). Full detail via getAlternates — keep the briefing lean.
+    snapshot = _read_json(pack_dir, "briefing.json") or _read_json(pack_dir, "snapshot.json")
+    alternates = (snapshot or {}).get("alternates")
+    if alternates:
+        result["alternates"] = views.alternates_hook(alternates)
+
     return result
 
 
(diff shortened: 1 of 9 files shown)


roznet:
Merged via rebase onto current main. The review's three findings (all non-blocking — no correctness/security issues; code was confirmed correct) were applied as a direct follow-up commit on main (`eefdbbe6`):

- **Parity table** — added `getAlternates` to the endpoint↔MCP-tool table in `designs/chatgpt-connector.md`.
- **GPT instructions** — added a getAlternates note (with the weather-vs-operational caveat) to the ChatGPT OpenAPI description in `app.py`, symmetric with the MCP server instructions.
- **Test gaps** — added ownership-gate coverage on the `packs.py /alternates` route the MCP client calls (private flight of another user 404s), and the `snapshot.json` fallback path.

Issue #328, opened 2026-06-30
Generic advisory 'mitigation' concept + VFR Feasibility worked example


roznet:
## Summary

Introduce a generic **mitigation** concept for route advisories: when an advisory flags a problem, also surface *what different decision could improve the rating* (a lower clear altitude, a point along the route where a cloud deck breaks, etc.). Mitigations are **advice only** — they never change the grade. Implement the generic data-model + aggregation plumbing, and populate it as a worked example on **VFR Feasibility** with two mitigations (vertical "fly lower to clear cruise IMC" and along-route "climb after X nm / descend before Y nm" to get through a corridor deck).

This is cut 1 (backend-only). UI/digest wiring and generalizing vertical mitigation to all altitude-dependent advisories are explicit follow-ups (see *Out of scope / follow-ups*).

## Background / motivation

Today an advisory grades the pilot's chosen altitude + date and stops there. The idea: when something is flagged, search for an **alternative or mitigating decision** and report it, while leaving the grade and the go/no-go decision to the pilot. Aligns with the project principle that the briefing is an attention-director, not a verdict.

### What already exists (do not duplicate — reuse/harmonize)

Three partial pieces exist, none of them a generic per-advisory mitigation:

1. **Icing escapes** — `_descend_below_icing` / `_climb_above_icing` in `src/weatherbrief/analysis/sounding/advisories.py` already compute an escape altitude with a `feasible` flag gated on **terrain clearance** (`_TERRAIN_CLEARANCE_FT = 1000`) and the aircraft ceiling. This is effectively a vertical mitigation, but it lives on a *separate channel* (`AltitudeAdvisories`), is icing-only, and never touches the advisory result models. **Reuse its terrain-gating pattern.**
2. **Altitude table** — `compute_altitude_table` in `src/weatherbrief/analysis/advisories/altitude_table.py` sweeps all altitude-dependent advisories across altitudes (2000 ft → ceiling, 2000 ft steps), producing `best_below_cruise` / `best_above_cruise` and a `diff_altitude_rows` primitive (improved/worsened between two altitude rows). It is surfaced **only in the LLM digest** ("ALTITUDE OPTIONS: Lower/Higher option improves X, worsens Y" — see `src/weatherbrief/digest/prompt_builder.py` ~L700-752). It is route-wide and **aggregate-only** (no per-model), and **does not capture the along-route/horizontal axis** at all.
[...]

Issue #300, opened 2026-06-25
Convective advisory headline: anchor extent on MODERATE+, show peak + cross-model range (not LOW-floor union)


roznet:
## Problem

The convective advisory headline conflates three different things into one number, making it read as far more alarming/consensus than it is. Concrete case — `lfqa_lfat_egtf-2026-06-28` (Sun), ECMWF:

> **HIGH convective risk over 204nm/222nm (92%)**

Decoding what that "92%" actually is (verified against pack 278, dev-user-001):

- A point counts as "affected" when **`max(thermo, NWP) ≥ LOW`** (`min_risk=2`, LOW is enum index 2) — the DD-floor takes the max of the two tracks per point.
- Of 24 route points, **22 (91%)** clear that bar — but **16 of those are only LOW**; just **6 reach MODERATE+**.
- **thermo supplies the floor at 18 points** (broad early-mid CAPE), **NWP at only 6** (late `nwp_lcl_top` towers).
- Peak anywhere is **HIGH** (one point).

So "HIGH … over 92%" really means *"peak is HIGH at one point; 92% of the route has at least LOW risk, two-thirds of which is non-firing low-end CAPE."* It does **not** mean 92%-at-HIGH.

### Three stacked confusions

1. **Threshold conflation** — the severity word (HIGH) is the *peak*, the extent (92%) is at the *min* threshold (LOW). Different bars, presented as one phrase.
2. **Track union** — the extent unions thermo + NWP, which the cross-section draws as *separate* layers (thermo first half, NWP towers at the end), so the % matches neither visual. (This is the #283 model-native independence working as designed — but the headline doesn't reflect it.)
3. **Model spread** — 92% is ECMWF alone; GFS 25%, ICON 67%. The colour (RED) is a real 3/3 consensus, but the *number* is one arbitrary model (the representative is just the first per-model result matching the aggregate status, `models/advisories.py:from_per_model`).

## Proposed change (option "b")

Anchor the headline extent on **MODERATE+** (actual convective concern), mention the **peak**, and show the **across-model range**:

> **MODERATE+ over 25–67% across models — peak HIGH**

- Keep the LOW-end favorability available but reworded as *primed*, not *firing*, so it doesn't masquerade as active convection.
- The colour grade is unchanged — only the **display text** changes.

## Implementation sketch

**`analysis/advisories/convective.py`**
[...]

PR #329, opened 2026-06-30
feat(advisories): generic mitigation concept + VFR worked example (#328)

roznet:
Introduce a generic per-advisory "mitigation" — what different decision
could improve a flagged sub-issue (a lower altitude, a point where a deck
breaks). Mitigations are advice only: they never change the grade, same
contract as cross_check.

Data model (models/advisories.py):
- MitigationKind {ALTITUDE, ROUTE_POSITION, TIMING} + Mitigation
  (addresses tag, localized detail, axis-scoped mitigated_status,
  altitude_ft/distance_nm/reference). Exported from models/__init__.
- ModelAdvisoryResult.mitigations (+ build() param) and
  RouteAdvisoryResult.aggregate_mitigations, both default-empty so old
  packs deserialize cleanly.
- _aggregate_mitigations(per_model, agg_status): representative-model
  policy (mitigations of the first per-model result matching the aggregate
  status), isolated in one swappable helper. Wired into from_per_model.

VFR Feasibility worked example (vfr_feasibility.py):
- Vertical mitigation (addresses cruise_imc): terrain-gated downward scan
  for the highest altitude that improves the en-route cloud axis, preferring
  a fully clear band. Refactored _check_enroute_vfr to take an explicit
  altitude_ft and extracted _enroute_vfr_status so cruise and candidate
  altitudes grade identically.
- Along-route mitigation (addresses climb_deck / descent_deck): locate where
  a corridor deck breaks and offer "climb after ~X nm" / "descend before
  ~Y nm", only when there is a genuine clear/blocked split.
- mitigated_status is axis-scoped, never the overall grade.

Localization: vfr.mitigation.{altitude,climb_after,descend_before} in all
four locales (en/fr/de/es).

Tests: tests/test_vfr_mitigation.py covers vertical happy/terrain-blocked,
along-route climb/descent/uniformly-blocked, co-occurrence, representative-
model aggregation, and backward-compat deserialization.

Claude-Session: https://claude.ai/code/session_01CUkyaaJirjKDV25R8EG3GL

### DIFF
--- a/src/weatherbrief/analysis/advisories/strings.py
+++ b/src/weatherbrief/analysis/advisories/strings.py
@@ -414,6 +414,12 @@ _STRINGS: dict[str, dict[str, str]] = {
         "de": "VMC verfügbar auf {alt:,} ft",
         "es": "VMC disponible a {alt:,} ft",
     },
+    "vfr.mitigation.altitude_marginal": {
+        "en": "Marginal VMC at {alt:,} ft",
+        "fr": "VMC marginale à {alt:,} ft",
+        "de": "Marginale VMC auf {alt:,} ft",
+        "es": "VMC marginal a {alt:,} ft",
+    },
     "vfr.mitigation.climb_after": {
         "en": "VMC climb to cruise possible after ~{dist} nm from departure",
         "fr": "Montée VMC vers la croisière possible après ~{dist} nm du départ",
--- a/src/weatherbrief/analysis/advisories/strings.py
+++ b/src/weatherbrief/analysis/advisories/strings.py
@@ -408,6 +408,24 @@ _STRINGS: dict[str, dict[str, str]] = {
         "de": "{cov}-Schicht im Sinkflug bei {icao}",
         "es": "Capa {cov} en descenso en {icao}",
     },
+    "vfr.mitigation.altitude": {
+        "en": "VMC available at {alt:,} ft",
+        "fr": "VMC possible à {alt:,} ft",
+        "de": "VMC verfügbar auf {alt:,} ft",
+        "es": "VMC disponible a {alt:,} ft",
+    },
+    "vfr.mitigation.climb_after": {
+        "en": "VMC climb to cruise possible after ~{dist} nm from departure",
+        "fr": "Montée VMC vers la croisière possible après ~{dist} nm du départ",
+        "de": "VMC-Steigflug auf Reiseflughöhe ab ~{dist} nm nach dem Start möglich",
+        "es": "Ascenso VMC a crucero posible tras ~{dist} nm desde la salida",
+    },
+    "vfr.mitigation.descend_before": {
+        "en": "VMC descent possible until ~{dist} nm before arrival",
+        "fr": "Descente VMC possible jusqu'à ~{dist} nm avant l'arrivée",
+        "de": "VMC-Sinkflug bis ~{dist} nm vor der Ankunft möglich",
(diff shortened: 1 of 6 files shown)


roznet:
Thanks for the review. Triage + actions (pushed in 8051beb):

**Fixed**
- **#2 Corridor logic duplication** — extracted `_corridor_points` as the single source of truth for corridor membership (nearer-half tiebreak) and the transitable-deck condition (`floor < top_ft < cruise`). Both `_check_corridor_vfr` (grade) and `_corridor_blocked_profile` (mitigation) now consume it. Behaviour-preserving — re-ran the existing corridor scenarios (BKN→AMBER, OVC→RED, short-route overlap, midroute-stays-GREEN) plus the new mitigation tests; grades/details unchanged.
- **#3 Inconsistent rounding** — `_corridor_mitigation` now rounds the distance once and uses the same integer for both the localized `detail` string and the structured `distance_nm` (phrasing is qualitative "~X nm"), so they can't diverge at `.5` boundaries.

**Deferred (intentionally, not done here)**
- **#1 TS types (`web/ts/types/advisories.ts`)** — leaving this for the UI fast-follow. Issue #328 explicitly scopes client/UI wiring out of cut 1 and notes the data shape is *"to be reviewed"* before surfacing in clients, so I'd rather not pre-commit the TS shape now. This is also consistent with current practice: the TS `ModelAdvisoryResult` mirror is already intentionally partial — the Python `affected_mod_points/pct/nm` fields (from the convective #300/#302 work) were never mirrored to TS either. The mitigation fields ride into the pack JSON via pydantic regardless. Will add `Mitigation`/`MitigationKind` + the two list fields (and run `/sync-ios-web`) when the popup/digest consumer lands.

---

roznet:
Triage of the round-2 review (pushed in ddc7ee3):

**Fixed**
- **#1 Misleading altitude detail (Important)** — `_vertical_mitigation` can return an AMBER band (cruise RED, no clear band above the terrain floor); the detail used to read "VMC available at X ft", contradicting the AMBER badge. Now branches on `best_status`: GREEN keeps "VMC available", AMBER uses a new `vfr.mitigation.altitude_marginal` ("Marginal VMC at X ft") in all four locales. Added `test_vertical_amber_band_uses_marginal_phrasing`.
- **#3 `_SEVERITY` KeyError footgun (Minor)** — hardened the trigger guard to `cruise_status not in (AMBER, RED)` (was `== GREEN`). This restates the "only flag AMBER/RED axes" contract (#328 decision 6) *and* makes every `_SEVERITY[...]` lookup provably safe — UNAVAILABLE / any future status can no longer reach the dict. (Preferred this over mapping `UNAVAILABLE: -1`, which is ambiguous: -1 would be safe as a cruise status but would wrongly count as an "improvement" if it ever appeared as a candidate.)

**Deferred**
- **#2 iOS Swift Codable DTO** — same call as the TS-types deferral: client wiring is out of cut-1 scope (issue #328), `Codable` silently ignores the unknown keys so nothing breaks on-device, and the fields ride in the JSON regardless. Will add the Swift `AdvisoryMitigation` struct + both list fields (and run `/sync-ios-web`) with the UI fast-follow when the feature is surfaced.

**Declined**
- **#4 `{alt:,}` English thousands separator** — accepting the English convention as intentional (as the review offers). Aviation altitudes are read with comma separators internationally, and there's no locale-aware number formatter elsewhere in the advisory strings to reuse.

---

Issue #330, opened 2026-06-30
Phase 2: surface advisory mitigations across web / iOS / digest / MCP (#328 follow-up)


roznet:
## Summary

Phase 2 of the advisory **mitigation** concept: surface mitigations to users across **web**, **iOS**, the **AI digest**, and **MCP** — as advice that *never* changes the grade. Phase 1 (the generic `Mitigation` model + the VFR Feasibility worked example) shipped in #328 / PR #329. The data already exists end-to-end; this issue is about *displaying* it well.

Implement as **four independent PR slices** in order: **Web → iOS → Digest → MCP**. Each slice below is self-contained with its own acceptance criteria. A Phase-3 stretch (interactive "what-if" chips) is explicitly out of scope.

## Background

A **mitigation** is "a decision that could improve a flagged sub-issue" (fly a lower altitude to clear cruise IMC; climb after N nm to thread a departure cloud deck). The model (`src/weatherbrief/models/advisories.py`):

```python
class MitigationKind(str, Enum):
    ALTITUDE = "altitude"; ROUTE_POSITION = "route_position"; TIMING = "timing"

class Mitigation(BaseModel):
    kind: MitigationKind
    addresses: str                    # stable tag: "cruise_imc", "climb_deck", "descent_deck" (NOT localized)
    detail: str                       # localized human phrasing (via adv_t)
    mitigated_status: AdvisoryStatus  # status of the ADDRESSED sub-issue if applied — NOT the advisory overall
    altitude_ft: int | None = None
    distance_nm: float | None = None
    reference: str | None = None      # "departure" / "arrival"
```

- `ModelAdvisoryResult.mitigations: list[Mitigation]` (per-model) and `RouteAdvisoryResult.aggregate_mitigations: list[Mitigation]` (representative-model policy). Both default empty.
- **Hard contract (preserve on every surface):** a mitigation NEVER alters `status` / `aggregate_status`. Same "informs but doesn't grade" contract as `cross_check` — which is the **rendering precedent to mirror** on all four surfaces.
- Today only VFR Feasibility produces mitigations, but all code here must be **generic** (any advisory may populate the lists).

## Governing principles (apply to all slices)

1. **Advice only — never regrade.** The badge/status stays first and unchanged; the mitigation is always visually/semantically subordinate.
[...]

PR #336, opened 2026-07-01
fix(refresh): keep just-departed flights on the live briefing path

roznet:
## Problem

Refreshing a flight even **1 minute after departure** flagged it "historical" (`departure_time < now`, zero grace) at three sites in `api/packs.py`. That single flag:

1. **Skipped the LLM digest** (`do_llm_digest = False`) → `has_digest:false`, bare `convective=RED,…` fallback assessment instead of an AI narrative.
2. **Dropped from the fast paid Open-Meteo API to the free-tier `historical-forecast-api.open-meteo.com`**, which hit repeated 60s read timeouts and stalled the SSE refresh stream → a "network connection lost" / half-failed refresh.

Reproduced from prod logs: a 5-leg flight (…→LSGS/Sion) departed 07:00 UTC, refreshed at 07:57 while still en route → digest skipped + historical-host timeouts.

## Fix

New `_classify_refresh_time(flight)` helper, used at all three refresh sites:

| State | Behaviour |
|---|---|
| Not yet departed | Normal live refresh |
| **In progress / within 3h of estimated arrival** | **Still live** (paid API + digest + GRAMET, no admin gate), but GRIB run-selection **pinned to departure** via `as_of_time` |
| Well past (beyond arrival + grace) | Historical archive path, as before |

### Why the GRIB pin matters
For an in-progress flight the *freshest* GRIB run's init can fall **after** the flight window, and `bracket_forecast_hours` silently clamps pre-init hours to `f000` (the analysis snapshot) — mis-enriching the flight. Pinning run-selection to `as_of_time = departure` (a mechanism GFS/ICON/ECMWF already honor) makes the chosen run actually forecast the window. This required unblocking `tasks/fetch.py`, which previously dropped `as_of_time` unless `historical_mode`.

Estimated arrival uses `flight.flight_duration_hours` (already on the flight — no extra route build). A 5h flight stays fully live for ~8h from departure.

## Tests
- 7 new `_classify_refresh_time` cases (future / just-departed / mid-flight / within-grace / beyond-grace / zero-duration) in `tests/test_historical.py`.
- Full suite green: `247 passed, 2 skipped` across refresh/pack/fetch/historical.

## Known cosmetic edge
A flight that departed *yesterday* but is refreshed within grace runs live correctly, though its pack still labels `is_historical` (that flag is `days_out < 0`) — labeling only, no behavioral impact.

### DIFF
--- a/src/weatherbrief/api/packs.py
+++ b/src/weatherbrief/api/packs.py
@@ -880,6 +880,38 @@ _STAGE_PROGRESS: dict[str, float] = {
 }
 
 
+# Grace period after a flight's estimated arrival during which it still counts
+# as "in progress" for refresh purposes. A pilot who refreshes shortly after
+# (or during) the flight legitimately wants a fresh, full briefing — paid API,
+# LLM digest, GRAMET, no admin gate — for the remaining route, not the degraded
+# historical path. Only once the flight is well past does it become historical.
+_INFLIGHT_GRACE = timedelta(hours=3)
+
+
+def _classify_refresh_time(flight, now: datetime | None = None) -> tuple[bool, datetime | None]:
+    """Classify a flight for refresh, returning ``(is_historical, grib_as_of)``.
+
+    - Not yet departed → ``(False, None)``: normal live refresh.
+    - In progress or within the post-arrival grace window →
+      ``(False, departure_time)``: still a live refresh (paid API + digest),
+      but GRIB run-selection is pinned to the departure time so the chosen run
+      actually forecasts the flight window. Without this, run-selection would
+      pick the freshest run — whose init can fall *after* the window — and
+      ``bracket_forecast_hours`` would clamp every pre-init hour to f000 (the
+      analysis snapshot), silently mis-enriching the flight.
+    - Well past (beyond estimated arrival + grace) → ``(True, None)``:
+      historical archive path (free API, no digest, admin-only).
+    """
+    now = now or datetime.now(timezone.utc)
+    if flight.departure_time >= now:
+        return False, None
+    duration_h = flight.flight_duration_hours or 0.0
+    flight_end = flight.departure_time + timedelta(hours=duration_h)
+    if now <= flight_end + _INFLIGHT_GRACE:
+        return False, flight.departure_time
+    return True, None
+
+
 def _build_route_config(flight, db_path):
     """Build a RouteConfig from a flight, resolving waypoints."""
     from weatherbrief.airports import resolve_waypoints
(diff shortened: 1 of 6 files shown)


roznet:
Thanks for the review. Actioned:

**Important — `decide_refresh` `days_out == 0` vs `<= 0` → FIXED.** Confirmed the gap: `_classify_refresh_time` marks a flight that departed yesterday (UTC) but is still mid-flight/within grace as live, so `_days_out_now` returns `-1` and both "always at least realtime" fallbacks fell through to `mode="none"`. Changed both branches (the `n_eligible == 0` case and the partial-updated case) to `days_out <= 0`, matching `_refresh_threshold`. Added two `decide_refresh` regression tests for `days_out == -1`. `test_packs.py` + `test_scheduler.py` + `test_historical.py` green.

**Minor — polluted diff (corridor mitigation, popups, `eval_workbench/ingest.py`) → RESOLVED by rebase.** Those weren't part of this change — the branch was cut from a local `main` that was 4 commits ahead of `origin/main`, so the PR dragged in unrelated (unpushed) work. Rebased onto `origin/main` (`rebase --onto`), so the diff is now exactly this change: `api/packs.py`, `tasks/fetch.py`, `tests/test_historical.py`, `tests/test_packs.py`.

**Minor — iOS ↔ web mitigation parity → deferred (not this PR).** Belongs to the #330 mitigation work, not the refresh fix; will scope via `/sync-ios-web` separately.

roznet:
Round-2 findings actioned (commit e3e22d16):

**Imp 1 — unbounded `flight_duration_hours` → FIXED.** Bounded the live window to `min(duration + 3h grace, 12h)` from departure via a new `_MAX_INFLIGHT_WINDOW = 12h` (matching the web UI's 12h duration cap). An oversized/unvalidated duration can no longer keep a flight "in progress" indefinitely — it drops to the admin-gated historical/archive path at most 12h after departure. Test added.

**Imp 2 — resolved `as_of` not threaded to finalize → FIXED.** `_prepare_refresh` now returns the resolved `as_of_time` (including the in-progress departure pin); threaded through all three finalize sites — sync (`_finalize_refresh`), stream (`_persist_pack_finalize`), and scheduler. So the persisted `BriefingPackMeta.days_out` / `is_long_range` now match what the pipeline actually used, instead of being recomputed from "today" (which went negative for a cross-midnight in-progress pack). Same-day case unchanged.

**Min (docstrings) — FIXED.** `refresh_briefing` / `refresh_briefing_stream` no longer imply `as_of_date` is historical-only.

**Min (multiple `_classify_refresh_time` calls) — not changed.** The endpoint 403-gate and `_prepare_refresh` still classify independently; the only divergence risk is a sub-second wall-clock crossing of the grace boundary between the two calls, which at worst flips one refresh between live/historical with no correctness impact. Threading a single classification through would need an endpoint refactor I judged not worth the churn — happy to do it if you'd prefer.

Tests green: `test_historical` + `test_packs` + `test_scheduler` (109) and `test_api` (88).

PR #338, opened 2026-07-01
feat(advisories): shared vertical-profile path-finder for VFR + icing mitigations (#335)

### LINKED ISSUE #335: Shared vertical-profile path-finder for altitude-dependent advisories (VFR hard-wall + icing-escape soft-wall)
## Summary

Replace the current per-axis VFR mitigation logic (climb_deck / descent_deck / cruise_imc, each with its own gates) with a **shared vertical-profile path-finder** over a `(route-distance × altitude)` grid, and make it reusable across altitude-dependent hazard advisories. First two consumers, deliberately one *hard-wall* and one *soft-wall* case:

- **VFR feasibility (hard wall):** you cannot legally enter a BKN/OVC deck.
- **Icing escape (soft wall):** you *can* transit an icing layer at a cost (thickness/severity) to reach clear air above or warm air below.

Building it cost-based from the start is what lets both reuse the same solver. This supersedes the ad-hoc corridor logic added in #328/#330 (terminal corridor, `under_deck_flyable`, midpoint scan, 25nm reposition cap, cruise-green gate) — those all become emergent properties of the path model.

Background: this came out of the #330 mitigation work. The corridor mitigation kept needing new gates to avoid nonsense ("climb to cruise after 10nm" when cruise is IMC later). Those are symptoms of decomposing one question ("is there a continuous VMC path, and what is it?") into three proxy axes.

## Core idea

Model each hazard as a **cost field**: for each route point and altitude bin, a cost.
- `∞` = hard wall (cannot occupy / cannot cross) — e.g. inside a BKN/OVC deck for VFR.
- finite = soft wall (occupy/cross at a penalty) — e.g. icing severity × thickness.
- `0` = feasible.

Bounded below by **terrain + margin** (per point) and above by the **flight ceiling** (per point). Then a small DP / shortest-path over the grid returns:
- the **min-cost continuous vertical profile** (climb/cruise/descent transitions), or
- the **blocking segment** when no feasible path exists ("no VMC between 180–220 nm — deck surface→FL80").

Grid is tiny (~60 points × ~40 altitude bins), so the general DP is effectively free vs. a bespoke connectivity solver.

### Design decisions (agreed)

1. **Cost-based, not connectivity-based.** Hard walls are just `∞`. This is the decision that makes cloud (hard) and icing/turbulence (soft) share one algorithm. A pure interval-connectivity solver would fit VFR but force a rewrite for icing.
2. **Above cruise is a valid solution** as long as it's below the ceiling. The planned cruise is a *preference*, not the upper bound; the ceiling is the bound. If the only feasible band is above cruise, report it ("VFR only possible above your planned cruise, at N ft, below ceiling").
3. **Pluggable per-advisory part = the hazard → cost-field mapping only.** The solver, terrain floor, ceiling cap, transition detection, and profile/blockage output are shared. Grading (R/A/G) and phrasing stay per-advisory.
4. **v1 ignores climb/descent rate.** Leave a hook for a rate/gradient constraint (a band that requires an impossible climb gradient) but don't implement it first.
5. **Vertical transitions accumulate crossing cost.** A climb from a→b at a point sums the cost of cells traversed; `∞` if any hard-wall cell is crossed (so VFR can't climb through a deck, but icing can climb through a thin layer at a penalty).

## Interface (sketch)

```
solve(cost_field[point][alt_bin], terrain_floor[point], ceiling[point],
      preferred_alt=cruise, rate_limit=None)
  -> Profile{ segments: [(dist_from, dist_to, alt_band)], transitions, total_cost }
   | Blockage{ from_nm, to_nm, reason }
```

Each advisory implements only `build_cost_field(ctx, model) -> cost_field, terrain_floor`.

## Worked example A — VFR (hard wall)

Cloud cost field: `∞` inside BKN/OVC (+ clearance margin), else `0`.

- **EDDN→EGSG** (real eval pack, cruise 10k, en-route IMC): the connected VMC band never reaches cruise → profile is "fly ~7,000 the whole way," and there is *no* "climb to cruise" claim. The cruise-green gate we added in #330 becomes unnecessary — it falls out.
- **lfmd→EGTF** (cruise clear, OVC deck at arrival): forward holds cruise, backward from arrival is forced low under the deck; they meet at the descent point → "descend before ~X nm" is *derived*, not special-cased.

Expected: reproduces today's good mitigations, drops the bad ones, and retires the corridor/`under_deck_flyable`/cap/gate special-cases.

## Worked example B — Icing escape (soft wall)

Icing cost field from temp + moisture + severity: high inside the icing band, **0 below the freezing level** (warm air is feasible even in cloud), finite (crossable) for thin/light layers.

- Today `icing_escape` only models a single transition ("descend to warm air"). The path model expresses the richer maneuver the current code can't: **climb above the icing to on-top, then descend below it (clear of terrain) before destination** — a two-transition profile — or report that no ice-free/warm continuous band exists.
- Porting `icing_escape` onto the shared solver is the validation that the interface is right (consumer #2), and it upgrades the feature.

## Scope

**In:**
- Shared cost-field vertical-profile solver + interface + tests (synthetic grids: single deck, multi-deck with gap, deck-to-terrain blockage, above-cruise-only band, no-path).
- Port `vfr_feasibility` mitigation onto it (behavior-compare against the current corridor logic over the eval corpus — the 30+ packs with corridor mitigations are the before/after set).
- Port `icing_escape` onto it (gains climb-over-then-descend).

**Out / future:**
- `turbulence`, `fiki_icing`, `cloud_top`, `vmc_cruise` as further consumers (validate interface with 2 first; don't over-generalize on guessed requirements).
- **Combined multi-hazard profile** (sum cost fields → one "clear of cloud + ice + turbulence" recommended altitude). This is the strategic prize the shared solver unlocks, but not v1.
- `headwind` as a pure-cost consumer (min-headwind profile) — fits the interface, future.
- Lateral (re-route) escapes — the model is 2-D (route vertical slice) only. Same limitation as today.
- Do **not** force non-path advisories in (`freezing_precip`, `llws`, `density_altitude`, `mountain_wind`).

## Non-goals / risks

- Grading semantics: keep each advisory's R/A/G as-is initially; the solver replaces the **mitigation** layer only, to bound regression risk. (Letting the path drive the grade is a separate, later discussion.)
- Greedy "single max altitude" is *not* sufficient — must be interval/cost reachability, or multi-deck-with-gaps routes get misjudged.

Relates to #328, #330.

roznet:
## Summary

Replaces the per-axis VFR mitigation logic (`cruise_imc` / `climb_deck` / `descent_deck`, each with its own compensating gates) with a **shared cost-field shortest-path solver** over a `(route-distance × altitude)` grid, and reuses it for `icing_escape`. Implements #335.

Hard walls are `∞`, soft walls are finite cost, so one solver serves both a **hard-wall** consumer (VFR: BKN/OVC in cloud = `∞`) and a **soft-wall** consumer (icing: crossable rime at a penalty, warm air below the freezing level free, SLD = `∞`). Grades are untouched — this replaces the **mitigation** layer only (advice-only).

Design doc: `designs/future/vertical-profile-solver.md`.

## What changed

- **New `vertical_profile.py`** — `CostModel` / `Profile` / `Blockage` + a lexicographic DP `solve()` and floor-anchoring helpers. Objective: *no hard-wall → lowest finite hazard → closest to preferred cruise → fewest transitions* (deviation must precede transitions, else a profile forced low by a departure deck never climbs back to cruise).
- **VFR port** — one solved profile yields the right mitigation in every case; the cruise-green mutual-exclusivity, midpoint scan, and `under_deck_flyable` gates become **emergent and are deleted**. The vertical floor is unified to `terrain + mitigation_min_base_agl_ft` (3,000 ft) for both "fly lower" and "fly under a deck"; the `mitigation_max_reposition_nm` cap is kept as a presentation filter.
- **Icing port** — additive (grade unchanged); gains the **climb-on-top** and **climb-over-then-descend** maneuvers the old single-transition code couldn't express.
- **Model** — `Mitigation` gains an optional structured `MitigationProfile` (segments + transitions) in cross-section coordinates for the future overlay; flat `altitude_ft` / `distance_nm` fields kept, fully backward-compatible.

## One deliberate behavior change

Per the design discussion, the VFR floor is the single conservative scud-running margin (3,000 ft). Consequence, by design: a tight terrain-to-deck gap leaving under 3,000 ft no longer produces a marginal "fly lower" tip (it did under the old 1,000 ft floor). Covered by `test_vertical_tight_terrain_gap_suppresses_mitigation`.

## Validation

- **Full suite green:** 3109 passed, 10 skipped.
- New tests: `test_vertical_profile.py` (solver, synthetic grids), `test_vfr_mitigation.py` (VFR port reproduces the design doc's worked examples), `test_icing_escape_mitigation.py` (icing port).
- **Eval-corpus behavior compare** (old-code vs new-code over 204 staging packs via `eval_workbench.rerun`): **equal-or-better, no regressions**. VFR `cruise_imc` removals = **40 blockage** (new solver finds a full-column cloud wall → no continuous VMC path; the old %-scan's lower tip was unflyable) + **14 reaches-cruise** (continuity-aware profile reaches cruise; old was over-eager) + **1 below-new-floor** (expected) + **1 changed** (highest continuously-reachable band) + 1 kept. Corridor tips: a refinement wash. Icing: ~40 new escape tips (ceiling-bounded).

## Follow-ups (not in this PR)

- Cross-section overlay that renders `MitigationProfile` (separate issue; this PR ships the data shape).
- Whether a realistic-but-high icing "climb on-top" should carry an O₂/altitude note (advice-only and ceiling-bounded today).

Closes #335


roznet:
Thanks — addressed the two blockers plus the two minors in `d2df69ad`:

- **Finding 1 (Critical, icing hard walls)** — fixed. `_icing_cell_cost` now treats **only thin/light RIME** as finite-crossable; MODERATE/SEVERE at any type, non-rime (clear/mixed) light ice, and SLD are all `∞`. This also closes the secondary `icing_type` gap you noted. Locked by a new unit test `test_icing_cell_cost_only_light_rime_is_soft`.
- **Finding 2 (Important, interior deck)** — fixed. Added `_has_interior_below_cruise`; `climb_deck`/`descent_deck` are now suppressed when the profile dips below cruise *in the interior* (not just at the terminals), so the misleading "climb to cruise after X nm" can't fire when the flight must descend again further out. New regression test `test_interior_deck_suppresses_corridor_mitigation` with a mid-route deck.
- **Finding 3 (Minor, dead var)** — removed `dist = model.distances_nm` in `solve()`.
- **Finding 4 (Minor, icing status binary)** — with finding 1 fixed, any residual escape cost can only be light rime, so GREEN/AMBER is now correct; added a clarifying comment.
- **Finding 5 (Swift DTO parity)** — deferred as you suggested; not a bug today (Codable ignores unknown keys) and the cross-section overlay is a separate follow-up. Noted for a `/sync-ios-web` pass when that lands.

Full suite green (3111 passed). Design doc updated to record both tightened invariants.

PR #339, opened 2026-07-02
feat(auth): reviewer carve-out for hardened deep-link login (H8)

roznet:
## Summary

Companion to **roznet/flyfun-common#7** (native auth-code deep-link flow). Together they close **SECURITY_AUDIT.md H8**.

The normal Google/Apple sign-in is captured inside `ASWebAuthenticationSession` and never reaches `onOpenURL`, so the only legitimate inbound deep link is the **App Store reviewer** token. This PR hardens `AppState.handleAuthCallback` to accept a bare-token deep link **only** when it carries a `scope:"review"` claim:

- The claim is read (unverified) **for routing only** — the server verifies the JWT signature on every API call, so a forged `review` token is inert (401s on first use, never reaches a real account).
- Any other bare-token deep link (an attacker's own token) is ignored → the login-CSRF / session-fixation vector is closed.
- The one-tap reviewer link keeps working.

`scripts/mint_reviewer_token.py` now stamps `scope:"review"` so the minted token passes the carve-out.

## Release sequencing

The code-exchange login activates only after flyfun-common#7 is published (PyPI + SPM tag), the server is deployed, and this app bumps its flyfun-common SPM pin. This app-side carve-out is independently safe to ship (it doesn't depend on the new flyfun-common API). Beta testers re-authenticate on the new build.

## Test plan
- App builds clean.
- `mint_reviewer_token.py` runs and the minted token decodes with `scope:"review"`.

### DIFF
--- a/app/flyfun-weather/flyfun-weather/App/AppState.swift
+++ b/app/flyfun-weather/flyfun-weather/App/AppState.swift
@@ -140,14 +140,41 @@ final class AppState {
         setupClient()
     }
 
-    /// Handle a deep-link auth callback (`flyfunweather://auth?token=…` or
-    /// `https://weather.flyfun.aero/auth/callback?token=…`).
+    /// Handle a deep-link auth callback.
+    ///
+    /// The normal Google/Apple sign-in is captured *inside*
+    /// `ASWebAuthenticationSession` and never reaches `onOpenURL`, so the only
+    /// legitimate inbound deep link here is the **App Store reviewer** token
+    /// (`flyfunweather://auth?token=<jwt>`), which carries a `scope:"review"`
+    /// claim. Any other bare-token deep link is a login-CSRF attempt and is
+    /// ignored. The claim is read only to *route* — the server still verifies
+    /// the signature on every API call, so a forged review token is inert
+    /// (401s on first use) and never reaches a real account.
     func handleAuthCallback(url: URL) {
         guard let token = callbackParser.token(from: url) else {
             Self.logger.warning("Invalid auth callback URL: \(url)")
             return
         }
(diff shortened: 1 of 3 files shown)


roznet:
Thanks — addressed the review. Summary of this round:

**Important #1 (residual forced-logout / session disruption) — fixed.** `handleAuthCallback` now bails when the app is **already authenticated** (`guard jwt == nil`), so a spoofed `scope:"review"` token can no longer be persisted over a logged-in user's real session token (the forced-logout-on-next-401 vector). The reviewer path only ever runs on a fresh, signed-out device, so this doesn't regress it. Account-takeover / session-fixation was already closed server-side (signature verified on every call); I rewrote the doc-comment to state exactly that guarantee rather than over-claiming.

**Important #2 (test coverage) — fixed.** Added `ReviewTokenScopeTests.swift` covering `unverifiedScope`: `scope:"review"`, missing scope, non-`review` scope, wrong-type scope, non-JWT-shaped string, non-JSON payload, and the base64-padding path. Made `unverifiedScope` `nonisolated` so it's testable off the MainActor. All 7 pass.

**Minor (broken doc path) — fixed.** The mint-script comment now points at `flyfun-common/designs/oauth-deeplink-hardening.md`.

**Important #3 (cross-repo duplication) — confirmed no duplication, no change.** flyfun-common#7 ships **no** scope-routing check; per `oauth-deeplink-hardening.md` the carve-out is intentionally app-side (the app decides whether to trust an inbound `onOpenURL` bare token). flyfun-common#7 only retains `token(from:)` for this path. If a second app later needs the same routing, we can lift `unverifiedScope` into flyfun-common then — flagged so it's not lost.

Issue #236, opened 2026-06-12
Investigate standalone_forecast memory peak (anon ~4.7 GB at peak — hits 6 GiB container cap, host swap pressure)


roznet:
## Summary

Production box (`weather.flyfun.aero`, 7.8 GiB host RAM) feels sluggish during/after the morning `standalone_forecast` cycle. Investigation shows the cycle's anonymous working set approaches the container's 6 GiB cap, forcing the kernel into heavy swap activity that affects every other service on the host.

**This is not a leak** — per-briefing `request_growth` numbers are healthy (returns to baseline). It's a single periodic background task whose peak memory is too large for the host envelope.

## Evidence — timeline (UTC, container restarted 21:10 Jun 11)

| Time | hwm (MB) | swap (MB) | Event |
|------|----------|-----------|-------|
| 21:48 | 1114 | 0 | post-restart |
| 04:00–06:00 | 1649→1978 | 43→147 | normal briefings |
| **06:09** | **3394** | 372 | **+1416 MB hwm jump** (first standalone cycle starts) |
| 06:15 | 3394 | 350 | `standalone_verification` light finishes (peak 2952 MB) |
| 06:20 | 3394 | 654 | `hewson.precompute` finishes — swap +302 MB |
| 07:15–**08:27** | 3394 | 754 | **`standalone_forecast` runs 72 min** |
| 08:27 | 3394 | 754 | Cycle ends. **peak_cgroup = 6122 / 6144 MiB** (22 MB from limit). Anomaly fires. |
| **08:29** | 3394 | **1311** | swap +557 MB in 2 min — kernel reclaiming after the cycle |

App's own anomaly detector caught it:
```
WARNING:weatherbrief.tasks.standalone_verification:Standalone cycle memory anomaly
  (source=standalone_forecast): peak_rss=3132MB peak_cgroup=6122
  baseline=3309 cgroup_limit=6144 (relative_threshold=4632, absolute_threshold=4915)
INFO:weatherbrief.scheduler:Standalone forecast cycle: 3 models, 46425 snapshots,
  0 observations, 0 scores (4344506ms)
```

72 minutes, 619 airports × 3 models (GFS, ICON-EU, ECMWF — 226 GRIB files), 46,425 snapshots accumulated.

## Why anon, not file cache

Initially I thought the cgroup peak was mostly mmap'd GRIB file cache that'd just get reclaimed under a tighter cap. The cgroup breakdown says otherwise:

```
[...]

PR #340, opened 2026-07-02
feat(timing): timing-scenario scan — background ECMWF departure-window discovery

roznet:
Implements the timing-scenario scan (`designs/plans/timing-scenario-plan.md`) — the reserved `MitigationKind.TIMING` axis. When a briefing flags a GRIB-dependent hazard that genuinely varies through the day, it surfaces **"a better departure time may exist"** — an attention-director, never a verdict.

## How it works

- **The v1 primitive** `extend_ecmwf_enrichment_daylight` re-decodes the daylight ECMWF fhours (decode-only, local ECPDS disk, BACKGROUND priority) into a separate `cross_section_ext.json`, reusing the production `_enrich_ecmwf` + `propagate_all`. **Honesty invariant**: a candidate hour is graded only if its whole ETA span falls inside the *decoded* coverage — never a silent `at_time()` clamp presenting OM-clamped values as ECMWF.
- **`run_time_scan`**: Relevance gate (scan-class advisory RED/AMBER) → daylight window (euro_aip solar) → ECMWF-horizon clip → extend enrichment → grade native-cadence candidates ECMWF-only via `run_alt_from_pack` → full-picture improve/worsen diff with trigger-weighted margin → `time_options.json`.
- **Detached background job**: runs *after* the briefing completes, so the refresh finishes as usual and the timing panel appears when the briefing is next viewed (+ a gated in-session client poll so it auto-appears). The expensive multi-model **confirm** is gated on a user tap (deferred ICON/GFS decode); the confirm-**downgrade** is a first-class designed outcome.

## Layers

- `timing_class` on `AdvisoryCatalogEntry` + `get_scan_class_ids()`; all 21 evaluators annotated per the plan (9 `scan` / 5 `cheap` / 7 `none`).
- `TimeWindowScan` / `TimeCandidate` / `TimeConfirmation` models + `time_options.json` / `cross_section_ext.json` artifacts.
- Endpoints: `GET .../time-options`, `POST .../time-options/confirm`.
- `alt_departure_time` reframed as the pinned **preferred departure time** (scan anchor + ranking tiebreak; **no DB migration**).
- Digest synchronous timing pointer (no numbers) + MCP/ChatGPT `timing_referral`.
- Web UI: neutral soft 🕐 hook → candidate list, provisional→confirmed ladder, downgrade shown plainly, capped ~3, **never auto-switches** the plan.

## Validation

- Full end-to-end on a live ECMWF run (`20260702_00z`): a RED midday flight (`vfr_feasibility`) surfaced 3 calmer AMBER windows (08/09/17Z) with the correct full-picture diff (early windows improve VFR but *worsen* airport_wind). Verified the scan runs detached — `execute_briefing` returns before `time_options.json` exists, which lands ~2 min later.
- New `tests/test_time_scan.py`; touched Python suites green; frontend `tsc --noEmit` + esbuild clean.

## Known v1 follow-ups (plan open questions)

- Detached execution uses a single background thread; subprocess isolation (#236) is a future option (Open q6).
- Confirm endpoint is synchronous, not SSE-streamed (Open q4).
- Variance pre-filter dropped per the plan's default (always-scan-when-flagged).
- `cross_section_ext.json` is ~7–13 MB per scanned pack (could gzip / drop after confirm).

### DIFF
--- a/src/weatherbrief/analysis/advisories/__init__.py
+++ b/src/weatherbrief/analysis/advisories/__init__.py
@@ -72,4 +72,4 @@ class AdvisoryEvaluator(Protocol):
     def evaluate(ctx: RouteContext, params: dict[str, float]) -> RouteAdvisoryResult: ...
 
 
-from weatherbrief.analysis.advisories.registry import evaluate_all, get_altitude_dependent_ids, get_catalog, resolve_enabled_ids  # noqa: E402, F401
+from weatherbrief.analysis.advisories.registry import evaluate_all, get_altitude_dependent_ids, get_catalog, get_scan_class_ids, get_timing_class_ids, resolve_enabled_ids  # noqa: E402, F401
--- a/src/weatherbrief/analysis/advisories/airport_wind.py
+++ b/src/weatherbrief/analysis/advisories/airport_wind.py
@@ -73,6 +73,7 @@ class AirportWindEvaluator:
     def catalog_entry() -> AdvisoryCatalogEntry:
         return AdvisoryCatalogEntry(
             id="airport_wind",
+            timing_class="cheap",
             name="Airport Wind",
             short_description="Crosswind and gusts at airports",
             description=(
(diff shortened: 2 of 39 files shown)


roznet:
won't merge, used another version

PR #346, opened 2026-07-03
fix(consensus): unify model-consensus across arrival card, alternates, map & advisory

roznet:
## Why

On the LFAT briefing (`egkb_lfat-2026-07-05`) the same airport was rated **VFR** on the airport arrival card but **MVFR** on the alternate card, and the "is an alternate required?" trigger was decided off an *average* ceiling. Investigation found **four surfaces each reducing per-model data differently**:

| Surface | Category | Numeric ceiling/vis |
|---|---|---|
| Arrival card | majority (user pref) | worst-in-pool |
| Alternate card | **worst (hardcoded)** | **mean** |
| Alternate-requirement | — | **mean** (mislabeled "worst") |
| Forecast map | worst/majority | min/max **or median-across-all** |

Plus two per-model ceiling definitions (`reconcile_ceiling` min vs `best_ceiling` priority-first) and a latent registry bug where a WORST preference silently kept a majority-built advisory aggregate.

## What

One canonical rule, applied to **every** numeric field:
- **majority** → MEDIAN within the winning-category pool (the models that voted for the shown category) — numbers can never contradict the badge;
- **worst** → worst across all models (min ceiling/vis, max wind);
- **wind direction** → circular mean (never min/max/median).

Backend:
- `airport_consensus.consensus` rewritten (pool-median / worst; mean removed); default mode now `majority`.
- `run_alternates`/`_assess`/`_build_alternate` take an `aggregation` arg, wired from the pipeline's `advisory_aggregation` preference → alternate card **and** its regulatory trigger now follow the same mode as the arrival card.
- `best_ceiling` = `min(sounding, nwp)` (fallback cloud_base/lcl), matching `reconcile_ceiling`.
- `registry.evaluate_all` always re-aggregates (fixes the WORST-preference bug).

Web:
- `computeConsensus` → median-in-pool for majority.
- Extracted `web/ts/helpers/airport-summary.ts` (`computeSummaryCondition`); arrival card now median-in-pool (was worst-in-pool), wind speed = median with dir/gust/runway from the representative model.

Consistency guarantee:
- `tests/fixtures/consensus_vectors.json` pins the Python and TS consensus to identical output via parity tests on both sides.

## Tests
- Backend: 3126 passed. Web: 420 passed. New parity + pool-median + registry-bug regression tests included.

## Deferred (accepted)
Alternates fetch the `gfs/icon/ecmwf` trio while the arrival card uses the full pack model set — cards can still differ slightly from the model-set difference alone. Separate follow-up.

### DIFF
--- a/designs/alternate-requirement.md
+++ b/designs/alternate-requirement.md
@@ -70,9 +70,9 @@ without euro_aip installed.
 | Input | Source |
 |---|---|
 | Destination raw TAF (D-0 only) | `snapshot.route_observations.airports[*].taf_raw` (matched on `obs.icao == destination`) |
-| Destination NWP fallback (D-2/D-1) | `RouteAlternates.destination_ceiling_ft` / `destination_visibility_m` — the destination's NWP-consensus assessment at ETA, stored by `run_alternates` (worst across models) |
+| Destination NWP fallback (D-2/D-1) | `RouteAlternates.destination_ceiling_ft` / `destination_visibility_m` — the destination's NWP-consensus assessment at ETA, stored by `run_alternates` under the user's advisory aggregation mode (majority = median of the winning-category pool; worst = worst across models). See "Aggregation mode & conservative bias" below. |
 | Destination ETA | `RouteAlternates.eta` (rounded ETA hour) |
-| Candidate ceiling/vis | A candidate TAF covering its ETA when available (D-0; reused from `route_observations` or gap-fetched), else `AlternateAirport.ceiling_ft` / `.visibility_m` (NWP-consensus, `mode="worst"`). `AlternateQual.source` records which. |
+| Candidate ceiling/vis | A candidate TAF covering its ETA when available (D-0; reused from `route_observations` or gap-fetched), else `AlternateAirport.ceiling_ft` / `.visibility_m` (NWP-consensus under the same aggregation mode). `AlternateQual.source` records which. |
 | Candidate / destination approach class | `best_approach_type` (candidates) / `procedures_query.approaches().most_precise()` (destination) |
 
 The NWP fallback deliberately reuses the alternates stage's own destination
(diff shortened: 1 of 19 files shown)


roznet:
## Review round 1 — addressed in 955fe489

**Critical — regulatory trigger follows the aggregation preference.** This coupling is **intentional and kept** (product decision): the destination NWP fallback follows the user's advisory aggregation mode so the alternate card shows the same category as the airport arrival card. Two mitigations make it safe, and both are now documented + tested:
- **A TAF supersedes the NWP path entirely.** `_build_destination_window` returns the TAF's own worst-case window and never reads the NWP consensus when a TAF covers the ETA. TAFs exist only at D-0 (when the determination matters most), so the aggregation mode only ever affects the no-TAF **D-1/D-2** fallback.
- The whole assessment is explicitly **planning-grade** (no published minima; DH is a proxy).
- Updated the stale `(worst across models)` wording in `models/alternates.py`, `designs/alternate-requirement.md` (new "Aggregation mode & TAF precedence" section) and `designs/alternates.md` (Consensus + Gotchas). Added `TestAggregationModeAffectsTrigger` pinning both the majority-vs-worst verdict flip **and** the TAF-supersedes-NWP precedence.

**Important — registry re-aggregation.** Fixed more than flagged. The unconditional `from_per_model` rebuild (a) discarded convective's synthesized `aggregate_detail` and (b) resurrected fronts' explicit all-`UNAVAILABLE` result into GREEN (`from_per_model` collapses all-UNAVAILABLE → GREEN — a latent bug that also bit the pre-PR majority default). Now: re-aggregate **only when the requested mode differs from the evaluators' majority build-default**, and **never** re-run an all-UNAVAILABLE set. Preserves custom detail under the default, keeps hidden advisories hidden, and still fixes the original worst-preference bug. New test: `test_reaggregation_preserves_custom_detail_and_unavailable`.

**Important — arrival-card parity.** Extended `consensus_vectors.json` parity to `computeSummaryCondition` (the third reducer), and aligned its ceiling/visibility rounding to 1 decimal so arrival card, forecast map and Python are bit-for-bit identical on the vectors.

**Minor — crosswind/headwind.** Added a majority-vs-worst test (`test_crosswind_headwind_follow_mode`).

Backend 3130 passed, web 426 passed.

roznet:
## Review round 2 — addressed in 35464c9e

**Important — headwind worst-case was inverted (real fix).** `weather-map-consensus.ts` had `headwind_kt.worst = _max`, the inverse of the canonical rule this PR codifies on the Python side (`airport_consensus._WORST_IS_MIN`). A positive headwind is favourable, so the conservative "worst" pick is the weakest headwind / strongest tailwind — **flipped to `_min`**. And you're right that the parity fixture didn't cover it: extended `consensus_vectors.json` + both parity tests (Python + TS map) to pin **crosswind_kt/headwind_kt**, so this class of divergence is now caught automatically. (crosswind stays `_max` — magnitude, higher is worse.)

**Minor — severity table consolidated.** `airport-summary.ts` now reuses `CAT_ORDER` from `weather-map-consensus.ts` instead of its own `FLIGHT_CAT_SEVERITY`, so a future category reorder can't update one and miss the other.

Web 426 passed; Python parity green (now including wind components).

This push will trigger another review round; per my 2-round cap I'm handing back rather than auto-looping. Both round-2 items were non-blocking and are now fixed, so this should be clean — recommend merge (`--rebase`) once the round-3 check is green.

PR #349, opened 2026-07-03
Cross-border operational-friction flag on weather alternates (#344)

roznet:
Builds the general **operational-friction** mechanism and ships the **cross-border** flag on it, per #344. A weather-better, close divert candidate can still carry an unplanned international arrival (customs, immigration, GAR, PPR) — this surfaces that as a graded amber/red flag.

## What's in it

- **Extensible flag channel** — `OperationalFlag` + `operational_flags` on `AlternateAirport`, plus `iso_country` plumbed through from the euro_aip model. One rendering path that future signals (water crossing #345, military, drive-time) reuse with no new plumbing.
- **Pure builder** — `analysis/operational_flags.py::cross_border_flag`. Anchors on the **destination** country; severity is purely the country-pair via `euro_aip.borders.crossing_requirements` (both formalities → **red**, exactly one → **amber**, neither → no flag), so membership rules live in the library, not here. **Point-of-entry modulates the message wording only, never the severity.** Missing/unresolvable country → no flag.
- **Rendering** — web ⚠ chip in `renderRouteAlternates` (severity-colored, click → detail popup, reusing the per-row popup pattern) + a short per-row tag in the text digest mirroring the FAA/EASA tags.

## Behaviour (acceptance criteria)

- LFAT (FR) dest + EGMD (GB) alt → **red**, detail mentions customs + immigration; POE → reassurance wording ✓
- FR dest + CH alt, non-POE → **amber** + "could not be verified" wording (not a downgrade) ✓
- FR→IE → amber (immigration only); FR→DE → no flag ✓

## Library dependency

Consumes `euro_aip.borders` (rzflight, already merged + pushed). While building this I simplified `CrossingRequirements` to `(immigration_required, customs_required)` on rzflight main (`ec209f5`): its `from_known`/`to_known` conflated "in a bloc table" with "recognized country" and marked GB/JE (the motivating red cases) as unknown, which is unusable for this consumer.

## Tests

- New `tests/test_operational_flags.py` (acceptance-criteria rows, POE overlay, missing-country) + end-to-end plumbing tests through `run_alternates` in `tests/test_alternates.py`.
- Full suite: **3132 passed, 8 skipped**. TypeScript typechecks (2 pre-existing unrelated errors) and esbuild bundles clean.

## Out of scope

Water crossing (#345), military/joint-use, drive-time, AIP customs free-text parsing, MCP/agent surface.

Closes #344.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #344, opened 2026-07-03
Operational-friction warnings on alternates: mechanism + cross-border grading


roznet:
## Motivation

The weather-alternates list can surface a divert field that's weather-better and close but carries **non-weather operational friction**. Canonical case: a flight to **LFAT** (Le Touquet, `FR`, point-of-entry) offered **EGMD** (Lydd, `GB`, point-of-entry) as an alternate — ~30 nm straight-line, but diverting there is an **unplanned international arrival** (customs, immigration, GAR; Lydd is also strictly PPR-by-phone). "Different country" is too blunt — FR→BE would also be "different country" with essentially none of that friction. We need a **graded** flag, not a binary one.

## Scope (this issue)

Build the general **operational-friction mechanism** and ship the **cross-border** flag on it, consuming the border-membership helper from the library. Water crossing is a follow-up (see companion issue).

**Depends on:** `roznet/rzflight` — *Country border-area membership + crossing-requirements helper* (the `euro_aip.borders.crossing_requirements` API). Membership rules deliberately live in the library, not here, so they aren't re-encoded/drifted.

## 1. Data model — extensible flag channel

Add to `AlternateAirport` (`src/weatherbrief/models/alternates.py`):

```python
class OperationalFlag(BaseModel):
    code: str          # "cross_border"
    label: str         # chip text, e.g. "Cross-border / customs"
    detail: str        # expandable explanation
    severity: str      # "amber" | "red"  (reuse the traffic-light palette)

class AlternateAirport(BaseModel):
    ...
    iso_country: str | None = None
    operational_flags: list[OperationalFlag] = Field(default_factory=list)
```

Plumb `iso_country` through from the euro_aip model (already available; currently unused anywhere in the codebase). This single `operational_flags` channel is reused by future signals (water crossing, military, drive-time) with no new rendering plumbing.

## 2. Cross-border flag builder (pure, no fetch)

In `tasks/alternates.py` (after candidate assembly, near the `AlternateAirport` construction ~line 307), call `euro_aip.borders.crossing_requirements(dest_country, alt_country)` and map to a flag. **Anchor on the destination's country** (the surprise is "planned to land in FR, instead cleared into GB"). Same country, unknown country, or both formalities false → no flag.

**Severity comes purely from the country-pair (not POE):**

| immigration | customs | Severity | Example |
|---|---|---|---|
| ✗ | ✗ | none (no flag) | FR→DE, FR→ES, FR→IT |
[...]

PR #350, opened 2026-07-04
Cross-border flag: model prepared-vs-surprise, not just border-present (#344)

roznet:
Follow-up to the merged #349, reworking **when** and **how hard** the cross-border operational-friction flag fires. The initial logic ("is origin→alternate a border?") over-flagged; this models **what the pilot is not prepared for**, using the flight actually *filed* (origin → destination) as the preparedness baseline.

## Model — three derived reasons

From `euro_aip.borders.crossing_requirements`: `planned = (origin→destination)`, `alt = (origin→alternate)`. Severity = worst reason that fires; none → no flag.

| Reason | Fires when | Severity |
|---|---|---|
| **Unprepared formality** | alt needs an axis (customs/immigration) the filed flight didn't | **red** if alt needs *both*, else **amber** (absolute grade) |
| **Different country than filed** | alt is a different country than the destination, needs a formality, but adds no *new* axis (prepared, wrong country) | **amber** |
| **Facility gap** | alt needs a formality but isn't a listed POE | **amber** ("could not verify") |

Anchored on the **origin** (departure) country — strictly the last departure airport, `route.origin` for single-leg. `point_of_entry` only ever adds the facility note, never changes severity.

## Behaviour

- FR-departed flight diverting to an EU field (`FR→IT`) → **silent** (this was the user-reported false amber).
- Same flight diverting to the UK → **red**.
- `UK→FR` diverting to a German POE → **amber** (prepared for the border, just a different country than filed), *not* red.
- `UK→FR` diverting to a French field that isn't a POE → **amber** (verify facilities).

Detail wording expands ISO codes to full country names via euro_aip's `CountryMapper`.

## Known limitation (documented, not coded around)

`euro_aip.borders` models Schengen + EU-customs only, so the **IE↔GB Common Travel Area** isn't represented — an Ireland↔UK pair reads as a full border and over-flags red. Accepted as a conservative over-warn.

## API / tests

`cross_border_flag(origin, destination, alt, alt_is_poe)`; `tasks/alternates.py` resolves the destination country alongside the origin (both wrapped, degrade to no-flag). Full suite **3157 passed**; the truth table above is pinned in `test_operational_flags.py`, plus `run_alternates` plumbing tests including one that proves the destination country is threaded (amber-not-red).

Refs #344.

### DIFF
--- a/designs/alternates.md
+++ b/designs/alternates.md
@@ -287,19 +287,35 @@ A weather-better, close divert candidate can still carry **non-weather** frictio
 future signals (water crossing #345, military/joint-use, drive-time) reuse the
 same shape with no new rendering plumbing.
 
-The cross-border flag is built by the pure `analysis/operational_flags.py::cross_border_flag(origin_country, alt_country, alt_is_poe)`,
-called from `_build_alternate`. It **anchors on the origin (departure) country,
-not the destination** — the formalities you face on landing at an alternate are
-set by the country you departed from, since that is the border the flight
-actually crosses. Destination-anchoring is wrong whenever origin ≠ destination
-(e.g. an FR→CH flight diverting to an Italian field is FR→IT — both EU, no
-formalities — even though CH→IT would read customs-required). Severity comes
-**purely from the country-pair** via `euro_aip.borders.crossing_requirements`
-(membership rules live in the library, not re-encoded here): both formalities →
-`red`, exactly one → `amber`, neither / same country / unresolved country → no
-flag. `point_of_entry` modulates the detail **wording only** (reassurance vs
-"could not verify entry facilities"), **never** the severity. `OperationalFlag`
-carries `code` / `label` / `detail` / `severity` (`amber`|`red`, never green).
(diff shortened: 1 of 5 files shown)


roznet:
Reworked to the full prepared-vs-surprise model we brainstormed (three derived reasons: unprepared formality → absolute red/amber, different-country-than-filed → amber, facility gap → amber). `cross_border_flag` now takes the destination too. Behaviour confirmed against the truth table: FR→IT silent, domestic→UK red, UK→FR→DE amber (softened), FR→FR non-POE amber. IE↔GB CTA over-flag documented as a known limitation. Full suite 3157 passed. Title/body updated.

roznet:
Merged (rebase). The two Minor findings were applied directly on main in `76ea6568` (no blockers, so kept off the PR to avoid another review round):
- `_country_name` now keeps connector words lower-case ("Isle of Man", "Bosnia and Herzegovina") instead of naive `.title()`.
- Renamed the CH→FR test to `test_switzerland_departed_diverting_to_uk_is_red` + fixed the docstring — CH is the origin (departed), bound for France.

PR #351, opened 2026-07-04
iOS: auto-download briefings for offline + age-based cache eviction

roznet:
Rebased onto current `main` (was 53 commits behind; clean, single commit, no conflicts) and opened for review — this branch had never been PR'd.

## What it does
Auto-caches the full briefing bundle when a briefing is displayed, so Skew-Ts and the whole pack open instantly and offline instead of paying a per-tap sounding-profile round-trip.

- **Offline Auto-Download setting** (`AppSettingsStore.autoDownloadMode`: off / Wi-Fi only / Wi-Fi & cellular; default Wi-Fi only, persisted in UserDefaults). Surfaced as a picker in Settings.
- **NetworkMonitor** — `NWPathMonitor` wrapper exposing `isConnected` / `isOnWiFi` to gate auto-download by connectivity.
- **`BriefingViewModel.maybeAutoDownloadLatest()`** — on briefing display and after a refresh produces a new pack, downloads the latest pack for offline use when opted in and the flight is today or later. Reuses `downloadCurrentPack` + its progress banner; no-op when already cached or a download is in flight.
- **Sounding profiles now cached** — every `(point, model)` sounding profile is written from the bundle under keys matching `soundingProfile()`, so a downloaded pack renders Skew-Ts fully offline.
- **Cache eviction** — `CachingBriefingRepository.pruneStalePacks(olderThanDays:)` drops packs whose flight departed > 7 days ago (`CachedPackEntry.departureTime`, backward-compatible, falls back to cached flight record). Runs on launch / foreground.

Design docs updated (`ios-app-overview`, `ios-app-data-models`).

## Tests
`AutoDownloadMode` connectivity gate; `CachedPackEntry` legacy decode + `departureTime` round-trip.

## Notes for review
- Not true background refresh — auto-download runs only while the briefing screen is open / on foreground, not via `BGTaskScheduler`.
- iOS-only; no backend changes, so independent of the pending web deploy.

### DIFF
--- a/app/flyfun-weather/flyfun-weather/App/AppState.swift
+++ b/app/flyfun-weather/flyfun-weather/App/AppState.swift
@@ -50,6 +50,11 @@ final class AppState {
     private(set) var repository: (any BriefingRepository)?
     let pirepOfflineStore = PirepOfflineStore()
     let userPreferences = UserPreferencesStore()
+    /// Local device settings (e.g. offline auto-download mode). Server-side flags
+    /// live in `userPreferences`; this is the on-device counterpart.
+    let settings = AppSettingsStore()
+    /// Live network reachability — gates Wi-Fi-only auto-download.
+    let networkMonitor = NetworkMonitor()
     /// (i)-popup help content (metrics + advisories). Seeded from disk cache or
     /// the bundled baseline at init; refreshed opportunistically when online.
     let helpCatalog = HelpCatalogStore()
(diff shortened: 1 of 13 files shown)


roznet:
Addressed the review in af4e65ea:

**Important (blocker) — fixed**
- `pollRefreshStatus()` now calls `checkCacheStatus()` + `maybeAutoDownloadLatest()` after adopting a new pack, mirroring `refresh()`. The "refresh already running from web/another device" path (via `checkActiveRefresh`) now auto-downloads for opted-in pilots instead of waiting for the next screen entry.

**Minor — all batched in (pushing for the blocker anyway):**
- **Duplicated ISO-8601 parsing** → hoisted the fractional-tolerant parser into `Date.parseISO8601` (in `Utilities/Extensions.swift`); `BriefingViewModel` and `CachingBriefingRepository` now share it.
- **Orphaned sidecar metadata on eviction** → `pruneStalePacks` now drops the whole flight directory once no index entry references the flight (new `BriefingCacheStore.hasCachedPacks` / `removeFlightDirectory`). Safe because offline recovery only walks index entries. Since a flight's departure is shared across its packs, they all age out in the same run, so the flight goes fully unreferenced.
- **No eviction test coverage** → extracted a pure `CachingBriefingRepository.isPackStale` (cutoff + legacy-fallback logic) and added `PackStalenessTests` covering it, plus a store-level test for the sidecar cleanup + the `f1`/`f10` prefix guard. (Testing the full repo directly is still blocked by the keychain-backed auth stack it wraps — the pre-existing deferral noted in `ServiceTests.swift` — so the logic was made pure instead.)

Test build + the touched suites (`PackStalenessTests`, `BriefingCacheStoreTests`, `CachedPackEntryTests`) pass on iOS 17 sim. Nothing deferred.

Issue #352, opened 2026-07-05
Flexibility: first-time explainer modal + (i) button + usage-count gate + analytics


roznet:
## Summary

Add a **first-time explainer modal** for the Flexibility / timing-scenario feature (PR #341), shown the first time a user selects a scan mode. Make the same explainer permanently reachable via an **(i) button** on the Timing Scenarios section of a briefing, and add a **product-analytics event** so flexibility usage shows up in the analytics digest.

The feature is **web-only** (iOS timing-scenario surfacing was out of scope in v1 — no Swift changes here).

### Why

Flexibility scanning is computationally and data-intensive (background ECMWF daylight decode + on-tap multi-model confirm). Pilots should understand what they're triggering, that it runs in two steps, that it runs in the background, and that it should be used deliberately — not out of curiosity. Today the toggle has no explanation.

---

## The gate — how we decide to show it

Two signals combined:

```
showExplainer = (userTimeScanCount === 0) && !sessionStorage['wb_flex_explainer_ack']
// on "Got it": sessionStorage['wb_flex_explainer_ack'] = '1'
```

- **Durable gate — all-time `time_scan` count for the user.** We already log one `api_usage_log` row per executed scan (`log_api_usage(service="time_scan", pipeline=<mode>, user_id, flight_id)` at `src/weatherbrief/tasks/time_scan_runner.py:181`). Once the user has genuinely run ≥1 scan, count ≥ 1 → never show again. This is per-account (survives devices / cache-clear) and doubles as the substrate for a future `time_scan` rate limit and a future heavy-user message (not in scope here).
- **Ephemeral guard — `sessionStorage`.** The count only flips to ≥1 *after a scan executes* (background job), so between "user picks a mode" and "scan ran" the count is still 0 and a pure count gate would re-fire on every dropdown change / editor reopen. `sessionStorage['wb_flex_explainer_ack']`, set on acknowledge, suppresses re-show for the rest of the sitting. **Deliberately `sessionStorage`, not `localStorage` and not an in-memory flag:**
  - in-memory resets on navigation — this is a multi-page app (separate HTML entrypoints for flights-list / flight-detail / briefing), so it would double-show across pages;
  - `localStorage` would suppress forever even if the user never actually ran a scan, losing the "keep gently re-informing until genuinely used" property the count gives us.

**Walk-through:**
[...]

PR #353, opened 2026-07-05
Add flexibility-explainer first-time modal and gate (#352)

roznet:
## Summary
Implements a first-time explainer modal for the Flexibility (timing-scenario) feature, shown only when a user picks a scan mode for the first time. The explainer is gated by a durable per-account "has used timing scan" flag and a session-scoped acknowledgment to prevent re-firing. The same HTML is reused by an always-reachable (i) popup on the Timing Scenarios header.

## Key Changes

**Backend (usage tracking & API)**
- Added `user_has_used_time_scan()` and `count_user_time_scans()` helper functions to query `ApiUsageRow` for `time_scan` service entries
- Extended `UsageSummary` model with `time_scan_used` (boolean) and `time_scan_count` (integer) fields to surface the durable gate substrate
- Updated `get_usage_summary()` to populate these fields from the database
- Added comprehensive unit tests for the time-scan usage helpers and summary integration

**Frontend (explainer component & integration)**
- Created `flexibility-explainer.ts` with:
  - Shared HTML content (`flexibilityExplainerHtml`) used by both the first-time modal and (i) popup
  - `shouldShowFlexibilityExplainer()` pure gate predicate (unit-tested): shows only when `!timeScanUsed && !sessionAcked`
  - `flexExplainerAcked()` / `ackFlexExplainer()` for session-scoped acknowledgment via `sessionStorage`
  - `openFlexibilityExplainer()` to display the popup and wire the "Got it" button
  - `maybeShowFlexibilityExplainer()` async trigger that fetches usage summary and conditionally opens the modal
- Wired first-time trigger in `flights-main.ts`: listens to flexibility dropdown change and calls `maybeShowFlexibilityExplainer()` when a non-`none` mode is selected
- Wired always-reachable (i) button in `briefing-main.ts`: added `.flex-explainer-btn` class detection to open the explainer directly
- Updated `preferences-adapter.ts` to include the new `time_scan_used` and `time_scan_count` fields in the `UsageSummary` interface
- Added product-analytics event `TIMING_SCENARIOS_USED` (emitted once per briefing when scan results render, with `mode` prop)
- Updated `flight-detail-ui.ts` to call `maybeShowFlexibilityExplainer()` on flexibility selection in the flight editor

**Styling & HTML**
- Added CSS for `.flex-explainer` and `.flex-explainer-actions` in `style.css` (heading, spacing, button alignment)
- Updated `briefing.html` to add `.flex-explainer-btn` class to the Timing Scenarios (i) button

**Testing**
- Added `flexibility-explainer.test.ts` with unit tests for the gate predicate covering all four combinations of `timeScanUsed` and `sessionAcked`

## Implementation Details

- **Session vs. durable gating**: The session flag (`sessionStorage`) prevents re-fire while the count is still 0 between "picks a mode" and "scan ran"; the durable flag (`time_scan_used` from usage summary) is the long-term gate that survives browser restart
- **Graceful degradation**: `sessionStorage` access is wrapped in try-catch to handle private-mode browsers
- **Shared content**: Both the first-time modal and (i) popup render identical HTML via `showPopupContent()` to prevent copy drift
- **Non-destructive acknowledgment**: "Got it" closes the modal and sets the session flag; the user's dropdown selection stands (no revert)

https://claude.ai/code/session_01FCnL59YxvS7w8q86oPjvPi

### DIFF
--- a/src/weatherbrief/analytics/events.py
+++ b/src/weatherbrief/analytics/events.py
@@ -49,6 +49,9 @@ class Event(StrEnum):
     COMPARE_OPENED = "compare.opened"
     COMPARE_AIRPORT_ADDED = "compare.airport_added"
     ALTITUDE_TABLE_OPENED = "altitude_table.opened"
+    # Flexibility / timing-scenarios feature. Emitted once per briefing when
+    # scan results actually render (props: {mode}).
+    TIMING_SCENARIOS_USED = "timing_scenarios.used"
 
     # Onboarding -------------------------------------------------------------
     TOUR_STARTED = "tour.started"
@@ -79,6 +82,7 @@ FEATURE_OF: dict[str, str] = {
     Event.COMPARE_OPENED.value: "compare",
     Event.COMPARE_AIRPORT_ADDED.value: "compare",
     Event.ALTITUDE_TABLE_OPENED.value: "altitude_table",
+    Event.TIMING_SCENARIOS_USED.value: "timing_scenarios",
     Event.AUTO_REFRESH_ENABLED.value: "auto_refresh",
     Event.BRIEFING_REFRESH_REQUESTED.value: "manual_refresh",
 }
@@ -92,6 +96,7 @@ KNOWN_FEATURES: tuple[str, ...] = (
     "skewt",
     "compare",
     "altitude_table",
+    "timing_scenarios",
     "auto_refresh",
     "manual_refresh",
     "detailed_mode",
--- a/src/weatherbrief/api/usage.py
+++ b/src/weatherbrief/api/usage.py
@@ -53,6 +53,12 @@ class MonthUsage(BaseModel):
 class UsageSummary(BaseModel):
     today: TodayUsage
     month: MonthUsage
+    # Durable, all-time "has this user ever run a timing scan?" flag. Backs the
+    # web flexibility-explainer gate (show the first-time modal only until the
+    # user has genuinely run ≥1 scan). Count is surfaced too for future
+    # heavy-user / rate-limit triggers (not enforced here).
+    time_scan_used: bool = False
+    time_scan_count: int = 0
 
 
 # --- Core functions ---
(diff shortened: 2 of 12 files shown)


roznet:
Thanks for the review. Triaged and addressed in `0403c7b`:

**Fixed (Important #1 + #2, one root cause):** The session-ack flag was only set by the "Got it" button, so `×`/`Esc`/backdrop dismissals left it unset — re-firing the modal (and refetching `/user/usage`) on every later non-`none` flexibility change; established users (`time_scan_used=true`) also re-hit `/usage` on every toggle. `maybeShowFlexibilityExplainer()` now sets the ack flag unconditionally right after evaluating the gate, so it resolves **exactly once per session**: any dismissal path suppresses re-firing, and the durable-used case short-circuits before the network call on later toggles (O(1) per session). A new session still re-evaluates via the durable count.

**Won't fix — #3 (`user_has_used_time_scan`):** This was explicitly requested by issue #352 ("Add `user_has_used_time_scan(db, user_id) -> bool` — existence check, **not** a COUNT") as the fast, index-narrowed existence check meant to back a future `time_scan` rate limit and heavy-user message (both listed as out-of-scope follow-ups in the issue). `get_usage_summary()` uses the COUNT because it needs `time_scan_count` anyway; the existence helper stands as documented substrate for those follow-ups, so I'm keeping it per the issue spec.

**Deferred — #4 (i18n):** Agree for consistency, but the app ships four full locales (en/fr/de/es) and this rich copy would need real fr/de/es translations, not just English keys. The issue sanctioned English copy ("tune wording freely"), so I'd rather land this as a dedicated follow-up with proper translations than ship English strings under localized keys. Happy to open a tracking issue if you'd like.

Frontend unit tests + flights/flight/briefing bundles rebuild clean.

---

roznet:
Round 2 addressed in `74ccd39`:

**Fixed (Important — re-entrancy race):** Good catch. The round-1 fix set the session-ack flag *after* `await fetchUsageSummary()`, leaving a gap where two `change` events firing before the fetch resolves (e.g. arrow-keying the flexibility `<select>`) both passed the sync guard and both opened the modal. Moved `ackFlexExplainer()` **before** the await so a concurrent second call bails at the sync `flexExplainerAcked()` check. Added regression tests: the concurrent-invocation path (asserts a single `/usage` fetch + single modal open), plus the already-used and already-acked branches. Full unit suite green (433 tests), bundles rebuilt.

**Minor — `user_has_used_time_scan` (re-raise of round-1 #3):** Position unchanged. It was explicitly required by issue #352 ("Add `user_has_used_time_scan(db, user_id) -> bool` — existence check, **not** a COUNT") as the fast, index-narrowed path meant to back a future `time_scan` rate limit / heavy-user message — both listed as out-of-scope follow-ups in the issue. `get_usage_summary()` uses the COUNT because it needs `time_scan_count` in the same query; there is no boolean-only call site *today* by design. I'm keeping it as documented substrate per the issue spec rather than removing spec'd code or adding an artificial call site.

That's two review rounds with real fixes landed. Handing back to @roznet for the merge decision — I won't iterate further automatically, but I'll continue watching the PR (CI + any new comments).

---

Issue #357, opened 2026-07-05
iOS: port Timing-Flexibility (Timing Scenarios) to the app


roznet:
Bring the **Timing-Flexibility / Timing Scenarios** feature to the iOS app at display parity with web. Backend + web are fully shipped (PR #341, all 4 slices); iOS has **zero** timing code today. This is a pure client port — **no backend work**, every endpoint/DTO is stable.

Full plan: `designs/plans/timing-scenario-ios-port.md`.

### Locked decisions
- **Placement:** a section inside the **Advisory tab** (`AdvisoryTabView`), mirroring web's inline placement — not a new `BriefingTab`.
- **Offline:** **online-only for v1** (timing data is a separate endpoint, not in the offline bundle; poll live, placeholder when offline).
- **Flexibility control:** in the **flight editor** (`AddFlightView`), per flight — not global Settings.

### Slices
- [ ] **Slice 1 — Model + Flexibility control.** Decode `flexibility` + `alt_departure_time` on `FlightResponse`/`CreateFlightRequest`; Flexibility `Picker` + net-new alt-departure `DatePicker` in `AddFlightView`; decode `time_scan_used`/`time_scan_count` from `/usage`; first-time explainer sheet.
- [ ] **Slice 2 — DTO decode + polling.** Swift Codable mirrors of the `time_scan.py` family; `@Published timeOptions` on `BriefingViewModel`; poll loop (backoff 3s→×1.5→cap 15s, poll-until-terminal, 404=no-data, tolerate 3 transient errors), online-only placeholder.
- [ ] **Slice 3 — Timing Scenarios panel.** State ladder (hidden/running/failed/skipped/results); candidate rows (day suffix, assessment pill, shift, improves/worsens, confidence line); expandable per-advisory dot-table (Current vs ECMWF-only vs All-models — verify iOS `AdvisoriesResponse` carries `per_model`); footnotes.
- [ ] **Slice 4 — Confirm / "Check all models" + Set-as-alternate.** `POST …/confirm` with client-side 429 one-at-a-time guard (disable other confirm buttons while any `confirm_pending`); "Check all models" on `ecmwf_only` rows; "Set as alternate" → PATCH + rescan.

### Notes
- No hard rate limit exists yet (`count_user_time_scans` is unenforced) — only the 429 one-confirm-at-a-time guard.
[...]

PR #358, opened 2026-07-06
Port timing scenarios (Flexibility) feature to iOS (#357)

roznet:
Implements the full Flexibility (timing-scenario) feature on iOS, mirroring the web implementation. This allows pilots to grade alternative departure times against their briefing and surface calmer windows.

## Summary

Adds the complete timing-scenario scanning workflow to the iOS briefing view:
- **Timing Scenarios panel** in the Advisory tab showing scan results, candidate windows, and per-advisory impact details
- **Flexibility mode picker** in the flight editor (none/day/alternate) with pinned alternate-time support
- **First-time explainer** modal gating the feature until a scan is actually run
- **Background polling** of scan status with exponential backoff and error tolerance
- **Multi-model confirmation** of provisional ECMWF-only candidates with full model coverage
- **Offline placeholder** when connectivity is lost during polling

## Key Changes

**New Views:**
- `TimingScenariosView.swift` — Main panel rendering the full state ladder (hidden → offline → running → failed → skipped → results)
- `CandidateRow` — Individual departure-time card with assessment, shift label, confidence line, and expandable detail table
- `TimingDetailTable` — Per-advisory dot-table comparing current vs candidate vs confirmed grades
- `FlexibilityExplainer.swift` — Reusable first-time/on-demand explainer sheet

**New Models:**
- `TimeOptionsResponse.swift` — Complete DTO family mirroring server `time_scan.py` (status, scan, candidates, confirmations, window metadata)
- `UsageSummaryResponse.swift` — Account usage flags for gating the first-time explainer
- `FlexibilityMode` enum — none/day/alternate modes with display labels

**ViewModel Updates:**
- `BriefingViewModel` — Timing-scenario polling loop with backoff, confirm action, and `anyConfirmPending` flag for button disabling
- `AddFlightViewModel` — Flexibility mode picker, alternate-departure editor, first-time explainer gate, and edit-detection for mode/alt-time changes

**View Updates:**
- `AddFlightView` — New Flexibility section with mode picker, alt-time controls (date/hour/minute/TZ), and explainer button
- `AdvisoryTabView` — Conditional rendering of TimingScenariosView when `showsTimingScenarios` is true

**Repository & Infrastructure:**
- `BriefingRepository` protocol — New `timeOptions()` and `confirmTimeOption()` methods
- `CachingBriefingRepository` — Online-only implementations (no caching for live polling)
- `FixtureBriefingRepository` — Test fixtures for timing scenarios
- `FlightResponse` — Added `flexibility` and `altDepartureTime` fields with `effectiveFlexibility` helper

## Implementation Details

- **Polling strategy**: Exponential backoff (3s → ×1.5 → cap 15s) with up to 3 transient-error tolerance; stops on terminal status (done/failed/skipped) or when any candidate is `confirm_pending`
- **Offline handling**: Detects loss of connectivity and shows placeholder; resumes polling when reconnected
- **Model coverage labels**: Honesty ladder (confirmed-in-window → ECMWF-only → confirmed) with "Check all models" affordance on provisional rows
- **Alternate-time validation**: Same-day only, matching web behavior
- **First-time explainer**: Durable `timeScanUsed` flag from `/usage` gates modal; session-scoped ack prevents re-firing within same app run
- **Time formatting**: All times display in UTC (matching web); ISO-8601 strings decoded as-is to avoid date-strategy conflicts
- **Accessibility**: Info button labels, dot status tooltips, and semantic HTML structure

The feature is fully integrated with the existing briefing refresh cycle and respects flight editability (owner-only actions hidden for subscribers).

https://claude.ai/code/session_01DWXNXXpKCjwtUEcJyVYpG5

### DIFF
--- a/app/flyfun-weather/flyfun-weather/Models/API/CreateFlightRequest.swift
+++ b/app/flyfun-weather/flyfun-weather/Models/API/CreateFlightRequest.swift
@@ -12,6 +12,11 @@ struct CreateFlightRequest: Encodable {
     /// Flight profile to associate; the server fills any unspecified flight
     /// fields (ceiling, speed, model choices) from this profile's settings.
     var profileId: Int? = nil
+    /// Timing-scenario Flexibility mode. `.alternate` is rejected on create
+    /// (it needs `altDepartureTime`, set via PATCH after creation — mirrors the
+    /// web create-then-patch flow), so the create picker only offers the day
+    /// modes. Omit for no scan.
+    var flexibility: FlexibilityMode? = nil
 }
 
 /// Request body for editing an existing flight via PATCH /api/flights/{id}.
@@ -26,6 +31,12 @@ struct UpdateFlightRequest: Encodable {
     var flightCeilingFt: Int? = nil
     var flightDurationHours: Double? = nil
     var profileId: Int? = nil
+    /// Timing-scenario Flexibility mode; omit for no change. `.alternate`
+    /// requires `altDepartureTime` (server 422s otherwise).
+    var flexibility: FlexibilityMode? = nil
+    /// Pinned alternate departure (ISO 8601), or `""` to clear it. Omit for no
+    /// change. Paired with `flexibility == .alternate`.
+    var altDepartureTime: String? = nil
 }
 
 /// How much of the briefing an edit invalidated, returned alongside the updated
--- a/app/flyfun-weather/flyfun-weather/Models/API/FlightResponse.swift
+++ b/app/flyfun-weather/flyfun-weather/Models/API/FlightResponse.swift
@@ -25,6 +25,14 @@ struct FlightResponse: Codable, Identifiable, Sendable {
     let latestBriefing: BriefingStatusInfo?
     /// Owner vs subscriber. Absent on older servers — treated as owner.
     let role: FlightRole?
+    /// Timing-scenario Flexibility mode. Absent on older servers → treated as
+    /// `.none`. `.alternate` grades the single `altDepartureTime`; the day modes
+    /// run the departure-window scan. See `TimeOptionsResponse`. Placed last so
+    /// the memberwise initializer's existing call sites only append two args.
+    let flexibility: FlexibilityMode?
+    /// The pinned alternate departure (ISO 8601), used when `flexibility` is
+    /// `.alternate` or as the "★ your alternate" row of a day scan. nil = none set.
+    let altDepartureTime: String?
 
     /// Parsed departure date for display.
     var departureDate: Date? {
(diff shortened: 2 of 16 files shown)


roznet:
Thanks for the review — addressed all three in `c0763e4`:

**1 (Important) — stale timing panel on poll give-up.** `pollTimeOptions` now calls `abandonTimeOptionsPolling()` when the transient-error budget is exhausted: it clears `timeOptions` (mirroring the web's `set({ timeOptions: null })`) so a stale "this window looks smoother" panel can't linger, or shows the offline placeholder when the failure is a mid-poll connectivity drop (addressing the "the `isConnected` check only runs once" point).

**2 (Important) — `TimeScanStatusDTO` resilience.** Given a hand-written tolerant `init(from:)` matching every sibling DTO, so a missing/renamed key degrades that field instead of failing the whole `TimeOptionsResponse` decode. `status` defaults to the terminal `.done` so schema drift can't spin the poll forever.

**3 (Minor) — tests.** Added `TimingScenariosTests` covering the tolerant decoders (full payload + minimal-required + empty-object status), the snake_case↔camelCase mapping, wire-enum values vs `models/time_scan.py`, and the usage gate.

**Deferred (noted, not a regression):** auto-restarting the poll on connectivity *restore*. The stale-data hazard is fixed (panel now honestly shows the offline placeholder rather than stale grades), and a pack switch / briefing refresh restarts polling. A reconnect/foreground observer is an enhancement for a follow-up rather than this PR.

---

roznet:
All four addressed in `ed91e3a`:

**1 (Important) — stale `alt_departure_time` on mode switch.** Confirmed against `flight-main.ts:318-320`: the web always sends `""` when `flexibility !== 'alternate'`. iOS now does the same (`"" ` to clear instead of `nil`/omit), so switching away from `.alternate` no longer leaves a stale alt that resurfaces as a spurious "★ your alternate" tag. (The separate "Set as alternate" pin path is unaffected — it sends only `alt_departure_time`.)

**2 (Important) — stale panel on pack switch.** `startTimeOptionsPolling` gained a `resetExisting` flag — `true` on pack load/switch/refresh (drops the prior pack's scan up front, mirroring the web's null-on-pack-install), `false` on same-pack confirm/set-alternate re-polls so their pending state doesn't flash away.

**3 (Important) — `TimeWindowScanDTO.baseline` non-tolerant.** Wrapped in `try?` with a `TimeBaselineDTO.empty` default, matching the file's contract — a missing `baseline` no longer fails the whole decode.

**4 (Minor) — one bad candidate dropped the whole list.** Candidates now decode through a never-throwing `LenientCandidate` wrapper, so a malformed element drops only itself instead of collapsing to a false "no better window found".

Tests extended for the tolerant baseline + lenient candidate array.

**Deferred:** the `/sync-ios-web` drift audit you flagged is a good call — I'll leave that to @roznet to run/schedule since it's a broader cross-client pass beyond this PR's diff.

Still worth a local `xcodebuild test` before merge — this environment has no Swift toolchain, so the changes (incl. the new tests) are reviewed but not compiler-verified.

---

roznet:
Merged to `main` by rebasing this branch's commits (now `66143b59`..`e2781f1f`) on top of main. GitHub didn't auto-mark it merged because the rebase changed the SHAs.

Verified locally (cloud couldn't): `xcodebuild` build + full test suite green on iPhone 17 sim.

Also addressed the **third** review round's findings on `main` in `77dd586f`:
- `FlightResponse.flexibility` now decodes tolerantly (unknown wire value → `.none`) so it can't fail the whole flights-list decode on the non-tolerant live path.
- `AddFlightViewModel` only sends `altDepartureTime=""` when actually leaving `.alternate` mode — an unrelated edit no longer wipes a day-scan flight's pinned '★ your alternate'.
- `pollTimeOptions` got an iteration cap + `stopTimeOptionsPolling()` wired to `BriefingContainerView.onDisappear` so a running scan can't keep the view model alive/polling after leaving the briefing.
- Added tests for the tolerant flexibility decode and the alt-time preserve/clear PATCH payload.

Deferred (noted in round 3): auto-restart poll on connectivity restore, and the shared web `flight-main.ts` alt-clear pattern — follow-ups, not blockers.

Issue #359, opened 2026-07-06
iOS: cache-first cold start for flight list (no spinner when cached list exists)


roznet:
## Problem

Every cold start of the iOS app shows a full-screen spinner (the "rotating wheel") before the flight list appears, even though a complete cached copy of the list is already on disk. The app blocks on the live `/api/flights` request before painting anything.

### Why it happens

Two independent facts combine:

1. **Fresh ViewModel each open.** `FlightListView.task` creates a brand-new `FlightListViewModel` (state `.idle`) on every appearance and calls `loadFlights()`. Because there is no in-memory list, `loadFlights()` takes the `.loading` branch and shows the full-screen spinner (`FlightListViewModel.swift:41-48`).

2. **Read path is network-first.** `CachingBriefingRepository.flights()` always `await`s `online.flights()` and only reads the cached `flights.json` **if that call throws** (`CachingBriefingRepository.swift:85-111`). The cached list is treated purely as an offline fallback, never as a fast-open source.

So the data to paint the list instantly (`BriefingCache/flights.json`, rewritten after every successful fetch) already exists, but is never used on the happy path.

### Precedent already in the codebase

The "keep the list on screen, sync in the background" pattern (stale-while-revalidate) is **already implemented for the warm/refresh path** (issue #1): when the scene re-activates with a list already loaded, `loadFlights()` keeps it visible and shows a subtle `isRefreshing` indicator instead of a spinner (`FlightListViewModel.swift:37-49`). This issue extends that same idea to **cold start**.

## Non-goals / rejected approaches

- **Do NOT make `flights.json` the source of truth.** The server remains authoritative; the cache is a fast first paint that is always revalidated against the network.
- **Do NOT drop the network call.** We still fetch `/api/flights` on every open and swap in fresh data — we just stop *blocking the first paint* on it.
- **Not related to #356.** #356 is a disk-leak/cleanup concern on the sidecar *write* side; this is the *read* side. They share the cache subsystem but do not overlap — neither fixes the other.

## Proposed implementation

Goal: on cold start, if a cached list exists, paint it immediately and revalidate in the background; only show the full-screen spinner when there is genuinely nothing cached.

Two seams, mirroring the existing #1 pattern:

1. **`CachingBriefingRepository` — expose a cache-only read** so the ViewModel can seed without a network round-trip. Add:
   ```swift
   /// Cached flight list from `flights.json`, if present. No network.
   func cachedFlights() async -> [FlightResponse]? {
[...]

PR #361, opened 2026-07-06
Cache-first cold start for flight list (#359)

roznet:
## Summary
Implements cache-first loading for the flight list on cold start, eliminating the full-screen spinner when a cached list is available. The cached list paints instantly while a fresh fetch happens in the background, improving perceived performance on app reopen.

## Changes

- **FlightListViewModel**: Added cache-first logic in `loadFlights()` that seeds the UI with cached flights before fetching fresh data from the network. If a cached list exists and is non-empty, it's painted immediately in `.loaded` state, then the network fetch proceeds and replaces it (or the offline fallback keeps it if the network fails).

- **CachingBriefingRepository**: 
  - Changed `online` property type from concrete `OnlineBriefingRepository` to protocol type `any BriefingRepository` to enable test doubles
  - Added `cachedFlights()` method to read the cached flight list without network access, supporting the cache-first seed path

- **CachingBriefingRepository+Testing**: New testing helper that builds a `CachingBriefingRepository` with injected online and cache doubles, avoiding the need to construct a full `APIClient` in tests.

- **TestSupport**: 
  - Added `beforeFlightsReturn` hook to `MockBriefingRepository` to gate network calls, allowing tests to observe intermediate ViewModel state
  - Added `TestGate` actor to suspend async operations until explicitly released, enabling tests to inspect the UI after cache seed but before network resolution

- **ViewModelTests**: Added three comprehensive tests covering the cache-first cold start scenarios:
  - `coldStartSeedsCachedListThenSwapsFresh`: Cached list paints immediately with subtle refresh indicator, then swaps to fresh data
  - `coldStartWithoutCacheShowsSpinnerThenLoads`: True first run (no cache) still shows full-screen spinner
  - `coldStartCachePresentNetworkThrowsStaysOffline`: Network failure keeps cached list on screen with offline indicator

## Implementation Details

The cache-first logic is gated on `state == .idle` to only apply on the initial load, not on subsequent refreshes. The server remains authoritative — the network result always replaces the seed. If the network fails, the offline fallback keeps the cached list with `isOffline` set, never showing an error state or empty list.

https://claude.ai/code/session_018Wv195EpSrRm9goLKeEqES

### DIFF
--- a/app/flyfun-weather/flyfun-weather/Services/CachingBriefingRepository+Testing.swift
+++ b/app/flyfun-weather/flyfun-weather/Services/CachingBriefingRepository+Testing.swift
@@ -1,0 +1,28 @@
+#if DEBUG
+import Foundation
+import FlyFunCommon
+
+extension CachingBriefingRepository {
+    /// Build a repository around an injected `online` double and a temp-dir
+    /// cache, for unit tests of the offline/cache-first read paths (#359).
+    ///
+    /// The `client` is an inert stub `APIClient` (dummy base URL, in-memory
+    /// token store, never authenticated). The flight-list read path exercised by
+    /// these tests goes through `online` and `cache` only and never touches the
+    /// client, so it stays a placeholder. Lives in the app module (guarded by
+    /// `DEBUG`) because the test target doesn't link `FlyFunCommon`.
+    static func makeForTesting(
+        online: any BriefingRepository,
+        cache: BriefingCacheStore
+    ) -> CachingBriefingRepository {
+        let store = InMemoryBearerTokenStore()
+        let session = RollingBearerSession(store: store)
+        let client = APIClient(
+            baseURL: URL(string: "https://tests.invalid")!,
+            tokenStore: store,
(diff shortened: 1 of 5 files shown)


roznet:
Merged to `main` by rebasing onto current main (now `6f103e4f`) — GitHub didn't auto-mark it merged because the rebase changed the SHA. Verified locally with `xcodebuild` build + full test suite green on iPhone 17 sim.

While verifying I caught a **latent compile error the cloud build never hit**: the cold-start tests called the actor-isolated `TestGate.open()` without `await`. Fixed on main in `951269fc`, along with both review minors:
- Extracted the shared `flights.json` read into `readCachedFlightsFromDisk()` (used by `cachedFlights()` + the `flights()` offline fallback).
- Populate `cachedFlightIds` during the cold-start seed (via `refreshCachedFlightIds()`) so downloaded flights don't flash as not-available-offline before the fetch lands — sequenced after `isRefreshing` is set so the seeded list always paints with the subtle indicator.

Issue #364, opened 2026-07-08
iOS App Intents — Siri / Shortcuts / Spotlight (Tier 1, iOS 26)


roznet:
Add the Tier-1 App Intents surface so pilots can drive FlyFun from Siri, Shortcuts, Spotlight, and the Action Button on **current iOS 26** — no Apple-Intelligence dependency. This is the on-device sibling of the MCP tool catalog (`src/weatherbrief/mcp/server.py`).

**Design:** [`designs/ios-app-intents.md`](https://github.com/roznet/flyfun-weather/blob/main/designs/ios-app-intents.md) — Phase 1.

### Scope
- **Entities:** `FlightEntity` (`AppEntity` + `IndexedEntity`), `AirportEntity`, backed by `CachingBriefingRepository` + `AirportDatabase`.
- **Intents:** `OpenFlightListIntent` ("show my FlyFun briefings"), `OpenBriefingIntent` (next-flight default), `CheckBriefingIntent`, `FlightsOverviewIntent`, `RefreshBriefingIntent`, `AirportWeatherIntent`.
- **App Shortcuts:** `FlyFunShortcuts: AppShortcutsProvider` with English phrases (each contains "FlyFun").
- **Full tiered resolver:** deterministic (`AirportDatabase`/date keywords) → **Foundation Models `{place, when}` fallback** (Foundation Models ships in iOS 26; gated on `SystemLanguageModel.default.availability`) → Siri disambiguation.
- **Navigation:** typed `PendingNavigation` seam on `AppState` (reuse the `onOpenURL` path).
- **App Group provisioning** (decided): move Keychain access-group + `BriefingCacheStore` into the shared container now, so later Widgets/Live Activities need no migration.
- **Expired-token fallback:** silent refresh via `RollingBearerSession`, else "Open FlyFun to sign in".
- **Spotlight:** donate `FlightEntity` on list-load/create, remove on delete.

### Acceptance
- "Hey Siri, show me my FlyFun briefings" → opens the flight list.
- "…open the FlyFun briefing for my next flight" → opens that briefing (works offline from cache).
- "…refresh my FlyFun briefing for the flight tomorrow to Fairoaks" → triggers a refresh and speaks the freshness-gated outcome.
- Intents appear in Shortcuts; flights are Spotlight-searchable.
- Core paths work on every device; the LLM resolver tier lights up only on Apple-Intelligence-capable devices.

### Notes
- No SiriKit to migrate (clean start on App Intents).
[...]

PR #367, opened 2026-07-08
feat(ios): App Intents — Siri / Shortcuts / Spotlight (Tier 1, #364)

roznet:
Phase 1 of designs/ios-app-intents.md: the on-device sibling of the MCP
tool catalog, so pilots can drive FlyFun from Siri, Shortcuts, Spotlight,
and the Action Button on current iOS 26 — no Apple-Intelligence dependency.

Entities (AppIntents/):
- FlightEntity (AppEntity + IndexedEntity) backed by the cache-first
  CachingBriefingRepository; AirportEntity backed by AirportDatabase.
- FlightEntityQuery / AirportEntityQuery (EntityStringQuery) resolve from
  the user's flights / local ICAO search.

Tiered resolver (FlightResolver + FlightQueryModel):
1. deterministic place↔ICAO (AirportDatabase) + relative-date keywords —
   the mandatory floor, offline, every device;
2. Foundation Models {place, when} guided generation, gated on
   SystemLanguageModel.default.availability;
3. Siri disambiguation. The LLM only flattens language; the deterministic
   matcher against real flights stays the authority.

Intents + Shortcuts:
- OpenFlightListIntent, OpenBriefingIntent (next-flight default),
  CheckBriefingIntent, FlightsOverviewIntent, RefreshBriefingIntent,
  AirportWeatherIntent (@AppEnum day), plus FlyFunShortcuts phrases.
- RefreshBriefingIntent speaks the freshness-gated outcome (queued /
  already-fresh / in-progress / rate-limited); interim "started" UX until
  the briefing-ready push lands (RefreshDriver keeps the SSE alive).
- Signed-out/expired-token: foreground intents open the app (auth gate);
  background intents speak the sign-in line (Decision 4).

Navigation + Spotlight:
- Typed PendingNavigation seam on AppState, consumed on scene .active
  (cold-launch-safe via PendingNavigationStore), reusing the onOpenURL path.
- SpotlightDonator reconciles the flight index on every list load.

Supporting:
- airportWeather(icao:day:hour:) added to BriefingRepository + all
  conformers, with a minimal AirportWeatherResponse DTO
  (GET /api/maps/airport-weather).
- Pure-logic unit tests for the deterministic resolver.

App Group (Decision 1) intentionally deferred: Phase-1 intents run
in-process, so no shared container is needed; provisioning the entitlement
would require Apple Developer-portal registration and break code-signing
for zero Tier-1 benefit. Documented in the design doc for the first
out-of-process consumer (Widgets).

Note: iOS-only Swift; not build-verified in this environment — needs an
Xcode build + on-device Siri/Shortcuts pass.

Claude-Session: https://claude.ai/code/session_017UpYxMMBNcmwwcDPny5PbG


roznet:
Thanks for the review — addressed all four findings in `5fe4928`.

**Blockers (fixed):**
1. **Airport DB load race** — added `AirportDatabase.ensureLoaded()` which awaits the in-flight `loadCached()` task; made `IntentSupport.ensureAirportDatabase()` `async` and now `await` it before every `search()` / `airport(icao:)` in `FlightResolver.resolve` and all three `AirportEntityQuery` paths. A freshly-spawned Siri process now loads the DB before the deterministic place resolver runs.
2. **Refresh false-success** — a dropped/empty SSE stream (no progress *and* no terminal event) now maps to `.failed` instead of `.started`, matching `BriefingViewModel.refresh()`'s anomaly handling. A genuine mid-stream `progress` event still correctly reports `.started`.

**Minor (batched in since I was pushing anyway):**
3. **Design-doc contradiction** — reconciled the "Auth & process model" section and Decision 1 with the top-of-file App-Group deferral note (both previously still said "provisioned in Phase 1").
4. **Spotlight reindex race** — `SpotlightDonator.reindex` now serializes calls (each chains after the previous), so an overlapping delete-all can't interleave with a still-running insert.

Nothing deferred. Note (unchanged from the PR body): this is iOS-only Swift not build-verified in the CI environment — still needs an Xcode build + on-device Siri/Shortcuts pass.

---

roznet:
Round 2 addressed in `ce97301` — all five findings.

**Blockers (fixed):**
1. **Signed-out UX was dead code for Check/Refresh** — since parameter resolution runs before `perform()`, `FlightEntityQuery.loadFlights()`'s throw pre-empted the Decision-4 guard. Dropped the throw: resolution now degrades to an empty match on any load failure, and a signed-out user *with a cached list* still resolves flights so `perform()` runs and speaks the sign-in line. Removed the now-unused `IntentAuthError`.
2. **Expired-token 401 uncaught in Check/Overview** — wrapped both `repo.flights()` calls in the same `catch APIError.unauthorized` → sign-in line pattern used by `AirportWeatherIntent`.
3. **`airportWeather` mislabeled snapped forecasts** — the main sentence now keys off `entry.icao` (the airport the data is actually for) instead of the originally-requested ICAO, so a snapped result no longer attributes the nearest airport's numbers back to the unmonitored one.

**Minor (batched in):**
4. **Deep-link target dropped on stale cache** — `applyPendingNavigation` now keeps the pending target while `viewModel.isOffline`, so a later successful fetch (e.g. a just-created flight) can still resolve it; it only drops when an authoritative online list confirms the flight is gone.
5. **Duplicated sign-in literal** — `RefreshDriver.classify` now references `IntentSupport.signedOutSpokenLine` (made `nonisolated`).

Nothing deferred. Same standing caveat: iOS-only Swift, not build-verified here — needs an Xcode build + on-device Siri/Shortcuts pass.

---

roznet:
Addressed in `a2e97c7`.

**Blocker (fixed):**
- **Tier 1 now scoped to upcoming flights** (finding 1). Free-phrase resolution matches over `upcoming(flights, now:)` in *both* tiers, so a long-past flown trip can't resolve a voice request ("Check my assessment for Cannes" no longer surfaces a months-old Cannes flight). Explicit id lookups (`entities(for:)`) and the Shortcuts picker (`suggestedEntities`) still see history — only free-phrase resolution is scoped. Regression test added (`ignoresPastPlaceMatch`).

**Minor (batched in):**
- **Spotlight reindex gated on `!isOffline`** (finding 3) — an offline cache snapshot can be stale/partial, so delete-all-then-add is deferred to the next online load rather than transiently dropping entries.
- **Pure-logic test coverage** (finding 3b) — `RefreshDriver.classify` (APIError → Outcome) and `IntentDialogs.checkSummary`/`overviewSummary`/`airportWeather`. The airport-weather test also locks in the round-2 snapped-airport attribution fix.

**Deferred (with rationale):**
- **`RefreshDriver.run` cancellation-awareness** (finding 2). Valid, but the correct fix — a cancellable continuation coordinating with the keep-draining task (which must outlive the early `.started` return so the SSE isn't torn down) — is non-trivial concurrency I can't compile-verify in this environment, and the current behavior is already bounded by `APIClient.streamSSE`'s 300s idle timeout. Flagging as a good on-device follow-up rather than shipping it blind.

---

PR #369, opened 2026-07-10
Fix iOS flight-create submitting raw Field-15 route tokens

roznet:
## Problem

Creating a flight on iOS from an ICAO Field-15 route (e.g. one containing a speed/level+rules token like `N0180VFR`) could fail with **"Invalid waypoint 'N0180VFR': must be 2-5 alphanumeric characters"**.

Both clients already share the same server-side parser (`POST /api/flights/interpret-route` → `parse_field15` + `resolve_waypoints`) — there is **no** on-device Swift parser. The divergence was purely in the iOS **submit workflow**.

The server has a deliberate two-stage design:
1. **Smart, context-aware interpretation** (`/interpret-route`) — `parse_field15` drops flight-plan syntax (speed/level, `DCT`, airway labels, SIDs) by grammar; `resolve_waypoints` places the survivors using full route context (departure/destination anchoring + detour filtering, needs ≥2 points).
2. **Dumb per-token guardrail** (`WAYPOINT_RE = ^[A-Z0-9]{2,5}$` in `CreateFlightRequest.validate_waypoints`) — a context-free backstop meant only for already-cleaned codes.

The **web** always awaits stage 1 at submit (`interpretAndConfirmRoute`) and aborts on failure, so raw syntax never reaches stage 2. **iOS** only ran stage 1 best-effort via the debounced `resolveRoute()`. If the pilot pasted a route and tapped **Create** before the 500 ms debounce + network round-trip finished (or if `resolveRoute()` failed silently), `routeInterpretation` was `nil`, `routeHasDroppedTokens` was `false`, and the form submitted the **raw** `waypoints` straight to the strict validator.

So it wasn't the wrong function or a bad triage — it was a sequencing bug: the context-aware step wasn't guaranteed to run before submit.

## Fix

Mirror the web: always run interpretation at submit time.

- **`AddFlightViewModel.interpretRouteForSubmit()`** — interprets on the server at submit and returns `ready` / `needsConfirmation` / `failed`. Clean routes auto-apply the interpreted waypoints; dropped-token routes defer to the confirm sheet; failures set `errorMessage` and abort — raw tokens are never submitted.
- **`AddFlightView.submit()`** — restructured to interpret before both create and edit (edit only when the route text changed), routing confirmation through the existing `RouteInterpretSheet` via a shared create/edit continuation.
- **`AddFlightViewModel.routeChangedFromOriginal`** — gates the edit-path round-trip so an untouched route needs no interpret call.

## Tests

- `MockBriefingRepository` now stubs `interpretRoute` and captures the create request.
- `RouteSubmitInterpretationTests` covers: clean route strips Field-15 syntax and is `ready`; the created flight carries interpreted (not raw) waypoints; dropped tokens defer to confirmation and keep the raw field; interpret failure and `<2` resolved waypoints both abort with an error; edit-path route-change gating.

> Note: this is an iOS target, so the new Swift tests could not be compiled/run in the CI-less Linux environment they were written in — they follow the existing patterns in the file but should be built in Xcode to confirm green.

---
_Generated by [Claude Code](https://claude.ai/code/session_01UJRLqEEkRpDzQs9ZMkgeCX)_

### DIFF
--- a/app/flyfun-weather/flyfun-weather/ViewModels/AddFlightViewModel.swift
+++ b/app/flyfun-weather/flyfun-weather/ViewModels/AddFlightViewModel.swift
@@ -158,6 +158,15 @@ final class AddFlightViewModel {
         return abs(departureDate.timeIntervalSince(originalDate)) > 1
     }
 
+    /// Whether the typed route differs from the flight being edited (case- and
+    /// spacing-insensitive). Gates submit-time interpretation on edit — an
+    /// untouched route is already clean, so it needs no round-trip. On create
+    /// there is no baseline, so this is always `true`.
+    var routeChangedFromOriginal: Bool {
+        guard let original = editingFlight else { return true }
+        return waypoints != original.waypoints.map { $0.uppercased() }
+    }
+
     /// Whether the edited alt-departure instant differs from the flight's stored
     /// one (or newly sets one). Used only in `.alternate` mode.
     private func altDepartureChanged(from original: FlightResponse) -> Bool {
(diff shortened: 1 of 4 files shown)


roznet:
Thanks — addressed all three findings in `5bc2195`.

**Critical (double-tap → duplicate flight)** — Added `isPreparingSubmit`, set synchronously at the top of `submit()` *before* the first suspension point and held across the entire flow (interpret → optional confirm sheet → create/save), released at every terminal branch (`.ready` after create, `.failed`, and the confirm sheet's `onDismiss` for both cancel and accept). The button is also disabled on it. It's kept out of `canSubmit` deliberately so `createFlight()`/`saveEditedFlight()`'s own `guard canSubmit` still passes on the `.ready` path. The second tap is now rejected atomically on the main actor before it can spawn a second `Task`.

**Important (stale `errorMessage` on `.needsConfirmation` / rebrief paths)** — `interpretRouteForSubmit()` now clears `errorMessage` at the top, so a successful retry that routes to the confirm sheet drops the previous failure banner instead of leaving it behind the sheet.

**Minor (triplicated route comparison)** — Factored into a single `private func routeDiffers(from:)` used by `hasChanges`, `hasForecastAffectingChange`, and `routeChangedFromOriginal`.

Also added `interpretClearsStaleErrorOnNeedsConfirmation` to the test suite. Note: these are iOS-target Swift changes and there's no Swift toolchain in this environment, so the new tests haven't been compiled/run here — they'll want an Xcode build to confirm green.

---

PR #370, opened 2026-07-10
iOS: seamless sync to latest online pack + live refreshing indicator

roznet:
## Problem

An already-open iOS app never re-fetched while foregrounded, so a briefing refreshed on the website (or by auto-refresh) left the **flight-list summary** and the **open briefing** stale until a cold relaunch. The only briefing-level update affordance was the ↻ button — which triggers a *new full generation* (SSE stream), not a sync to what's already online. There was also **no pull-to-refresh on the briefing at all**.

## Approach — sync vs refresh

Two distinct operations, matching the web:

| | **Sync** (new) | **Refresh** (unchanged, ↻ button) |
|---|---|---|
| What | Adopt the newest *already-generated* pack | Generate a *new* briefing |
| Cost | 1 `latestPack()` request; no-op if unchanged | Full pipeline, minutes, SSE |
| Triggers | appear, pull-to-refresh, foreground, push | explicit button |
| UX | seamless (quiet swap, no spinner flash) | progress banner |

## What's in here

**1. Seamless sync (`4c255989`)**
- `BriefingViewModel.syncLatestPack()` — no-ops when the pack timestamp is unchanged, else adopts via a new shared `applyPack()` (consolidates 3 duplicated adopt-pack blocks). Reload is *quiet* — sections keep current data and swap in place (`LoadingState.hasData`).
- Wired to: briefing pull-to-refresh (was absent), return-to-foreground (`scenePhase`), and an `AppState.externalSync` nudge the APNs handlers now raise (push refreshes visible data, not just the badge). List already synced on appear/foreground/pull; now also on the push nudge.

**2. Live "Refreshing…" flight-list indicator (`5db0876c`)**
- iOS has no push for progress, so — like the web — the list polls `GET /api/refresh/active` (existing endpoint) every 5s while visible and marks each returned flight's row. One user-scoped request per tick regardless of flight count.
- When a flight leaves the active set (refresh finished), the loop re-syncs the list — the same loop doubles as the completion trigger.
- New `ActiveRefreshResponse` + `activeRefreshes()` on the repository protocol (online-only; caching passes through; fixture/mock return `[]`).

**3. Polish (`cdf22cd9`)**
- **Long-range hero fix:** the Advisory hero rendered only `pack.assessment`, so a far-out briefing (which carries `outlook` instead) showed a blank summary. The server already sends `outlook`/`outlook_reason` and the flight card already falls back to it — added the fields to iOS `PackMetaResponse` and the same assessment-else-outlook fallback to the hero (shared `heroCard`; reusable `OutlookBadge`).
- Row label "Refreshing…" (matches web).
- `quietLog` flag suppresses the 5s poll's request log.
- Offline-safe poll: skips the round-trip when `NetworkMonitor` is offline (clears stale indicators) + exponential backoff on failure (5s→cap 60s, reset on success).

The ↻ button is unchanged (explicit "generate a new briefing").

## Testing

- `xcodebuild build` (generic iOS Simulator) + full `flyfun-weatherTests` target green.
- ⚠️ Not yet verified on device — the seamless swap, pull-to-refresh gesture, and long-range hero need a real authenticated session (mock mode can't render the briefing).

### DIFF
--- a/app/flyfun-weather/flyfun-weather/App/AppState.swift
+++ b/app/flyfun-weather/flyfun-weather/App/AppState.swift
@@ -72,6 +72,28 @@ final class AppState {
     /// the bundled baseline at init; refreshed opportunistically when online.
     let helpCatalog = HelpCatalogStore()
 
+    /// External "server data changed" nudge (e.g. a refresh push). The flight list
+    /// and any open briefing observe this and re-sync to the newest online pack —
+    /// push is just one more sync trigger, not a special path. `token` makes every
+    /// signal distinct (so back-to-back nudges each fire); `flightId` carries the
+    /// affected flight when known. Bump via `signalExternalSync`.
+    private(set) var externalSync = ExternalSyncSignal(token: 0, flightId: nil)
+
+    /// A distinct, observable data-sync nudge. `Equatable` so SwiftUI `onChange`
+    /// fires once per signal.
+    struct ExternalSyncSignal: Equatable {
+        var token: Int
+        var flightId: String?
+    }
+
+    /// Request a seamless re-sync of the visible flight list / open briefing to
+    /// the latest online pack. Cheap by design — the observers no-op when nothing
+    /// changed. `flightId` is the affected flight when the trigger names one
+    /// (briefing push), else nil (e.g. a silent badge-sync).
+    func signalExternalSync(flightId: String?) {
+        externalSync = ExternalSyncSignal(token: externalSync.token &+ 1, flightId: flightId)
+    }
+
     /// Weak handle to the live instance so a foregrounding App Intent can trigger
     /// `consumePendingNavigation()` immediately when the app process is already
     /// alive (warm foreground / already-active). Nil on a cold launch — the scene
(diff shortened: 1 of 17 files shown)


roznet:
Thanks — addressed both **Important** findings in `24f5fdcc`, and folded in both Minors since I was pushing anyway.

**Important 1 — history-pick clobbering:** `syncLatestPack()` now gates on a new pure/static `isViewingLatestPack(current:history:)`. If the pack on screen isn't the newest one known (i.e. the pilot deliberately picked an older pack from the history menu), a foreground/push-triggered sync is skipped. A brand-new server pack still adopts, since it isn't in `packHistory` yet so a viewer on the current latest passes the guard.

**Important 2 — cross-pack section contamination:** took the per-loader timestamp-check option. All five section loaders now re-check `pack?.fetchTimestamp == timestamp` after their `await` before applying the result (same pattern as `pollTimeOptions`). A late load from a superseded sync/refresh/history-pick is dropped instead of painted under another pack's hero/toolbar, so sections always match the active `pack` regardless of interleaving.

**Minor — tests:** added `autoSyncOnlyWhenViewingLatestPack` pinning the history-pick guard (incl. fractional-second ordering). A full overlapping-`applyPack` integration test needs configurable mock-repo infra (the current `MockBriefingRepository` stubs throw) — deferred as follow-up; the stale-guard above is the concrete safety net.

**Minor — poll retain:** added a comment documenting the intentional strong-`self` capture in `pollActiveRefreshesLoop` (bounded by `stopActiveRefreshPolling()` on disappear/background).

Build + full `flyfun-weatherTests` target green.

Issue #371, opened 2026-07-10
Briefing notification preferences UI — scope / channels / per-flight (web + iOS)


roznet:
## Summary

The server half of briefing-refresh notifications shipped in #368 (APNs push, email, badge, prefs model, migration 075). But the **preferences have almost no UI**: iOS exposes only a single "Briefing Push Alerts" toggle (`notify_push`), and web exposes **none** of the `notify_*` prefs. This issue builds the full settings surface across web + iOS so users can actually control *whether*, *when*, and *how* they're notified — plus the small backend changes needed to make the semantics correct.

Design doc: `designs/ios-app-briefing-notifications.md`. Server prefs already live in `app_prefs_json` (`notify_scope`, `notify_change_only`, `notify_email`, `notify_push`) and per-flight `FlightRow.notify_override`.

## Mental model

Two orthogonal axes, at two scopes:

- **Whether / when** — `notify_scope` (global) · `notify_change_only` (global) · `notify_override` (per-flight exception)
- **How** — `notify_email` · `notify_push` (global channels)

**Core invariant: channels never express "off."** The *only* way to silence notifications is via scope/override — you cannot disable all *available* channels. This makes the silent dead-state ("scope says notify, but no channel on → nothing happens") structurally impossible instead of a warning.

---

### A. Account › Notifications (new section — web has none today; iOS extends the existing push toggle)

| Control | Type | States | Semantics | Backend |
|---|---|---|---|---|
| **Briefing updates** | 3-stop picker | **Off** / **Assessment changes** *(default)* / **Every update** | *Off* → never (default-resolution, see below). *Assessment changes* → notify only when an update moves the flight's **assessment** (the GREEN/AMBER/RED headline) or early **outlook**. *Every update* → notify on every refresh, changed or not. | folds `notify_scope` + `notify_change_only` |
| **Email** | toggle | on/off | Deliver by email. Default **on**. | `notify_email` |
| **Push** | toggle | on/off | Deliver as APNs push. Default **off**. **iOS:** toggling on requests permission + registers this device. **Web:** shown only if ≥1 device registered; otherwise a disabled row *"Install the FlyFun iOS app to enable push."* | `notify_push` |

[...]

PR #372, opened 2026-07-10
Implement briefing-notification preferences UI and per-channel trigger rules

roznet:
Completes the briefing-notification feature (issue #371) with full preferences control on web and iOS, and refines the server-side notification gate to separate the shared qualification logic from per-channel delivery rules.

## Summary

This PR implements the complete notification preferences system across all three clients (server, web, iOS) with a new per-channel trigger rule that suppresses email for the user's own in-app manual refresh while allowing it for non-user-present triggers (scheduler, Siri, MCP, background). The shared gate (`notify_qualifies`) is simplified to be channel- and trigger-agnostic, driving only the badge and base qualification decision.

## Key Changes

**Server (`src/weatherbrief/notify/dispatch.py`)**
- Refactored `notify_qualifies()` to remove `triggered_by` parameter — now purely scope + override + change-only, channel-agnostic
- Added `email_should_send(triggered_by)` to implement per-channel trigger rule: email fires only for non-user-present refreshes (scheduler/Siri/MCP/background), never for the user's own in-app manual refresh (`"user"`)
- Updated docstrings to clarify the two-layer decision model: shared gate → badge + base decision, then per-channel rule layers on top
- Added `_USER_PRESENT_TRIGGERS` constant to define which refresh sources suppress email

**API & Preferences (`src/weatherbrief/api/preferences.py`, `src/weatherbrief/api/packs.py`)**
- Extended `PreferencesResponse` with `notify_decay_notice` (one-time fail-safe when last device removed) and `push_device_count` (for conditional UI)
- Added `_count_push_devices()` helper to count registered APNs devices
- Implemented `apply_last_device_decay()` to re-enable email when the last push device is unregistered while email is off (channel invariant fail-safe)
- Added `_normalize_source()` to clamp client-declared refresh sources to allowed set (`user`, `siri`, `mcp`); unknown values default to `"user"`
- Updated `_notify_refresh_complete()` to use granular `triggered_by` from `result.usage` instead of coarse `force_email` flag

**Devices (`src/weatherbrief/api/devices.py`)**
- Added decay logic to `unregister_device()`: when the last device is removed and email is off, re-enable email and set the decay notice flag

**Web UI (`web/ts/settings-main.ts`, `web/settings.html`, `web/ts/adapters/preferences-adapter.ts`)**
- Implemented `renderNotificationSettings()` to populate Account › Notifications section with:
  - **Briefing updates** 3-stop picker (Off / Assessment changes / Every update) folding `notify_scope` + `notify_change_only`
  - **Email** toggle (locked on when it's the sole available channel while updates are active)
  - **Push** toggle (disabled if no registered device)
- Added channel invariant enforcement: turning off the last enabled channel reroutes to Briefing updates = Off
- Implemented `foldBriefingUpdates()` / `unfoldBriefingUpdates()` to convert between the 3-stop and scope + change-only
- Added decay notice display and dismissal
- Removed legacy per-refresh "Email me when done" checkbox from freshness bar

**iOS UI (`app/flyfun-weather/flyfun-weather/Views/SettingsView.swift`, `app/flyfun-weather/flyfun-weather/Models/API/PreferencesResponse.swift`)**
- Implemented notification preferences section with same 3-stop + channel toggles as web
- Added channel invariant enforcement and decay notice handling
- Extended `PreferencesResponse` with `briefingUpdates` computed property and device count
- Added `BriefingUpdates` enum and preference update methods (`updateBriefingUpdates()`, `updateNotifyEmail()`, `dismissDecayNotice()`)

**Per-Flight Override (Bell) (`app/flyfun-weather/flyfun-weather/

https://claude.ai/code/session_018rtqBv4vVdXgx7bpBfoDEU


roznet:
Thanks for the review — all four findings addressed in `058b252`:

- **Critical (iOS `setPush` dead-state)** — confirmed reachable. `setPush` now reroutes *Briefing updates → Off* when turning push off would strand the user (email already off), mirroring `setEmail` and the web `pushToggle` reroute. No more scope-on-with-no-channel via the UI.
- **Important (server-side invariant)** — added a backstop in `update_preferences`: if a merged update leaves `notify_scope != off` with both channels false, scope is clamped to `off` (mirrors the client reroute), so the invariant now holds for *any* writer — a raw `PUT` or a future client — not just the two hand-written forms. `apply_last_device_decay` still covers the last-device path (re-enable email); the two fail-safes are intentionally different (involuntary device loss → keep notifications, restore email; voluntary both-off → interpret as "silence" → scope off). Added tests for the clamp and the valid push-only case.
- **Minor (iOS↔web parity)** — unified `emailLocked` to the web model: email is locked-on only when there's *no registered device at all*; with a device present both toggles stay editable and the last-one-off reroutes to Off. iOS and web now present the same interaction model.
- **Minor (i18n)** — translated the ~20 new `settings.notify.*` / `notify.override.*` / `notify.resolved.*` strings into de/es/fr (were en-only).

Verification: web `tsc` clean + 438 vitest pass; Python compiles (the private-dep test suite can't run in my sandbox, so the new gate/clamp/decay tests are written + reviewed but need CI to execute).

---

roznet:
Round 2 addressed in `0684aa9`:

- **Important #1 (decay vs explicit Off)** — `apply_last_device_decay` now early-returns when `notify_scope == "off"`, so losing the last device no longer resurrects email against a user who deliberately silenced everything. Reworked the decay tests to build a real push-only state (`scope=all`, email off — note the round-1 clamp turns a bare `{email:false}` into `scope=off`, which is exactly why the old test needed updating) and added a test that decay respects an explicit Off.
- **Important #2 (web dropdown dead-end)** — the "Briefing updates" select now enables email when leaving Off with no channel on, instead of snapping back to Off. Mirrors iOS `setBriefingUpdates`; the channel toggles keep their reroute-to-Off behaviour. The two platforms now behave the same in both directions.
- **Important #3 (unlocalized labels)** — you're right, the `data-i18n` attributes were dead markup. Dropped them and set the section title/hint/decay-notice/updates label/options/Email/Push labels via `t()` in `renderNotificationSettings`, and added the 9 missing keys to en/fr/de/es.
- **Minor #4 (`notify_decay_notice` contract)** — `update_preferences` now only honors `notify_decay_notice == false`; a client can't raise it (server owns the decay path).

**Deferred — Minor #5** (decay-notice dismissal: web on Save vs iOS on disappear): both behaviours are reasonable and it's a judgment call, so I left it rather than churn the UX. Happy to align them (e.g. web auto-dismiss on view) if you'd prefer.

Verified locally: web `tsc` clean + 438 vitest pass; Python compiles (the gate/clamp/decay tests run in CI — that's still the first place the private-dep suite actually executes).

---

roznet:
Good catch — fixed in `21d1a81`.

**Important #1 (decay skipped on APNs prune)** — confirmed: `_dispatch` pruned dead tokens but never ran the fail-safe, so a push-only user (email off) whose last token died via a routine send — the app-deleted-without-sign-out path — was silently stranded. `_dispatch` now takes `user_id` and, after pruning, fires `apply_last_device_decay` when that dropped the user's last device, mirroring `unregister_device`. Both `send_briefing_push` and `send_silent_badge_push` pass it. Added a test (dead-token prune on a push-only user's last device → email re-enabled + decay notice).

**Minor #2 (duplicated count query)** — extracted `notify/push.count_user_devices`, now used by the conditional-push UI (`preferences._count_push_devices`), the sign-out decay path (`devices.unregister_device`), and the new prune path, so the three can't drift.

Verified: Python compiles; the decay/prune tests run in CI (private-dep suite doesn't install in my sandbox).

---

PR #378, opened 2026-07-11
Port NWP fallback logic for unavailable cloud/icing/convective layers

roznet:
## Summary

Implements NWP→fallback substitution for models without native NWP data (e.g., far-out ECMWF flights). When a model can't provide a requested method, the system now marks it unavailable in the UI and substitutes the fallback method at render time, ensuring the cross-section never silently renders blank layers. This mirrors the web client and backend behavior, achieving iOS↔web parity.

## Key Changes

**iOS App (`flyfun-weather`)**
- Added `NwpFallback` enum with three core functions:
  - `unavailableLayers(in:)` — detects which layers lack data for the current model (NWP clouds when `nwpCloudLayers` is nil, SFIP when no zones present, NWP convective when `hasNwpConvective` is false)
  - `applyFallback(enabledLayers:unavailable:)` — builds effective render-time layer map by disabling unavailable layers and substituting fallbacks (NWP clouds → same-style DD clouds, NWP/SFIP icing → Ogimet-DD, NWP convective → thermodynamic)
  - `ddSubstituteId(for:)` — maps each unavailable NWP layer to its fallback
- Updated `CrossSectionViewModel` to expose `unavailableLayers`, `effectiveEnabledLayers`, and `substitutedLayers` as computed properties
- Modified `CrossSectionView` to use `effectiveEnabledLayers` instead of stored `enabledLayers` for rendering
- Enhanced `CrossSectionConfigSheet` to:
  - Grey out unavailable methods in the dropdown menu
  - Display "— no data" suffix for unavailable options
  - Show info labels explaining why NWP methods are substituted (e.g., "No native NWP cloud data for this model at this range — showing DD clouds instead")
- Added comprehensive unit tests (`NwpFallbackTests`) covering all three functions with 8 test cases

**Backend (`weatherbrief`)**
- Clarified `SoundingAnalysis.nwp_cloud_layers` semantics: `None` means no native NWP envelope at this lead time (no GRIB enrichment), distinct from `[]` (model clear sky)
- Updated `_format_sounding_context` in `prompt_builder.py` to emit native NWP cloud decks (same signal the app cross-section draws) instead of the bulk Open-Meteo low/mid/high summary, which mislabels ECMWF
- Added three test cases validating the new behavior: absent native layer, clear sky, and native decks with coverage/base/top/CC/T

## Implementation Details

- The stored `enabledLayers` preference is never mutated — switching back to an NWP-capable model auto-restores NWP with no persisted "downgraded" flag, matching web behavior
- Substitution is purely render-time: the effective map is computed fresh each render from the stored preference + current model's unavailable set
- Cloud layer substitution handles all six variants (source × style): soft/natural/square NWP clouds each map to their DD counterpart
- Single-layer fallbacks (icing, convective) only activate if the user explicitly enabled the NWP method and hasn't already enabled the fallback
- The digest now explicitly states "no native NWP cloud layer at this lead time" when `nwpCloudLayers` is None, matching the app's empty NWP layer rather than contradicting it with a bulk summary

https://claude.ai/code/session_01Ne6PLaiwnH4sDz1pkPKhdg

### DIFF
--- a/app/flyfun-weather/flyfun-weather/ViewModels/CrossSectionViewModel.swift
+++ b/app/flyfun-weather/flyfun-weather/ViewModels/CrossSectionViewModel.swift
@@ -81,6 +81,32 @@ final class CrossSectionViewModel {
         dataVersion += 1
     }
 
+    // MARK: - NWP availability & fallback (port of web getUnavailableLayers +
+    // applyNwpFallback). See `NwpFallback`. All three are pure functions of the
+    // already-observed `vizData` + `enabledLayers`, so reading them inside the
+    // Canvas / config sheet registers the right `@Observable` dependencies.
+
+    /// Layer ids the currently-rendered model can't provide (no native NWP data,
+    /// etc.) — greyed / disabled in the config sheet. Empty until `vizData` loads.
+    var unavailableLayers: Set<String> {
+        guard let vizData else { return [] }
+        return NwpFallback.unavailableLayers(in: vizData)
+    }
+
+    /// Render-time enabled map: the stored preference with unavailable layers
+    /// disabled and DD substituted for any wanted-but-unavailable NWP layer. The
+    /// stored `enabledLayers` preference is never mutated (switching back to an
+    /// NWP-capable model auto-restores NWP). Mirrors web `briefing-main.ts`.
+    var effectiveEnabledLayers: [String: Bool] {
+        NwpFallback.applyFallback(enabledLayers: enabledLayers, unavailable: unavailableLayers)
+    }
+
+    /// Layers auto-substituted at render (DD standing in for an unavailable NWP
+    /// method) — the config sheet flags these so the user knows why NWP looks off.
+    var substitutedLayers: Set<String> {
+        NwpFallback.substitutedLayers(enabledLayers: enabledLayers, effective: effectiveEnabledLayers)
+    }
+
     func toggleLayer(_ id: String) {
         enabledLayers[id] = !(enabledLayers[id] ?? false)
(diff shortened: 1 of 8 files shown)


roznet:
Thanks for the review — both findings addressed (branch rebased onto current `main` first, then fixes pushed).

**Important — digest regression for always-`None` models (fixed).** You're right: `nwp_cloud_layers is None` conflates two cases, and the previous "no native NWP cloud layer at this lead time" line dropped real bulk cloud data and misframed it as temporal for UKMO/MétéoFrance/GEM (Open-Meteo-only, permanently `None`). The `None` branch now falls back to the bulk Open-Meteo low/mid/high summary under a disambiguated label — `Bulk cloud [model] (Open-Meteo summary, no native NWP layer): Low=…, Mid=…, High=…` — so no data is lost and it's never mislabeled as native "NWP cloud". The bare `no native NWP cloud layer` line (no "at this lead time" framing) is used only when the bulk summary is also absent. Added `test_build_context_nwp_cloud_absent_falls_back_to_bulk` (asserts the bulk label, no `NWP cloud: Low=` mislabel, no `at this lead time`) and `test_build_context_nwp_cloud_absent_no_bulk`.

**Minor — unused `substitutedLayers` (removed).** Dropped both `CrossSectionViewModel.substitutedLayers` and `NwpFallback.substitutedLayers` — they were computed/tested but never read by any view (the config sheet already conveys substitution via inline "showing DD instead" hints). Adjusted `NwpFallbackTests` to assert on the effective map directly instead. If a method group ever becomes multi-select, the web-parity helper can be re-added and wired to dim the substituted row.

Nothing deferred.

---

PR #380, opened 2026-07-11
Centralize heading/bearing formatting logic

roznet:
## Summary
Extracts repeated heading/bearing formatting logic into a single reusable function to ensure consistent zero-padded display across the codebase (web and iOS).

## Changes
- **New `formatHeading()` function** in `web/ts/units.ts`: Zero-pads compass headings to 3 digits with optional rounding step (10° for wind chips/METAR/TAF, 1° elsewhere). Does not append the ° symbol — call sites add it themselves.
- **New `paddedHeading` property** in Swift (`Extensions.swift`): Equivalent zero-padding for iOS, serving as the single source of truth for heading display in `AirportConditionsView` and `PirepListView`.
- **Replaced inline formatting** across web modules:
  - `advisories-ui.ts`: Runway heading and wind direction formatting
  - `briefing-ui.ts`: Wind direction in tooltips and metric value display
  - `skewt/interaction.ts`: Wind direction in sounding profile tooltips
  - `cross-section/tooltip-formatters.ts`: Sun azimuth display
  - `weather-map.ts`: Wind direction in map legend
  - `pirep-ui.ts`: PIREP wind direction display
- **Updated iOS views**:
  - `AirportConditionsView.swift`: Wind display using `paddedHeading`
  - `PirepListView.swift`: PIREP wind display using `paddedHeading`

## Implementation Details
- `formatHeading()` rounds to the nearest `step` degrees first, then zero-pads to 3 digits (e.g., 10° → "010", 350° → "350")
- Eliminates scattered `Math.round(deg / 10) * 10` and `padStart(3, '0')` patterns
- Maintains backward compatibility — all call sites already append ° or @ symbols as needed

https://claude.ai/code/session_0183gnUXnfYFXTgtwZfkWD2b

### DIFF
--- a/app/flyfun-weather/flyfun-weather/Utilities/Extensions.swift
+++ b/app/flyfun-weather/flyfun-weather/Utilities/Extensions.swift
@@ -68,6 +68,15 @@ extension String {
     }
 }
 
+extension Int {
+    /// Zero-pad a compass heading/bearing to 3 digits (10° → "010"). Does NOT
+    /// append the ° symbol — call sites add °/@ as they do today. Single source
+    /// of truth for heading display (AirportConditionsView, PirepListView).
+    var paddedHeading: String {
+        String(format: "%03d", self)
+    }
+}
+
 extension URL {
     /// Extract a query parameter value by name.
     func queryParam(_ name: String) -> String? {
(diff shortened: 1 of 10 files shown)


roznet:
Addressed the review finding in `065729e`.

**Confirmed the finding was correct.** The model-comparison table is `renderComparisonTable`, whose value cells format every variable with `val.toFixed(1)` — so a 10° heading rendered as `10.0`, not `010`. The `wind_direction_deg` branch I'd added to `formatMetricValue` was dead code: that function is only called from the sounding convective/indices table (line 2998), whose metrics never include `wind_direction_deg`.

**Fix:** route the `wind_direction_deg` cell in `renderComparisonTable` through `formatHeading` (all other metrics keep their 1-decimal display — no behavior change elsewhere), and removed the unreachable branch from `formatMetricValue`.

Nothing deferred. `tsc --noEmit` is clean on the touched file (the vitest suite couldn't run in this environment — deps not installed).

---

Issue #392, opened 2026-07-12
Flight-level assessment has no 'unavailable' state — returns GREEN when we have no data


roznet:
Split out of #391, because this one is **not a bug fix — it's a missing state**, and it needs a product decision before any code.

## The problem

`derive_assessment_from_advisories` (`tasks/advise.py:643`) filters out UNAVAILABLE advisories, and then:

```python
if not statuses:
    return ("GREEN", "No advisory data available")
```

So the **flight-level traffic light** — the one a pilot sees on the flight list and at the top of the briefing — is **GREEN when we have nothing to say**. The reason string is honest; the colour is not.

Note #391 does *not* fix this. It makes individual advisories UNAVAILABLE, which this function then filters out, landing on the empty list and still returning GREEN. If anything #391 makes this path *more* reachable.

## Why it's its own issue

The assessment is not a local value. It feeds:

- `pipeline.py`
- `connectors/views.py` — MCP **and** the ChatGPT connector
- `notify/dispatch.py`, `notify/email.py`, `notify/push.py` — the digest email and push copy
- `tasks/alternates.py`, `tasks/time_scan.py`, `tasks/time_scan_runner.py`, `tasks/alternate_requirement.py`, `tasks/refresh_delta.py`
- the web flight list + briefing badge (`assessment: string | null`)
- iOS's `Assessment` enum

Adding a fourth state touches all of them.

## The decision

**What does a pilot see when we know nothing?**

1. A fourth state — grey / "No data" / "Not assessed" — rendered wherever GREEN/AMBER/RED are today.
2. Reuse the existing `null` assessment (web already types it `string | null`) and let each surface decide.
3. Something else — e.g. treat it as AMBER "unknown, look closer".

And the follow-ons:

- Does a no-data briefing still fire a **push notification**? (Probably not — it's not news.)
- Does the **email digest** say anything, or stay silent?
- What do **MCP / ChatGPT** return? They currently get `"GREEN"`, which is actively misleading to an agent reasoning about the flight.
- Does it change the flight-list tag?

Worth noting against our own framing: the briefing is an attention-director, not a go/no-go machine. A grey "we couldn't assess this" is consistent with that — arguably *more* consistent than a green that implies we looked and it was fine. But it's still a call to make.

## Suggested first step

Decide (1) vs (2), then do the backend + web slice; iOS and the connectors can follow.

Issue #387, opened 2026-07-12
Settings page: tiered advisory parameters (pilot vs advanced), catalog-driven ordering, setup interview, live preview


roznet:
# Settings page: tiered advisory parameters (pilot vs advanced), catalog-driven ordering, setup interview, live preview

## Problem

The profile settings page renders all 21 advisories with all **83 tunable parameters** as one flat list, grouped by weather category (icing first, airport buried mid-page, `flight_rules`/`sun`/`fronts` missing from the web `CATEGORY_KEYS` so they sort last by accident). It overwhelms most users, yet ~15 of those fields are genuinely valuable personal choices (crosswind limits, personal minima).

The parameters are not one kind of thing:

1. **Pilot/aircraft facts** (~17 fields) — personal minimums and capability limits. High value, per-pilot/per-plane.
2. **Advisory relevance toggles** (21 on/offs) — "I don't fly IFR", "not FIKI-equipped".
3. **Meteorological calibration** (~66 fields) — how the evaluators decide (`no_escape_pct_red`, `organized_shear_kt`, …). Calibrated deliberately (see `designs/meteorology-decisions.md`); almost nobody should touch these. This is the overwhelm.

## Design decisions (settled — do not relitigate)

- **No page-level compact/detailed mode.** Progressive disclosure per advisory row: toggle + pilot-tier fields always inline; the rest behind a collapsed "Advanced (n)" expander. "Compact" is just the default state; a global expand-all affordance covers power users.
- **Never invisibly hide a modified value.** A collapsed Advanced section containing non-default values shows a chip ("2 modified"). Modified values display their default alongside ("25 kt · default 20").
- **Metadata is backend-owned, three pieces at three levels** (one param is NOT enough):
  1. `audience: "pilot" | "advanced"` on `AdvisoryParameterDef` — drives inline vs expander. Default `"advanced"` so new params are safe-by-default. Exactly two values; no filter UI, no tag taxonomy.
  2. **Ordering lives in the registry, not the client.** Category order + entry order within category defined backend-side; `/advisories/catalog` returns entries in display order (plus an ordered category list with keys). Web deletes `CATEGORY_KEYS`; clients render in the order served. iOS/MCP inherit for free.
  3. **Setup interview is a separate declarative preset structure** — it cannot be derived from `audience` because it encodes identity facts spanning advisories (FIKI flips `fiki_icing` on AND `icing_escape` off).
[...]

PR #394, opened 2026-07-12
Add an UNAVAILABLE flight-level assessment — missing data must not read as GREEN (#392)

roznet:
Implements #392.

## The problem, and the half of it the issue didn't name

`derive_assessment_from_advisories` filtered out UNAVAILABLE advisories, found an empty list, and returned `GREEN` — "No advisory data available". The reason string was honest; the colour was not.

But fixing that function alone would have changed **nothing** for a normal briefing. The pack's traffic light is normally the **LLM digest's** assessment (`api/packs.py::_build_pack_meta` prefers `result.digest.assessment`); the derived value is only the provisional/fallback. And `guardrails.py` hard-enforces `VALID_ASSESSMENTS = {GREEN, AMBER, RED}`, so the LLM is structurally incapable of saying "no data" and nothing clamped it against the manifest. The derived fix would have been invisible on the surface a pilot actually looks at.

It was also hiding somewhere worse: the same function grades **every candidate departure time** in the time-scan. A slot we had no data for scored GREEN and could present as *the best time to fly*.

## Decisions

| | Choice |
|---|---|
| **State** | `UNAVAILABLE`. `NULL` keeps its existing, different meaning (no pack / not briefed yet). AMBER would conflate "we found a hazard" with "we found nothing to look at". |
| **LLM path** | Deterministic clamp. Pipeline skips the digest when nothing graded (no LLM cost to narrate an empty manifest); `_build_pack_meta` clamps as a backstop; `/digest/generate` 422s. |
| **Trigger** | Binary — *zero* graded advisories. Partial coverage still grades on what graded; the grey advisory cards carry the gaps. No coverage threshold: it would need calibration and a call on which advisories are "core". |
| **Notify** | None. Not weather news, not actionable — it says our data is missing, not that their weather changed. Holds even for a per-flight "always notify" override. |

Because the LLM never emits the new state, `guardrails.py`, the `WeatherDigest` `Literal`, and the eval-workbench validators are all untouched — a much smaller diff than "adding a fourth state touches all of them" suggested.

## Rendering was nearly free

- web `assessmentClass` already has `default → badge-none` (grey)
- iOS `Assessment` already has an `.unavailable` case (grey), and `AssessmentStringBadge` decodes with `?? .unavailable`

Only the briefing **banner** needed work — it interpolates its CSS class from the string (`assessment-${level}`), and `.assessment-unavailable` didn't exist, so it would have rendered as an unstyled white box. Same in `report/briefing.html`.

## Also fixed — the same defect, two more places

`web/ts/briefing-main.ts:677` and `TimingScenariosView.swift:431` were ternary chains whose **else-branch was GREEN**, so any unrecognised status — UNAVAILABLE above all — was painted as "assessed and clear".

## Verification

- `3365 passed, 9 skipped` — full backend suite
- New `tests/test_assessment_unavailable.py` (16 tests) covers the derivation, the `None`-vs-empty distinction, the digest clamp, the pipeline skip, and notify suppression. **Each has a control test** (a graded briefing still notifies; the digest still runs; the digest's verdict still wins) so none can pass vacuously.
- New API test drives the real `/digest/generate` endpoint → 422, asserting the LLM is never called.
- iOS `BUILD SUCCEEDED`; web `tsc` clean (the two `label-panel.ts` errors pre-exist on `main`).

## Follow-up, deliberately not done here

The `assessment_reason` contract (`"icing=RED, cloud=AMBER"` — red/amber only) **cannot express a gap**, so a *per-advisory* UNAVAILABLE is still invisible in the timing-grid dots, which fall back to `?? 'GREEN'`. The dot colour maps no longer paint unknown green, but closing this properly means extending the reason format — which changes user-visible email/report copy for normal packs. Worth its own decision.


roznet:
## Round 1 review — actioned

Rebased onto current `main` (PR #387) first — clean, no conflicts. All findings addressed in `62d62e3` except one I'm pushing back on.

### Fixed

**The alt-departure "Better" chip** — you were right, and this was the important one. `countAdvisoryStatuses` counted UNAVAILABLE as neither red nor amber, so an alt time where *nothing graded* scored `{red: 0, amber: 0}`, beat a planned time with real hazards, and rendered a green **"Better"** chip. A slot with no data presenting as the better time to fly — the exact defect this PR exists to remove, one hop down, in a file the PR had already touched.

Replaced with `compareAdvisories`, which mirrors the backend's own rule (`tasks/time_scan.py::_compare` — "either side unavailable → not comparable"): only advisories **both** sides graded are comparable; when nothing is comparable the chip is **withheld** rather than invented. Extracted to `helpers/alt-departure-compare.ts` so it's unit-testable without the Leaflet/DOM graph — **8 tests**, including the regression and controls proving a genuine improvement still reads "Better" (a fix that blanked real improvements would just trade one false reading for another).

**iOS hero-card English leak** — confirmed. `AdvisoryTabView` passed `assessmentReason` through unconditionally; for UNAVAILABLE that's the internal English diagnostic the web deliberately suppresses. Now substitutes localized copy.

**Agents handed an undocumented value** — MCP + ChatGPT contracts now list `UNAVAILABLE`, plus an `ASSESSMENT_UNAVAILABLE_NOTE` payload guardrail (sibling of `CROSS_CHECK_NOTE`/`MITIGATION_NOTE`) telling the model it is *not* a clear sky and must not be narrated as reassuring.

> **Worth flagging:** putting that note in the docstrings pushed two ChatGPT OpenAPI descriptions to **448** and **461** chars, over the connector's ~300 limit (every other endpoint sits deliberately under it). That would have broken the connector. Hence the payload guardrail — which is also where this codebase already puts exactly this kind of semantics.

**Minor** — `badge-unavailable` in both duplicated `assessmentClass()` impls; the eval tool's unstyled `U` badge; Siri phrasing ("graded unavailable" → "I couldn't assess your flight"); stale GREEN/AMBER/RED-only field comments.

### Not taken — on-demand briefing email

`POST /{timestamp}/email` isn't covered by the UNAVAILABLE notify suppression, and I don't think it should be. That suppression is deliberately about **automatic** notifications ("it's not news"). Here the user **explicitly asked** for the email — silently dropping their request would be worse than an honest empty-handed one. Email is English-only, so there's no locale leak either. I gave the state a deliberate colour rather than leaving it on the accidental grey fallback.

### Verification

`3388` backend tests · `478` vitest across 30 files · `tsc` clean (the two `label-panel.ts` errors pre-exist on `main`) · iOS **BUILD SUCCEEDED**.

Issue #393, opened 2026-07-12
Advisory evidence contract + single-assessment loop (from #389)


roznet:
Two ideas worth taking from #389, the independent #223 implementation. Both are structural; neither changes any meteorology. Do these **after** #391, so the per-evaluator tests exist to pin behaviour through the refactor.

The #389 branch is checked out at `../pr-389` for reference.

---

## Part 1 — Contract additions (small)

We shipped the highlight mechanism (#373/#375) with `AdvisoryHighlights` = `{ribbon[], regions[]}`. It's the better *rendering* contract than #389's and we're keeping it — our ribbon is a gapless partition of the route, so "unavailable here" can't render as silence, which is a property #389 has to reconstruct in the status dimension.

But #389 carries three things on its regions that ours doesn't, and they're the reason its chips can do more than ours:

- [ ] **`representative_model` on `RouteAdvisoryResult`.** Today "which model's geometry do we highlight?" is answered *twice*: the backend picks one to source `aggregate_detail` (`models/advisories.py:361`), and the frontend has `representativeModel()` (`web/ts/visualization/cross-section/advisory-highlights.ts:33`) — a line-for-line TypeScript reimplementation of the same rule. Two copies of one rule, maintained by hand, in two languages. Emit it from the backend, read it on the client, delete the TS copy.

  Convective already strains this: `convective.py:409` deliberately overwrites `aggregate_detail` with a *cross-model* range ("35–60% of route, peak HIGH") that belongs to no single model, while the chart still highlights exactly one. The field makes the choice explicit.

- [ ] **`primary_method_id` on `ModelAdvisoryResult`.** The stable id of the method that actually controlled the grade (`sfip`, `ogimet_nwp`, `dewpoint_depression`, and compound ones like `nwp_with_dd_floor`). This is what a method badge on the chip must be sourced from — *not* the user's selected method in settings, which may differ from the one that drove the result.

- [ ] **`reason_code` / `metric_id` / `method_id` on our existing `HighlightRegion`.** Stable, non-localised tokens. `metric_id` is what lets a chip jump the user to the *right* cross-section layer rather than just "the chart".

These are additive and legacy-safe; old packs deserialize with them absent.

---

## Part 2 — Single-assessment loop (big)

[...]

PR #395, opened 2026-07-12
Fix #391: prevent absent data from grading as clear/safe

### LINKED ISSUE #391: Missing data must never read as GREEN (advisory stack)
Follow-up to the independent audit in #389, which re-implemented #223 from scratch and — as a side effect — found a class of defect in our advisory stack that we have too.

**One root cause: absent data is graded as if it were assessed and clear.** Each evaluator counts a route point (or an airport, or a model) into its denominator as soon as *some* input exists, then guards the actual hazard predicate behind a `is not None` check. A point with a sounding but no vertical motion, a mountain point with no wind, an airport with no conditions — all score as "assessed, nothing flagged" → GREEN.

The `#389` worktree is checked out at `../pr-389` for reference. Its per-evaluator fixes are worth reading, but **do not copy its `guard_status_for_data_state`** — see "Do not do this" below.

## Already fixed on main

- [x] `worst([])` / `majority([])` returned GREEN → now UNAVAILABLE (`dec094f0`). This cascaded: the airport evaluators return `from_per_model(id, [], params)` with an *empty* list when their domain is absent, so Flight Category / Airport Wind / Density Altitude / LLWS were all showing GREEN with no airport data. Fixed by the same diff.
- [x] DD/NWP cloud Jaccard could exceed 1.0, suppressing disagreement flags (`75bbb4eb`). GFS layer lists self-overlap by construction — `_build_grib_layers` emits low/mid/high bands *and* a convective deck that spans all three. Measured 2.83 on a realistic set.
- [x] Route-map SFIP coloured on a private 20/50/80 scale while the backend uses 15/30/55, so SFIP 18 (LIGHT icing) rendered green "Low" (`b77c31d8`).

## To do — per-evaluator "assessed vs. missing"

Each needs a judgement call on what *assessed* means for that evaluator. The shape of the fix is the same everywhere: only count a unit into the denominator when the predicate could actually run, and let genuinely-clear evidence still grade GREEN.

- [ ] **`turbulence.py`** — `total += 1` fires when a *sounding* exists, but every hazard predicate sits behind `if vm is not None`. A model with soundings but no `vertical_motion` scores total=N, affected=0 → GREEN "smooth". Careful: CAT comes from `vm.cat_risk_layers` (Richardson) and is independent of omega — do not gate the whole point on omega, which is the mistake #389 made and then had to revert.
- [ ] **`mountain_wind.py`** — `total += 1` for each mountain point, then `if wind is None: continue`. If every mountain point's wind lookup fails, `max_wind` stays `0.0` → GREEN "no wave risk". Note the highlight ribbon already marks these points UNAVAILABLE correctly — the grade and the evidence disagree. Separately: `terrain_ft is None` (no elevation profile) → total=0 → GREEN "no terrain", i.e. missing elevation reads as "no mountains".
- [ ] **`cloud_top.py`** and **`vmc_cruise.py`** — `sounding is None → continue` *before* `total += 1`, so missing points shrink the denominator and a surviving clear subset stays GREEN. Only `total == 0` is UNAVAILABLE.
- [ ] **`freezing_precip.py`** and **`enroute_precip.py`** — `total += 1` per sounding, then `if precip is None: continue`. Points with no precipitation assessment stay in the denominator and dilute the percentage below the threshold. (2 snow points + 8 unassessable points reads as 20% snow, not 100%.)
- [ ] **Ogimet-NWP empty-zone path** — `assess_icing_zones_ogimet_nwp` returns `[]` when `clouds is None` (`sounding/icing.py:432`), and `_resolve_analyses` puts that empty list straight into the active slot (`tasks/advise.py:122`). `fiki_icing`, `icing_escape` and `ifr_feasibility` then read zero zones and grade **clear**. A user on Ogimet-NWP with a model that has no native cloud envelope gets GREEN icing from absent data. Distinguish "method unavailable" from "method ran, found nothing".
- [ ] **`registry.py:154`** — a bare `except Exception` logs and then simply doesn't append the result, so an evaluator that throws makes its advisory **vanish** from the briefing entirely. Return an explicit UNAVAILABLE result with diagnostics instead.
- [ ] **`vfr_feasibility.py:77` / `ifr_feasibility.py:73`** — literally `if ctx.airport_conditions is None: return GREEN, "", GREEN, GREEN`. The airport axis is hardcoded clear on missing data.
- [ ] **`airport_wind._wind_status`** — returns GREEN when both crosswind and gust are `None`; a wind-less condition row reads as calm.
- [ ] **`model_agreement.py:81`** — branches on `ModelDivergence.agreement` alone, but `agreement` is `GOOD` when every metric is absent. `models/analysis.py:840` *documents* this trap ("a consumer that branches on `agreement` alone must also check `mean`") and this consumer falls into it.
- [ ] **`descend_below_icing`** (`sounding/advisories.py:606`) — a freezing-rain model contributes `per_model_ft = None`, which is filtered out of `valid_alts`; `min()` of the remaining models then yields a descent escape. We recommend "descend to 3,000 ft" while one model says descending keeps you in freezing rain. Safety-relevant.
- [ ] **`affected_nm`** (`models/advisories.py` `build()`) — computed as `total_distance_nm * affected / total`, a point-count proportion rather than route geometry. Wrong on unevenly-spaced routes. #389's midpoint-owned-cell convention (each point owns the interval to the midpoints of its neighbours, union the qualifying intervals) is the right fix.

## Explicitly NOT in this issue

**SFIP no-omega weight (#389's A223-07).** They renormalise the T/RH/CLW weights when omega is absent, arguing the omega weight is "lost". I checked and I think they're wrong: `membership_vv` (sfip.py:152) returns **−0.3 to +0.5**, not 0–1 — it's a *signed modifier* where 0.0 means "neutral vertical motion", and `omega is None` also returns 0.0. So absent omega is already treated as "no information → no adjustment". Renormalising would make omega-less models (ICON) systematically more icing-prone than GFS with neutral omega. That's a **recalibration**, and it needs a `designs/meteorology-decisions.md` entry and evidence, not a drive-by fix. Their own audit policy forbids exactly this.

## Do not do this

#389's `evidence.py:127` converts `partial + GREEN → UNAVAILABLE` unconditionally, and its turbulence marks a model "partial" unless it has *both* omega and Richardson. Omega is model-dependent, so that would **blank the turbulence advisory for every omega-less model** even when its Richardson assessment is complete and genuinely clear. It trades false-green for false-blank. We want a coverage tolerance, not a binary.

## Testing

Two tests on main asserted the bug by name (`test_all_unavailable_returns_green`, `test_empty_returns_green`) and one web test asserted the old SFIP scale. Expect more of these — when a test enshrines missing-is-clear, fix the test.

roznet:
## What & why

Regression fixes for #391: several evaluators were incorrectly grading absent or unassessable data as clear (GREEN) or safe, when they should return UNAVAILABLE. This affected:

1. **Coverage tolerance** (VMC Cruise, Cloud Top): A clear verdict from a small subset of assessed points (e.g., 2 of 10) was grading GREEN, silently vouching for the 8 unassessed points. Now a clear verdict below the 50% coverage threshold becomes UNAVAILABLE.

2. **Turbulence**: Soundings without a `vertical_motion` assessment (lite analysis / old pack) were counting as smooth (GREEN). Now they're UNAVAILABLE. Also fixed: an omega-less model with valid CAT layers still grades normally (not blanked).

3. **Cloud Top**: Same coverage-tolerance issue as VMC Cruise.

4. **Model Agreement**: All-absent metrics (every model missing a variable) were graded as "GOOD agreement" → GREEN. Now only real comparisons (mean ≠ None) establish agreement.

5. **VFR/IFR Feasibility**: Missing airport data was hardcoded as GREEN. Now it's UNAVAILABLE, so the airport axis doesn't contribute a false clear verdict.

6. **Mountain Wind**: Missing elevation data was read as "no mountains" → GREEN. Now it's UNAVAILABLE. Also: points with no wind data now correctly grade UNAVAILABLE instead of 0 kt → GREEN.

7. **Airport Wind**: Rows with no crosswind AND no gust (no wind data) were grading GREEN. Now they're UNAVAILABLE.

8. **Freezing Precip / En-route Precip**: Unassessable points (no precip assessment) were diluting the denominator, understating percentages. Now they're excluded from the count.

9. **Ogimet-NWP Icing**: When Ogimet-NWP runs on a model with no native cloud envelope, it returns `[]` (method unavailable, not "no icing"). Now flagged via `active_icing_available=False` so icing evaluators grade UNAVAILABLE.

10. **Registry**: A throwing evaluator was silently dropped from results (reading as "not a concern"). Now it emits an explicit UNAVAILABLE result with a diagnostic.

11. **Altitude Advisories**: When any model shows FZRA, descent is not a safe universal escape. The FZRA model's `per_model_ft=None` was being filtered out, allowing other models' escapes to stand. Now the None is preserved so `min()` correctly yields None (infeasible).

12. **Affected extent (nm)**: Point-count proportions misplace extent on unevenly-spaced routes. Now prefers the flagged ribbon span when present (correct geometry), falls back to proportion only when no ribbon.

## Issue linkage

Closes #391

## Testing / verification

Comprehensive regression test suite added:
- `test_clear_subset_below_coverage_is_unavailable` (VMC Cruise, Cloud Top): verifies coverage tolerance
- `test_hazard_on_partial_coverage_still_flags`: ensures flagged verdicts bypass coverage check
- `test_no_vertical_motion_is_unavailable` (Turbulence): lite analysis handling
- `test_omega_less_model_with_cat_still_grades`: CAT assessment without omega
- `test_all_absent_metrics_is_unavailable_not_good` (Model Agreement): absent-metric handling
- `test_no_airport_axis_not_hardcoded_green` (VFR/IFR): missing airport data
- `test_no_airport_but_clear_enroute_still_green`: guards against over-correction
- `test_ogimet_nwp_no_envelope_is_unavailable_not_green` (new file): Ogimet-NWP method unavailable
- `test_throwing_evaluator_yields_unavailable_not_vanishing` (Registry): exception handling
- `test_uneven_route_uses_ribbon_span_not_proportion` (Aggregation): affected_nm geometry
- `test_mountains_

https://claude.ai/code/session_01Br1KMnCkvuUZBuVMsBdQXD

### DIFF
--- a/designs/advisories.md
+++ b/designs/advisories.md
@@ -437,7 +437,7 @@ Recalculate loads route analyses + elevation + cross-sections from disk, applies
 
 ## Gotchas
 
-- Evaluator exceptions are caught and logged — one failure doesn't break the whole advisory set
+- Evaluator exceptions are caught and logged — one failure doesn't break the whole advisory set. The failed evaluator is not dropped: the registry appends an explicit UNAVAILABLE `RouteAdvisoryResult` with a localized diagnostic `aggregate_detail` (via `adv_t("evaluation_failed", …)`), so a crash reads as "could not assess", never as a silently-absent "not a concern" (#391)
 - `ModelAgreementEvaluator` has `per_model=["all"]` (not actual model names) since it's cross-model
 - `format_extent` falls back to percentage-only if route has too few points for meaningful distance
 - `wind_at_altitude` picks the level via `pick_wind_at_pressure` and the hour via `at_time` — don't reintroduce a "first hourly" shortcut, it lags the route point's valid time on long legs
--- a/src/weatherbrief/analysis/advisories/_helpers.py
+++ b/src/weatherbrief/analysis/advisories/_helpers.py
@@ -351,6 +351,29 @@ def min_icing_clearance(
     return min_dist
 
 
+# A clear (GREEN) verdict may only speak for the whole domain when at least this
+# fraction of it was assessable. Below it, a clear subset of a mostly-unassessed
+# domain is UNAVAILABLE — not GREEN (#391). Data-coverage tolerance (missing-data
+# handling), NOT a meteorological threshold.
+_MIN_ASSESSED_FRACTION = 0.5
+
+
+def below_coverage(assessed: int, domain: int) -> bool:
+    """True when a would-be-GREEN verdict rests on too small a share of its domain.
+
+    ``assessed`` is the count of units the evaluator could actually grade;
+    ``domain`` is the assessable universe it speaks for (route points for most
+    evaluators, mountain points for ``mountain_wind``). Used to downgrade a
+    would-be-GREEN to UNAVAILABLE when coverage is thin (#391) — a clear subset
+    does not establish the unassessed remainder is clear. Only ever applied to a
+    GREEN verdict; a flagged (AMBER/RED) verdict always stands, so real hazard
+    evidence on partial coverage is never diluted. Returns False for an empty
+    domain (``domain == 0`` is handled as UNAVAILABLE by the evaluators' own
+    "nothing to assess" branch, not here).
+    """
+    return domain > 0 and assessed < domain * _MIN_ASSESSED_FRACTION
+
+
 def pct_above_threshold(
     affected: int,
     total: int,
--- a/src/weatherbrief/analysis/advisories/airport_wind.py
+++ b/src/weatherbrief/analysis/advisories/airport_wind.py
@@ -48,6 +48,12 @@ def _wind_status(
     gust_red: float,
 ) -> AdvisoryStatus:
     """Determine wind status from crosswind and gust against thresholds."""
+    # No crosswind AND no gust means we have no wind data to grade — the row is
+    # unassessable, not calm (#391). A genuinely calm airport still has a
+    # crosswind of 0.0 (best_runway present with 0 wind), which grades GREEN.
+    if crosswind_kt is None and gust_kt is None:
+        return AdvisoryStatus.UNAVAILABLE
+
     status = AdvisoryStatus.GREEN
 
     if crosswind_kt is not None:
@@ -159,7 +165,12 @@ class AirportWindEvaluator:
                 s = _wind_status(xwind, cond.wind_gust_kt, xwind_green, xwind_red, gust_green, gust_red)
                 statuses.append(s)
 
-                wind_str = _format_wind_detail(cond, rwy_id, loc)
+                # A wind-less row is UNAVAILABLE (see _wind_status) — say so
+                # rather than rendering it as "calm".
+                if s == AdvisoryStatus.UNAVAILABLE:
+                    wind_str = adv_t("no_data", loc)
+                else:
+                    wind_str = _format_wind_detail(cond, rwy_id, loc)
                 label = adv_t(label_key, loc)
                 parts.append(f"{label} {icao}: {wind_str}")
 
--- a/src/weatherbrief/analysis/advisories/cloud_top.py
+++ b/src/weatherbrief/analysis/advisories/cloud_top.py
@@ -5,6 +5,7 @@ from __future__ import annotations
 from weatherbrief.analysis.advisories import RouteContext
 from weatherbrief.analysis.advisories._helpers import (
     FlaggedCell,
+    below_coverage,
     build_regions,
     build_ribbon,
     format_extent,
@@ -22,15 +23,6 @@ from weatherbrief.models import (
     ModelAdvisoryResult,
     RouteAdvisoryResult,
 )
-
-
-# A clear (GREEN) verdict may only speak for the whole route when at least this
-# fraction of route points carried a sounding for the model. Below it, a clear
-# subset of a mostly-unassessed route is UNAVAILABLE — not GREEN (#391). This is
-# a data-coverage tolerance (missing-data handling), NOT a meteorological
-# threshold: a genuinely-clear, well-covered model still grades GREEN and any
-# flagged (AMBER/RED) verdict stands regardless of coverage.
-_MIN_ASSESSED_FRACTION = 0.5
 
 
 @register
@@ -162,10 +154,7 @@ class CloudTopEvaluator:
             # is too small to represent the route becomes UNAVAILABLE — a clear
             # 2-of-20-points model has not established the other 18 are clear. A
             # flagged verdict is never downgraded (real hazard evidence stands).
-            if (
-                status == AdvisoryStatus.GREEN
-                and total < len(ctx.analyses) * _MIN_ASSESSED_FRACTION
-            ):
+            if status == AdvisoryStatus.GREEN and below_coverage(total, len(ctx.analyses)):
                 status = AdvisoryStatus.UNAVAILABLE
                 detail = adv_t("no_data", loc)
 
--- a/src/weatherbrief/analysis/advisories/cloud_top.py
+++ b/src/weatherbrief/analysis/advisories/cloud_top.py
@@ -24,6 +24,15 @@ from weatherbrief.models import (
 )
 
 
+# A clear (GREEN) verdict may only speak for the whole route when at least this
+# fraction of route points carried a sounding for the model. Below it, a clear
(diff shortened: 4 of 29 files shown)


roznet:
## Addressed review findings (rebased onto current `main`, force-pushed)

**Finding 1 + 2 (Important) — coverage tolerance not applied to siblings.** Fixed. Extracted a shared `below_coverage(assessed, domain)` helper in `_helpers.py` and applied it to every would-be-GREEN verdict where the same "clear subset of a mostly-unassessed domain grades GREEN" defect was reachable:

- `turbulence`, `model_agreement`, `freezing_precip`, `enroute_precip`, `icing_escape`, `fiki_icing`, and `ifr_feasibility`'s icing axis — domain = route points.
- `mountain_wind` — domain = **mountain points** (so a genuinely-flat route still grades GREEN, preserving the "GREEN not UNAVAILABLE" UX).
- `cloud_top`/`vmc_cruise` refactored onto the shared helper (were inline).

A flagged (AMBER/RED) verdict is **never** downgraded — added over-correction guards verifying SEVERE CAT / active FZRA / OVC at 2-of-10 assessed points still flag, and full-coverage clear still grades GREEN.

**Finding 3 (Important) — ribbon-derived `affected_nm` diverged from `affected_pct`/`affected_points`.** Reverted `affected_nm` to the point-count proportion so the three stay consistent (the ribbon's amber/red membership doesn't match `affected` for `enroute_precip`'s light rain or the feasibility composites' multi-axis ribbons, which is exactly the contradiction you flagged). The genuinely geometry-accurate extent needs per-point affected distances threaded into `build()`; noted in-code as belonging with the #393 single-assessment refactor rather than here.

**Finding 4 (Minor) — raw-English crash detail.** Added an `evaluation_failed` localized template and route the registry fallback through `adv_t(ctx.locale)`.

Batched all four into one push since #1/#2 required a round anyway. New regression tests per sibling (clear-low-coverage → UNAVAILABLE) plus the over-correction guards; the `affected_nm` geometry test was replaced with a consistency test. Full suite green (3400+ passing).

---

roznet:
## Addressed round-2 review findings (rebased onto current `main`, force-pushed)

**Finding 1 (Important) — `ifr_feasibility` convective axis had no coverage tolerance.** Added the same `below_coverage` guard the icing axis already had, so a no-convection verdict from soundings at too few route points is UNAVAILABLE rather than a false GREEN into `_worst_status`.

**Finding 2 (Important) — `vfr_feasibility` en-route cloud axis had no coverage tolerance.** Added the guard (and the missing `below_coverage` import), matching the standalone `vmc_cruise` fix.

**Both composites also got a composite-level coverage guard.** The per-axis guards alone don't close the aggregate: a single assessed corridor point, the missing-precip→GREEN mapping, or a real-GREEN airport axis can each carry the composite to GREEN via `worst()` — yet VFR/IFR feasibility is fundamentally en-route-driven, so GREEN with 8/10 route points unassessed overstates confidence. A would-be-GREEN composite on thin en-route coverage now grades UNAVAILABLE. Verified: 2/10 clear **+ confirmed-VFR/IFR airports → UNAVAILABLE**; HIGH convective at 2/10 still **RED**; full-coverage clear still **GREEN**.

**Finding 4 (Minor, safety-adjacent) — icing-escape mitigation priced Ogimet-NWP-unavailable cells as clear.** `_icing_cell_cost` now returns `INF` for an unassessable cell above the freezing level (below it is genuinely warm → still free), so the solver can't route an altitude "escape" through an unassessed segment as if confirmed ice-free. Advice-only, but no longer misleading.

**Finding 5 (Minor) — stale doc.** Updated the `designs/advisories.md` Gotchas line to describe registry's UNAVAILABLE-with-diagnostic crash behavior.

**Finding 6 (Minor) — needless local import.** Moved registry's `adv_t` import to top level (`strings.py` has no dependency back on `registry`, so it dodged a non-existent cycle).

**Finding 3 (Minor) — deferred.** The corridor-deck "no sounding at any corridor point → GREEN" case is now moot for the aggregate grade (the composite coverage guard covers it). I left the per-axis corridor-UNAVAILABLE nicety as a fast-follow rather than reworking the corridor logic here.

New regression tests: conv/en-route/composite coverage (incl. the confirmed-airport case), a HIGH-convective partial-coverage over-correction guard, and a `_icing_cell_cost` unit test. Full suite green.

---

PR #409, opened 2026-07-13
Populate advisory method_id from the effective analysis method (#408)

### LINKED ISSUE #408: Evidence contract: populate method_id from the effective (not requested) analysis method
Follow-up to #393 (PR #406) and #402 (PR #407), both now merged.

## The gap

Two halves of the same idea exist, built from opposite ends, never joined:

**Provenance is produced but never consumed.** `_resolve_analyses` (`tasks/advise.py`) already stamps `cloud_method_effective` on each sounding (`models/analysis.py:758`) — `nwp` / `nwp_synthesized` / `dd` — and crucially stamps `dd` **when the NWP layers were missing and it silently fell back** (`advise.py:130-138`). Nothing reads it. It has been dead since `04bfa795`.

**Provenance is consumed but never produced.** #393 declares `method_id` on every evidence region (`models/advisories.py:197`, threaded through `_helpers.py:260`) and `primary_method_id` on the advisory (`models/advisories.py:346`), documents it as *"the analysis method that produced the region's evidence (e.g. `ogimet_nwp`, `sfip`, `nwp_with_dd_floor`), when one controlled it"*, types it in the web (`web/ts/types/advisories.ts`) — and **no evaluator populates it**. It is `None` in every production path; only `tests/analysis/advisories/test_evidence.py:123` sets it, to prove the plumbing carries it.

So this issue is not about inventing a mechanism. It is about finishing one.

---

## Part 1 (producer) — complete the effective-method trio

Add alongside the existing `cloud_method_effective` on `SoundingAnalysis`:

- **`icing_method_effective`** — `ogimet_nwp` / `ogimet_dd` / `sfip_nwp`; unset when the method could not run (pairs with the existing `active_icing_available=False` at `advise.py:150`, the #391 "method unavailable, not clear sky" flag).
- **`convective_method_effective`** — `nwp` / `thermo`.

Stamp both in `_resolve_analyses` at the exact branch where each fallback decision is taken — the cloud code at `advise.py:126-138` is the pattern to copy.

### This closes a live honesty gap

`advise.py:169`:

```python
nwp = sounding.convective_nwp
updates["convective"] = nwp if nwp is not None else sounding.convective_thermo
```

A pilot who asked for **NWP convective** silently grades on **thermo** wherever `convective_nwp` is None, and **nothing anywhere records that it happened.** This is worth fixing on its own merits, independently of whether `method_id` is ever wired.

## Part 2 (consumer) — populate `method_id` / `primary_method_id`

Evaluators stamp `method_id = sounding.<axis>_method_effective` on the evidence cells they emit. `build_regions` already threads `cell.method_id → region.method_id` (`_helpers.py:260`), so there is **no plumbing to add**.

Axis mapping (verified against the code):

| Evaluator | Axis |
|---|---|
| `icing_escape`, `fiki_icing`, `ifr_feasibility` | icing |
| `cloud_top`, `vmc_cruise` | cloud |
| `enroute_precip`, `turbulence`, `mountain_wind`, `freezing_precip`, `model_agreement` | none → `method_id` stays `None` |

That last row is not a gap — it is exactly what *"when one controlled it"* in the docstring is reserving space for.

**`primary_method_id`** rolls up from the driving region, and must reuse #393's existing `representative` lookup (`models/advisories.py:412-443`) rather than introducing a second one. The two-loop drift risk is precisely what #393 collapsed; do not re-open it.

---

## ⚠️ The trap this issue exists to prevent

**`method_id` must record the EFFECTIVE method, never the REQUESTED one.** They diverge exactly where the evidence matters:

- cloud: asked `square_nwp` → actually graded on **DD layers** when `nwp_cloud_layers` was None
- convective: asked `nwp` → actually graded on **thermo** when `convective_nwp` was None
- icing: asked `ogimet_nwp` → the method **could not run at all** without an NWP cloud envelope

An implementation that stamps `method_id` from the requested method **lies at precisely the points where a pilot most needs the truth** — labelling a region `square_nwp` when it graded on DD. That is the same failure class #391/#393 exist to eliminate: fallback data presenting as something it isn't. Putting a single requested-method label on `RouteContext` is the obvious move and it is the wrong one.

**And #402/#407 compounds it.** Since sparse-persistence, most profiles store **no engine-method key at all** (absence now *means* the declared default via `ENGINE_METHOD_DEFAULTS`). So `profile_settings.get("icing_method")` returns `None` for the majority of users. Anyone reaching for the profile to fill this field gets nulls for most users and a lie for the rest. **Do not source `method_id` from the profile.**

## Payoff

- Gives `cloud_method_effective` a consumer at last.
- Surfaces `nwp_synthesized` as distinct from model-native `nwp` — a chip can honestly tell a pilot *"this cloud grade came from synthesized layers, not the model's own"*, decision-relevant information we currently compute and throw away.
- Makes the #393 evidence contract honest about provenance under fallback, which is the contract's entire purpose.

## Tests

- Each fallback branch in `_resolve_analyses` stamps the effective method it actually used (NWP requested + NWP data absent → `dd` / `thermo`).
- An advisory graded under fallback reports the **fallback** method in `method_id`, not the requested one.
- A sparse profile (no engine keys — the post-#407 majority) still reports a concrete `method_id`, never `None`, on a method-controlled axis.
- Evaluators with no engine-method axis report `method_id is None`.

roznet:
## What & why

Finishes the #393 evidence contract, whose two halves were built from opposite ends and never joined:

- **Provenance produced but never consumed.** `_resolve_analyses` stamped `cloud_method_effective` on each sounding, but nothing read it — dead since `04bfa795`.
- **Provenance consumed but never produced.** #393 declared `HighlightRegion.method_id` and `ModelAdvisoryResult.primary_method_id`, threaded them through `build_regions`, and typed them in the web — but no evaluator populated them. They were `None` in every production path.

This joins the two ends. **The rule that matters:** `method_id` records the *effective* method, never the *requested* one — they diverge exactly where a fallback fires, which is where a pilot most needs the truth (a region graded on DD labelled `square_nwp` would be a lie). It is never sourced from the profile, where the post-#407 sparse majority store no engine key at all.

### Part 1 — producer
Add `icing_method_effective` and `convective_method_effective` to `SoundingAnalysis`, siblings of `cloud_method_effective`, and stamp them in `_resolve_analyses` at the exact branch each fallback decision is taken:
- **convective** → `"nwp"`, or `"thermo"` when `convective_nwp` was absent and it silently fell back. This closes a live honesty gap on its own merits: a pilot who asked for NWP convective grades on thermo wherever `convective_nwp` is None, and nothing recorded it.
- **icing** → `"ogimet_nwp"` / `"sfip_nwp"`; left **unset** when Ogimet-NWP could not run at all (no native cloud envelope), pairing with the existing `active_icing_available=False`.

### Part 2 — consumer
The five method-controlled evaluators — `icing_escape`, `fiki_icing`, `ifr_feasibility` (icing axis) and `cloud_top`, `vmc_cruise` (cloud axis) — stamp `method_id = sounding.<axis>_method_effective` on their evidence cells (`build_regions` already threads `cell.method_id → region.method_id`, so no plumbing) and roll `primary_method_id` up from the **driving region** via a new `driving_method_id` helper: the flagged region whose severity matches the model's grade, read from the same `highlights` the grade produced so the badge cannot drift from the geometry. The aggregate view reuses the existing `representative_model` to pick which per-model `primary_method_id` a chip shows — no second selection rule.

Evaluators with no engine-method axis (`enroute_precip`, `turbulence`, `mountain_wind`, `freezing_precip`, `model_agreement`) stamp nothing → `method_id` stays `None`, exactly the "when one controlled it" space the docstring reserves. Convective's effective method is *produced* (the honesty fix) but not yet consumed by a `method_id` axis.

## Issue linkage

Closes #408

## Testing / verification

- New `tests/analysis/advisories/test_method_provenance.py` (18 tests) covering the four scenarios the issue specified: each fallback branch stamps the method it actually used; an advisory graded under fallback reports the fallback method (not the requested one); a sparse profile still reports a concrete method, never `None`, on a method-controlled axis; and evaluators with no engine-method axis report `method_id is None`.
- Full suite: **3468 passed, 12 skipped**. The one failure (`test_google_login_redirects`) is a pre-existing, environment-dependent OAuth-redirect test that also fails on a clean tree without these changes.
- Design docs updated: `advisories.md`, `analysis.md`, `data-models.md`, and the `convective-analysis.md` "Priority: Low" recommendation marked done.

---
_Generated by [Claude Code](https://claude.ai/code/session_01LeawZfXjzEWvzcZ33GQiZc)_


roznet:
Thanks for the review — all three findings addressed in 427213b.

**Important — method badge lost on percentage-driven RED (fixed).** You're right, and it's exactly the worst case. Fixed centrally in `driving_method_id` rather than per-evaluator:
1. Exact stamped match at the grade severity (unchanged).
2. If none, fall back to the highest-severity stamped region **strictly below** the grade — the AMBER-capped regions the coverage percentage climbed over (cloud_top ≥60%, ifr `icing_pct`, fiki clear-cruise fraction). A GREEN grade has no lower region, so it still yields no badge.

I did **not** use the plain "highest-severity stamped region" fallback, because in the IFR composite that would misattribute a **convective-driven** RED (method-less `tower`) to the concurrent AMBER icing axis. So there's a guard: if an *unstamped* region reached the grade severity, a method-less axis drove it → return `None`. New test `test_ifr_convective_driven_red_does_not_borrow_icing_method` pins that down, and the cloud_top/ifr/fiki integration tests now assert `primary_method_id` on the RED grade (the assertion gap that let this through).

**Minor — misleading `"ogimet_dd"` comments (fixed).** Corrected in `icing_escape.py` and `models/analysis.py`: explicit `ogimet_dd` needs no swap, so the field stays `None` — same accepted no-swap gap as cloud's DD source. The comments now say so instead of implying it's stamped.

**Minor — asymmetric field freshness (fixed).** `_resolve_analyses` now sets `icing_method_effective` explicitly (to `None`) on the no-envelope branch, matching cloud/convective which write their effective on every branch, so a stale value can't survive `model_copy`.

Full advisory + icing/cloud/convective/connector suites green (628 passed).

---

roznet:
Good catch — all three addressed in 8864568.

**Critical — wrong-axis badge in the IFR composite (fixed).** You're right that severity-only matching over the pooled region list has no per-axis linkage, and the round-1 unstamped guard doesn't help because `exact` isn't None — it's just the wrong axis. Rather than teach `driving_method_id` about axes, IFR (the only composite that mixes a method-bearing axis with a method-less one) now resolves its badge itself: it runs `driving_method_id` over the **icing-only** regions (single-axis, so the helper is correct there, including the extent-escalated RED) and then **gates on `icing_status == status`** — the icing method is badged only when the icing axis actually reached the composite grade. Your exact scenario (1/20 in-buffer icing → AMBER region but `icing_pct` 5% < 20 → icing GREEN; one MODERATE convective point → AMBER composite) now yields `primary_method_id = None`, pinned by `test_ifr_below_threshold_icing_does_not_borrow_method_for_conv_amber`.

**Important — `build_regions` merging across a `method_id` change (fixed).** Agreed this is the same silent-lie class the PR targets, and per-point effective method genuinely can vary within one model (NWP layers at some points, DD fallback at others). `method_id` now joins the run key, so a method change splits the region instead of mislabeling it. `reason_code`/`metric_id` stay first-cell (dominant reason / target layer), and I updated the `FlaggedCell` docstring to match. Covered by `TestBuildRegionsMethodSplit` (splits on change, still merges when uniform).

**Minor — escalation tie-break (documented).** Added a comment noting the tie-break is route-order and only reachable within a single method-bearing axis (composites now resolve per-axis first), so any tied member is a fair representative.

Suites green: advisories + icing/cloud/convective/connector/highlight/altitude (631, and 508 across the broader advisory-touching set).

---

roznet:
Fixed in 9f10a94 — and taken to the root this time.

You're right about the missing upward direction (AMBER grade off RED-only regions in `vmc_cruise`/`icing_escape`). Rather than add the mirror fallback and risk a round-4 in the *next* direction, I removed the severity-matching from `driving_method_id` entirely. The key realization: after round 2 moved the IFR composite to per-axis resolution, **every** caller now passes the regions of a single method-bearing axis (the four single-axis evaluators emit one region kind; IFR passes only its icing regions). Within one axis all flagged regions carry that axis's effective method, so the badge is simply the method of the flagged evidence — the **highest-severity stamped region** — with no comparison of region severity to the grade at all.

That's immune to *both* escalation directions (the round-1 escalate-past and the round-3 land-below cases collapse to the same code path) and it's less code. Behaviour is unchanged for every previously-correct case; the IFR composite still gates its icing badge on `icing_status == status`, so a convective-driven grade still shows no icing badge.

Tests rewritten to the single-axis contract (both directions, highest-severity representative, flagged-gate) plus `vmc_cruise`/`icing_escape` integration regressions for the grade-below-region case. Suites green: 633 advisory + icing/cloud/convective/connector, 510 across the broader advisory-touching set.

Note: this is the third review round, so I'm pausing automated iteration here and handing back to @roznet — happy to keep going on any further findings on request.

---

Issue #405, opened 2026-07-12
S2: Sparsify prod settings + stop seeding defaults in the templates (#402)


roznet:
Slice 2 of #402. Depends on S1 (#403).

## Change

**One-time sparsify of `flight_profiles.settings_json`:**

> delete any `advisories.params[adv][key]` whose value equals the current catalog default;
> delete `icing_method` / `cloud_method` / `convective_method` equal to the S1 declared default.

This is lossless because of the audit finding in #402: **every template-seeded param already differs from its catalog default**, so the rule cannot touch a deliberate override — template's or pilot's.

**Do NOT touch `advisories.enabled`.** See #402 — `fronts` declares `default_enabled=False` but is injected by default when the `auto_front_detection` artifact exists, so an explicit `fronts: false` is a real opt-out that this rule would wrongly delete.

**Also stop seeding the defaults in `configs/system_profiles.json`** — remove the three method keys from any template whose value equals the declared default, so new profiles are born sparse. Templates that deliberately differ keep theirs.

## Expected effect on prod

- 95 profiles touched (81 users)
- 5,149 param values deleted, 195 preserved
- **zero grading outcomes changed** — every deleted value is identical to the default the engine already applies

## Vehicle

Alembic data migration (plain Python, dialect-agnostic — SQLite dev / MySQL prod) **or** extend the lazy `_MIGRATIONS` registry in `api/user_migrations.py` to run over `flight_profiles` (it currently only touches the legacy `user_preferences.app_prefs_json`, which the pipeline no longer consults for these fields). The lazy registry is the established pattern; alembic is simpler for a one-shot. Pick during implementation.

Import the **live** catalog at migration time rather than snapshotting the defaults as a literal — that is what guarantees the no-op. (Snapshotting is more reproducible but would silently stop being a no-op if a default moved between merge and deploy.)

## Safety

- Dry-run mode printing before/after counts per profile.
- A test asserting the no-op directly: for a sample of real prod-shaped profiles, the advisory results are **byte-identical** before and after sparsifying.
- Verify the prod numbers above are reproduced by the dry run before committing the write.

## Acceptance for #402

Move one real default in code, deploy, confirm via the advisory-preview endpoint that untouched profiles follow it and customized profiles do not.

---

## Update — 2026-07-14, after #410 merged

#410 landed (PR #413) and changed the ground this slice stands on. Three corrections to the spec above:

**1. `cloud_method` is now `cloud_source` — the "Change" section above is stale.**

[...]

Issue #223, opened 2026-06-07
Advisory cross-section presets — Phase 2/3 (highlighting, emphasis, Skew-T linkage, non-XS chip actions)


roznet:
Follow-up to #219 (Phase 1, shipped via #220). The Phase-1 config (`advisory-presets.ts`) was deliberately built to absorb these as **new optional directive fields** on `AdvisoryPreset` + one applier branch each — no change to existing presets or call sites (see #219 Decision 6).

**Update 2026-07-10 — this is now the umbrella.** The highlighting work was designed (scrim = focus + verdict ribbon = judgement, backend-owned geometry) and sliced into implementation issues:

- #373 — **Highlight mechanism**: `AdvisoryHighlights` schema (ribbon partition + cutout regions + peak) on `ModelAdvisoryResult`, web scrim/ribbon rendering, chip → representative-model switch, Highlight layer toggle, deep-link — proven on `vmc_cruise` + `convective`. Full design rationale and decisions live there.
- #374 — **iOS rendering** of the same (data-driven — no per-advisory config to mirror). Depends on #373 only.
- #375 — **Remaining advisory emitters** (`icing_escape`, `fiki_icing`, `turbulence`, `cloud_top`, `vfr`/`ifr` composites, `mountain_wind` behind a mock-first gate) + new `freezing_precip` / `enroute_precip` chips. Depends on #373 only; parallel with #374.

## Original deferred items — disposition

- **Area / band highlighting** *(highest value)* — → #373 / #374 / #375. The `highlights?: HighlightDirective[]` preset-field sketch below was revised: highlight geometry is per-advisory-instance **data** (emitted by evaluators, per model), not static preset config — a bare dropdown lens has no advisory instance to highlight. See #373 "Design decisions".
- **Layer emphasis** — **dropped.** The scrim subsumes it: it dims by *region*, so even the relevant layer's non-affected extent is de-emphasized, which per-layer alpha can't express. Phase-1's clean-slate resolver already removes irrelevant layers entirely.
- **Skew-T overlay linkage** — `skewtOverlays?` + side-panel directives shipped with #308. The remaining piece, jump-to-worst-point, is covered by `peak_dist_nm` in #373 (chip selects the peak route point; the existing cross-section↔Skew-T link follows).
- **Chip actions for non-cross-section advisories** — still open here: model-quality → Compare mode; airport advisories → airport-profile drawer; fronts → map layout.
[...]

PR #413, opened 2026-07-13
Split cloud_source from cloud style; wire the cross-section to the graded profile; de-fossilize the DD template seeds (#410)

### LINKED ISSUE #410: Split cloud_source from cloud style; wire the cross-section to the profile it's graded on; de-fossilize the DD template seeds (#402)
Slice of #402. Depends on #403 (merged, **not yet deployed**). Must land **before** #405 — see *Ordering*.

## Summary

Three defects, one root cause: **`cloud_method` conflates a grading choice with a render style, and the cross-section reads the engine methods from a table nobody writes.**

1. **The profile's engine methods never reach the cross-section.** `preferredMethods` — which drives compact-mode layer selection — is read from the *account-level* `app_prefs_json`, not from the profile. That field is empty for **all 534 users**, so it falls back to a constant. Every user's compact-mode cross-section shows NWP clouds + Ogimet-NWP icing + NWP convective, **regardless of the profile they are graded on**.
2. **The settings page's "Cloud Style" select does nothing at all.** It writes a string no renderer reads. The real cloud style already lives in `vizSettings.cloudStyle` (localStorage), driven by the cross-section's own control.
3. **The DD engine seeds in `vfr_only` / `ifr_conservative` are a fossil** — the pre-`001_method_defaults_v2` defaults, never updated when the account-level defaults moved to NWP. They pin **873 profiles (61% of prod)** to an engine nobody chose.

Net effect today: 873 profiles are *graded* on DD icing / DD cloud / thermo convective while their cross-section *displays* NWP layers. The chart paints evidence the grade never used — exactly the failure mode #391/#393/#408 exist to kill.

This also silently defeats #405: its rule is "delete any stored value equal to the current default", and `square_nwp` (the declared default) exists on **7 of 1,428 profiles**. As written, #405 would de-seed clouds on seven profiles and leave the rest frozen.

---

### 1. The divergent read

`web/ts/briefing-main.ts:230`:
```ts
preferredMethods = { clouds: prefs.cloud_method, icing: prefs.icing_method, convection: prefs.convective_method };
```
`prefs` comes from `fetchPreferences()` → `GET /user/preferences` (`web/ts/adapters/preferences-adapter.ts:114`), which serves `app_prefs_json` via `_parse_app_prefs` (`src/weatherbrief/api/preferences.py:172-190`). The web's `PreferencesUpdate` interface has **no `cloud_method` field**, so the web has never written it. The settings page writes to the *profile's* `settings_json` instead.

Prod (`user_preferences`, 534 rows):
```
account cloud_method:      {None: 534}
account icing_method:      {None: 534}
account convective_method: {None: 534}
```

So `_parse_app_prefs` always falls through to `ENGINE_METHOD_DEFAULTS`, and `preferredMethods` is the constant `{clouds: 'square_nwp', icing: 'ogimet_nwp', convection: 'nwp'}` for every user. Same values pre- and post-#403 — this is live in prod today, not introduced by #403.

Severity: `loadDisplayMode()` (`web/ts/store/briefing-store.ts:20-26`) defaults to **`compact`** — the exact mode `getCompactLayerOverrides(preferredMethods)` governs.

The account-level engine methods are otherwise **dead**: the pipeline reads exclusively from `profile_settings` (`src/weatherbrief/api/packs.py:1111`), and the only live caller of `load_service_toggles()` reads a boolean out of it.

### 2. Style is already a UI setting

- `web/ts/visualization/types.ts:34` — `cloudStyle?: 'natural' | 'soft' | 'square'` on `VizSettings`
- `web/ts/store/briefing-store.ts:59` (default `'square'`), `:80` (persisted to `localStorage['wb_vizSettings']`), `:682` (`setCloudStyle`)
- `web/ts/visualization/controls/panel.ts:113-125` — the cross-section's own compound cloud control: source checkboxes + a style `<select>` with the same three options as the settings page

So the two controls are already disconnected: changing style in Settings does nothing to the chart, and changing it on the chart does nothing to the profile. **62 profiles / 52 users** carry a styled `cloud_method`; none of it has ever rendered anything.

The backend has never wanted the style — `_cloud_source_from_method` (`src/weatherbrief/tasks/advise.py:62-74`) throws it away:
```python
return "nwp" if cloud_method.endswith("nwp") else "dd"
```

The conflation already costs us: `web/ts/visualization/cross-section/advisory-highlights.ts:88-93` has to parse the style off, swap the source, and recompose, with a comment explaining why. That block deletes itself once the fields are separate.

**iOS already models this correctly** and needs zero changes: `CloudSource` and `CloudStyle` are orthogonal enums derived from the layer id (`app/flyfun-weather/flyfun-weather/Views/CrossSection/Layers/CrossSectionPresets.swift:46-62`, `:179`), persisted in `UserDefaults.crossSectionEnabledLayers`. iOS never reads `cloud_method`. It is the model the web should converge on.

### 3. Prod profile census (1,428 profiles, style normalized to source)

| Class | Count | Action |
|---|---:|---|
| `vfr_only` / `ifr_conservative` carrying the exact fossil triple `(ogimet_dd, dd, thermo)` | **873** | **clear → follow the NWP default** *(grading change)* |
| `ifr_fiki` carrying the all-NWP triple (= the declared defaults) | 437 | de-seed — no grading change |
| Custom / no template, no engine methods at all | 83 | already follow the default (the ~6% #403 flips) |
| Custom, all-NWP | 18 | de-seed — no grading change |
| Custom, explicit DD triple (legacy account-prefs migration) | 14 | **keep** — not template-seeded |
| Custom `(ogimet_dd, dd, None)` | 1 | keep |
| Template profiles with a *genuinely* user-diverged method | **2** | keep |

**Only two profiles in all of prod have a deliberately-diverged engine method.** Everything else is either the pristine template seed or absent.

`flight_profiles.updated_at` is **not** a usable "the user chose this" signal, and it's worth saying why: it is *row*-level, so it is bumped by writes that have nothing to do with the engine methods — the `is_default` bulk clear in `api/profiles.py` (a Core `update()`, which honours `onupdate`), or any settings save that only touched cruise altitude. Empirically, **361 of the 420 pristine `vfr_only` profiles have `updated_at > created_at` while still carrying the exact fossil seed triple.** The content comparison against the template seed is the reliable signal, and it's per-field rather than per-row.

Raw `cloud_method` distribution, for reference: `dd` 857, `nwp` 426, *absent* 83, `natural_dd` 29, `soft_nwp` 12, `natural_nwp` 11, `square_nwp` 7, `square_dd` 2, `soft_dd` 1.

---

### Part A — Split `cloud_source` from cloud style

- `ProfileSettings.cloud_source: Literal['dd','nwp'] | None` replaces `cloud_method` (`src/weatherbrief/api/profiles.py:45` — whose comment already *claims* `"dd" or "nwp"`).
- `ENGINE_METHOD_DEFAULTS` (`src/weatherbrief/analysis/advisories/engine_methods.py:29`) becomes `{icing_method: 'ogimet_nwp', cloud_source: 'nwp', convective_method: 'nwp'}`.
- `_resolve_analyses` takes the source directly; **delete `_cloud_source_from_method`**. Keep a read-path fallback (`cloud_source or <derive from legacy cloud_method>`) for one release so the migration can run after deploy rather than as a flag day.
- Thread through: `advise.py`, `packs.py`, `pipeline.py`, `time_scan.py`, `time_scan_runner.py`, `preferences.py`, `models/analysis.py`.

New key name rather than reusing `cloud_method` with bare `dd`/`nwp` values: those bare values are exactly the legacy form that `user_migrations.py:46-64` (`001_method_defaults_v2`) still rewrites `dd → square_nwp` — **silently flipping the source from DD to NWP**. Reusing the key would collide with that. A fresh key makes the migration unambiguous and lets the old one be dropped clean.

### Part B — Delete the style from the profile and the settings page

- Remove the `#input-cloud-style` select (`web/settings.html:470-480`) and `parseCloudMethod` / `composeCloudMethod` (`web/ts/settings-main.ts:95-121`).
- Keep the "Cloud Source" select; it now writes `cloud_source`. Fix its hint text — *"Advisory cloud detection is independent of this setting"* is false; the source **is** what grades.
- `PREFERRED_METHOD_LAYER.clouds` (`web/ts/visualization/cross-section/layer-registry.ts:110-126`) collapses from an 8-entry composed-string map to a **composition** of source (profile) × style (`vizSettings.cloudStyle`) — mirroring iOS's `cloudLayerId(source:style:)`.
- `advisoryMethodOverrides` cloud special-case collapses to a plain assignment.

**No data migration for the style.** It has never rendered anything from the profile, so dropping it is a no-op for all 62 users — their cross-section style already comes from their own localStorage.

### Part C — Wire the cross-section to the profile it's graded on

`preferredMethods` must stop reading the account blob. Source it from the briefing's profile — better still, from the `*_method_effective` the pack already stamps per sounding (#408), which correctly reflects backend fallback and is what `advisoryMethodOverrides` already does for the advisory-highlight path.

Then **retire the account-level engine methods entirely**: drop `icing_method` / `cloud_method` / `convective_method` from `PreferencesUpdate`, `_parse_app_prefs`, and the app-prefs blob. Empty for all 534 users, never written by any client, never read by the pipeline. This kills the phantom source of truth permanently.

### Part D — De-fossilize the templates

`configs/system_profiles.json`: set all three templates to the NWP methods, then **de-seed** them (all three now equal the declared defaults, so new profiles are born sparse — the #405 goal).

Migration over the 873 fossil profiles: where `system_template_key ∈ {vfr_only, ifr_conservative}` **and** the source-normalized triple is exactly `(ogimet_dd, dd, thermo)`, delete all three keys so the profile follows the NWP default. Anything that diverges from the seed in any of the three is a deliberate override and is left alone.

---

## ⚠️ The one grading change

**Part D flips 873 profiles (61% of prod) from DD icing / DD cloud / thermo convective to NWP.** This is the largest behaviour change in all of #402 and it must be isolated:

- Its own commit, its own release note.
- Measure it before shipping, per `designs/…` / the advisory-behaviour-diff method: replay real prod packs old-vs-new with the **same** config and diff the resulting statuses. Do **not** rely on dev packs — they under-represent field coverage and overstate missing-data effects.
- Everything else in this issue is a strict no-op: Parts A and B preserve the resolved source exactly (`endswith('nwp')`), and Part C changes only which layers are *displayed*.

Note Part C and Part D interact, and in our favour: fixing the wiring (C) alone would make those 873 profiles' cross-sections visibly flip from NWP layers to DD layers — correct, but it would read as a regression. Doing D at the same time means display and grading agree on NWP, and the change is invisible to them.

---

## Ordering

1. **Deploy #403** — merged, still not live in prod.
2. **This issue.**
3. **#405** — the sparsify + template de-seed. Its "delete any value equal to the default" rule only becomes meaningful once `cloud_source` exists: it then de-seeds ~455 NWP profiles cleanly, and the 873 fossils have already been cleared here rather than being mistaken for deliberate overrides and preserved forever.

## Tests / acceptance

- Absent `cloud_source` → resolves to `nwp`; explicit `dd` honoured. Legacy `cloud_method` (all nine observed forms) still resolves to the correct source through the fallback.
- The cross-section's preferred cloud/icing/convective layers match the profile the briefing was graded on — assert directly for a DD profile (today this fails).
- Saving a profile writes no `cloud_style` anywhere; `vizSettings.cloudStyle` is the only home.
- Behaviour diff on real prod packs: the **only** status changes are the 873 fossil profiles; the 437 + 18 all-NWP de-seeds and the 62 style drops produce byte-identical advisory output.
- `GET /user/preferences` no longer returns engine methods; nothing 500s.

## Out of scope

- Advisory `enabled` flags — see #402 (the `fronts: false` trap).
- The `cloud_style` analytics dimension (`alembic/versions/071_analytics_xsection_config.py`, `analytics/events.py:127`) keeps its name — it reads `vizSettings.cloudStyle`, which survives. Keep the dimension name stable or the daily rollup series breaks.
- `PREFERRED_METHOD_LAYER` also carries an `icing: ieng` entry the backend doesn't know, and a `turbulence` group that is never passed in `preferredMethods`. Small dead ends; clean up opportunistically.

roznet:
## What & why

Three defects, one root cause: **`cloud_method` conflated a grading choice (dd/nwp source) with a render style (natural/soft/square), and the compact cross-section read engine methods from an account blob nobody writes.** This splits the two and rewires the chart to the profile it's actually graded on.

### Part A — split `cloud_source` from render style
- `ProfileSettings.cloud_source` (`Literal["dd","nwp"]`) replaces `cloud_method`.
- `ENGINE_METHOD_DEFAULTS` → `{icing_method: "ogimet_nwp", cloud_source: "nwp", convective_method: "nwp"}`.
- `_resolve_analyses` takes the source directly; **`_cloud_source_from_method` deleted**. New read-path fallback `engine_methods.cloud_source_from_settings` reduces a legacy `cloud_method` (all nine observed forms) to its source **for one release**, so the migration runs after deploy rather than as a flag day.
- Threaded through `advise.py`, `pipeline.py`, `packs.py`, `time_scan.py`, `time_scan_runner.py`, `models/analysis.py`.

### Part B — delete the style from the profile and the settings page
- Removed the dead "Cloud Style" `<select>` (it wrote a value no renderer read); the "Cloud Source" select now writes `cloud_source` and its false hint is corrected. `PREFERRED_METHOD_LAYER.clouds` collapses to a **composition** of source × `vizSettings.cloudStyle` (mirrors iOS `cloudLayerId(source:style:)`), and `advisoryMethodOverrides` collapses to a plain source assignment. Style now lives only in the client's `vizSettings.cloudStyle` (localStorage) — a no-op drop for the 62 profiles that carried a styled value.

### Part C — wire the cross-section to the graded profile
- `preferredMethods` no longer reads the account `app_prefs_json` (empty for all 534 users); it derives from the pack's per-model `primary_method_id` (the **effective** method, reflecting backend fallback) with the catalog defaults as the floor, so a **DD-graded profile now shows DD layers** in compact mode.
- Retired the account-level engine methods entirely: dropped `icing_method`/`cloud_method`/`convective_method` from `PreferencesResponse`, `PreferencesUpdate` and `_parse_service_toggles`. `GET /user/preferences` no longer returns them; nothing 500s.

### Part D — de-fossilize the templates (the one grading change)
- `configs/system_profiles.json`: dropped the engine keys from all three templates (they now equal the NWP defaults, so new profiles are born sparse).
- **`alembic/versions/076`** clears the exact DD fossil triple `(ogimet_dd, dd, thermo)` — cloud source-normalized — off `vfr_only`/`ifr_conservative` template profiles so they follow the NWP default. **Flips the 873 fossil profiles (61% of prod) from DD/DD/thermo to NWP** — the single behaviour change. A profile diverging in any axis is a deliberate override and is left untouched; all-NWP / custom / the 2 genuinely-diverged profiles are untouched (the all-NWP de-seed is #405's job and is byte-identical for grading).

## Issue linkage

Closes #410

## Testing / verification

- **Backend:** full pytest suite green (3485 passed) except one pre-existing, environment-dependent test (`test_google_login_redirects` — Google OAuth isn't configured in the sandbox; unrelated to this change). Updated Python tests to the new contract (`test_engine_method_defaults`, `test_clouds`, `test_method_provenance`, `test_ogimet_nwp_icing`, `test_reason_codes`, `test_profile_sparse_persistence`, `test_advisory_preview_api`).
- **Migration:** exercised against a seeded sqlite alembic chain — fossil `vfr_only`/`ifr_conservative` (incl. a styled `square_dd`) cleared; all-NWP `ifr_fiki` and a diverged `vfr_only` left untouched.
- **Web:** `npm run build` (esbuild) exit 0; `vitest` 499 passed / 31 files; updated `advisory-highlights`, `layer-registry`, `advisory-presets` unit tests and the `flight-lifecycle` fixtures.

### ⚠️ Grading-change measurement still owed before ship
Part D is the one grading change and the issue asks for an **advisory-behaviour diff on real prod packs** (old-vs-new, same config) to confirm the only status changes are the 873 fossil profiles. That measurement requires prod pack data and hasn't been run here — it should be done before this ships (its own release note), per the issue's ⚠️ section.

https://claude.ai/code/session_01PF9fMVJR4LrwkWQFenswDh

---
_Generated by [Claude Code](https://claude.ai/code/session_01PF9fMVJR4LrwkWQFenswDh)_


roznet:
Thanks — both Critical findings confirmed and fixed in `6eca8b0`. Also rebased the branch onto `main` (resolving the #223 conflict in `briefing-main.ts` / `advisory-highlights.ts`), so it's mergeable again.

**Critical #1 — settings page reports wrong `cloud_source` for legacy profiles.** Confirmed. Added `_resolved_response_settings` in `api/profiles.py`, applied in `_profile_to_response`, so a pre-#410 profile still carrying a legacy fused `cloud_method` (anything the migration doesn't clear — the styled values etc.) resolves its source for display instead of deserializing to `null` and showing the NWP default for a DD-graded profile. Sibling of the grading path's `cloud_source_from_settings`; the two system-template call sites (`list_system_templates`, `update_system_template`) carry no legacy key so they were already correct. Regression test added (`test_response_resolves_legacy_cloud_method_for_display`).

**Critical #2 — stale `cloud_method=options.cloud_method` at `pipeline.py:453`.** Confirmed — the earlier bulk rename matched only the 12-space-indented `run_advisories` call and missed the 16-space alt-departure call, leaving an `AttributeError` swallowed by that block's broad try/except. Fixed the kwarg; added a guard test asserting the field rename and that no `options.cloud_method` reader survives in the module (`test_briefing_options_uses_cloud_source_not_cloud_method`).

**Minor #3 (migration number)** — no change needed: the migration this PR adds is `076_defossilize_engine_methods.py` and `designs/advisories.md:400` already references `alembic 076`. The "078" in the review looks like a misread (there is no 077/078 in the tree).

**Minor #4 (dead `user_migrations 001` `cloud_method` rewrite)** — agreed it's now dead for that key; left as-is since it's harmless (account-level blob nobody reads) and touching the account-migration framework is out of scope here. Fine to prune in a follow-up.

**iOS parity note** — deferred to a `/sync-ios-web` pass; iOS already has `cloudLayerId(source:style:)` and ignores the retired engine-method JSON keys, so nothing looks broken.

Still owed before ship (unchanged, per the PR description): the advisory-behaviour diff on real prod packs to confirm the only status changes are the 873 fossil profiles — that needs prod data and hasn't been run here.

---

Issue #415, opened 2026-07-14
Extend forecast map to ECMWF's horizon (D+6) with per-day model/hour availability


roznet:
## Context

Investigating why the forecast map showed ECMWF `07-13T12Z` while GFS/ICON showed `07-14T00Z` (it was a benign write-ordering window in the standalone cycle) surfaced three things worth acting on: a day of data we already pay for and discard, a real correctness bug past 90 h, and a clean opportunity to extend the map to ECMWF's actual horizon.

## Findings

**1. D+4 is fetched, stored, and thrown away.** `MODEL_FORECAST_DAYS = 4` makes the standalone cycle sample D+0…D+4 (25 sample hours × 619 airports = the 15475 rows we write per model per cycle). But the map caps at D+3 — `cache_builder._FORECAST_DAYS = (0,1,2,3)`, `api/maps.py` `day: int = Query(..., le=3)` in three places, `maps-main.ts` `'fc.day': values: [0,1,2,3]`, and four hardcoded `data-day` buttons in `maps.html`. Surfacing D+4 costs **zero** cycle time.

**2. ECMWF precipitation is wrong past step 90 h — today, in production.** `fetch_ecmwf_grib_snapshots` step-diffs accumulated fields against `step_h - 1`. Past 90 h the delivered cadence is 3-hourly, so `step_h - 1` is never on disk; `_decode_a1` returns the `empty` sentinel, `prev_pp` is `None`, and the guard `if cur_pp is not None and prev_pp is not None` silently skips the diff. `precipitation_mm` / `snowfall_cm` are therefore left **accumulated since init**. `scoring.py:204` uses these as a *presence* test (`(precipitation_mm or 0) > 0`), so every D+4 sample scores as "precip present" if it rained anywhere earlier in the run. This already pollutes the bias leaderboard.

**3. The model horizons land almost exactly on day boundaries**, which makes a tiered map natural rather than hacky.

ECMWF delivery (00Z/12Z runs, from `delivery_config.json`, verified against files on disk): hourly 0–90 h, 3-hourly 93–144 h, 6-hourly 150/156/162/168, **hard stop at 168 h**. Both `a1` (surface) and `a2` (pressure levels) run the full 168 h, so soundings survive to the wall.

With `step = 24·day + hour` and sample hours `[6,9,12,15,18]`:

| Day | Steps (00Z run) | GFS | ICON | ECMWF |
|---|---|---|---|---|
[...]

PR #417, opened 2026-07-14
feat(maps): extend forecast map to ECMWF's horizon (D+6) with a per-day model/hour grid

roznet:
Extends the forecast map from D+3 to **D+6** — ECMWF's real 168 h wall — and makes the UI describe the data rather than assume it.

Addresses #415

## Why D+6, and why the grid isn't rectangular

ECMWF delivers hourly steps to 90 h, 3-hourly to 144 h, then 6-hourly at 150/156/162/168, and stops dead at 168 h. With `step = 24·day + hour` the model horizons land almost exactly on day boundaries:

| Day | Steps (00Z run) | GFS | ICON | ECMWF |
|---|---|---|---|---|
| D+0…D+3 | 6 → 90 | ✓ | ✓ | ✓ hourly |
| D+4 | 102 → 114 | ✓ | ✓ (≤120 h) | ✓ 3-hourly |
| D+5 | 126 → 138 | ✓ | ✗ >120 h | ✓ 3-hourly |
| D+6 | 150 → 162 | ✓ | ✗ | ✓ 06/12/18Z only |
| D+7 | 174+ | ✓ | ✗ | ✗ past the wall |

- **ICON drops out on a clean boundary.** Its cloud-diag GRIB stops at 120 h; D+4's last slot is 114 h and D+5's first is 126 h. So D+5/D+6 are two-model days — ICON simply isn't fetched past its horizon, rather than being stored with an empty ceiling column that would read as agreement.
- **D+6 is sampled at three hours, not five.** Past 144 h, 06/12/18Z land exactly on ECMWF's 6-hourly steps and 09Z/15Z never do. Five slots there would leave two permanently ECMWF-less.
- **GFS reaches 384 h but does not set the limit.** A day only GFS can reach is a day with nothing to cross-check it — the opposite of what the map is for. D+7 is not offered.

## Cost

D+4 was **already being fetched and discarded** (the cycle sampled D+0…D+4; the map capped at D+3), so it is free. Measured against today's cycle (~64 min): D+5 ≈ +8 min, D+6 ≈ +4 min → **~77 min**. Also stops ~45 DWD requests per cycle that could only 404 past 120 h.

## Dynamic UI

`/forecast/days` now reports, per day, which **hours** and which **models** actually have data. The day and hour pickers are rendered from that instead of from hardcoded buttons, and a model that can't reach the selected day is disabled with an explanation. The rules live in one place (`tasks/forecast_grid.py`) instead of being restated across the cycle, the cache builder, the API, and two frontend files.

## Correctness fix (live bug)

ECMWF `tp`/`sf` arrive accumulated since init and were step-diffed against `step - 1` — which only exists inside the hourly region. Past 90 h the lookup missed, `_decode_a1` returned the empty sentinel, and the guard `if cur_pp is not None and prev_pp is not None` silently skipped the diff, leaving the value **accumulated since init**. `scoring.py:204` uses precipitation as a *presence* test, so every D+4 sample scored as "raining" if it had rained anywhere earlier in the run — this has been polluting the bias leaderboard.

Now diffed against the previous *delivered* step, with the window it covers recorded per row (`precip_period_h`, migration 076) so a 3 h or 6 h total is never silently compared against a 1 h one.

## Tests

- `test_forecast_grid.py` — every offered slot is checked against ECMWF's real delivered step list; asserts D+7 falls off the wall, ICON stops inside its GRIB horizon, and no day is ever left with a single model.
- `test_standalone_ecmwf_precip.py` — the regression itself: with 1 mm/h accumulating, no sample may report the running total. Covers the 1 h / 3 h / 6 h cadences.
- `test_standalone_om_grid.py` — GFS/ICON honour the same per-day grid and the horizon is enforced in code, not merely requested of Open-Meteo.

Full suite: 3516 passed.

## Follow-ups (not here)

- Scoring is still window-unaware — it reads a 3 h/6 h ECMWF total as if hourly. The window is now on the row for that fix to use.
- Keeping three models out to D+6 would mean sourcing cloud-diag from ICON *global* rather than ICON-EU. New fetcher code, not a config change.

https://claude.ai/code/session_01HMCCPjiw7wgqVPTW8HpUFF


roznet:
Thanks — the `/forecast/days` finding was a real one, and better than "the mock is stale".

**Fixed (pushed):**

- **The broken spec, and the hole behind it.** You were right that the generic `**/api/maps/forecast*` route swallows `/forecast/days`. But the reason that took the *page* down rather than just the assertion is the part worth fixing: `refreshDayAvailability()` reads `.filter` off the destructured `days` outside the `try`, so a 200 with an unexpected shape throws past the handler and aborts init — and since this PR removed the static day buttons from markup, you get a page with no pickers and no map instead of a degraded one. A malformed 200 is now treated like a failed request and falls back to the same offline grid the catch branch already had. There is a spec test for it that fails without the guard (verified by reverting).
- **Off-grid hour in a shared link.** `?fc.day=6&fc.hour=9` now syncs the snapped hour to the address bar on the hydration path, not just on click.
- **Dead code removed.** `GET /maps/forecast/hours`, `fetchAvailableHours()`, `TestAvailableHoursEndpoint`, and the spec mock. Confirmed no remaining callers anywhere including iOS.
- **`designs/forecast-page.md`** updated — it still described a rectangular D-0..D-3 grid and the old hours endpoint.

Also added three spec tests for the behaviour this PR is actually about: the day picker rendering out to D-6, the far day offering three hours instead of five, and ICON being disabled at D-6 rather than silently absent.

Verification: 12/12 Playwright, 3515 backend tests, tsc clean.

**Not changed:** nothing deferred — every finding in the review is addressed in this push.

roznet:
Round 2 finding fixed — and chasing it turned up a real regression of my own that the review missed.

**`designs/metar-taf-accuracy.md`** — fixed. The "Horizon" row and intro both said "up to 4 days"; now per-model from `MAP_FORECAST_DAYS` (gfs 6, icon 4, ecmwf 6).

**The one underneath it.** That doc also documents the #236 parse-time memory optimisation: `hour_filter` is passed to `OpenMeteoClient.fetch_multi_point` as a *memory* bound, not a correctness filter — Open-Meteo returns every hour on the wire regardless, and the filter decides how much gets materialised as Pydantic objects. Checking whether that line was still true showed it was not:

Making `sample_hours=None` mean "the per-day grid" left `hour_filter` as **None on exactly the path the standalone cycle takes**, silently disabling the bound — in the same change that raised the horizon from 4 days to 6, so ~50% more response to parse, not less. Correct output, quietly expensive: ~80% of the response built only to be discarded, held alive across each chunk's sounding analysis in several threads. That is the term #236 identified as dominating the subprocess's transient memory.

Fixed by splitting the two concepts: `explicit_hours` carries the flat-list semantics, `hour_filter` stays a real set (the superset of hours any day can offer). The per-day grid is a subset of that superset, so bounding at the superset discards nothing the loop wants. Amusingly `all_sample_hours()` — which I deleted as dead code last round on your suggestion — is precisely what expresses this, so it is back with a stated purpose. Pinned by a test that fails if the bound is dropped again (verified by reintroducing the regression).

3517 backend tests pass. Nothing deferred.

Issue #420, opened 2026-07-14
iOS: forecast map (iPad + iPhone)


roznet:
Port the pan-European forecast overview map (`maps.html`, forecast tab) to the iOS/iPad app. Full design: `designs/future/ios-forecast-map.md`.

**Blocked by #419** — the client reads its colours *and* its majority consensus from what that issue builds. It cannot start first.

**Lands as two PRs** (issue = unit of tracking, PR = unit of review).

### Why this map isn't Windy

Every metric on it is a pilot's question already answered, where Windy makes you answer it yourself from several layers: **flight category** in the future (vs. combining a cloud layer and a visibility layer in your head), **crosswind on the best runway** (vs. wind speed + direction + recalling runway headings), an **aviation ceiling** (vs. squinting at cloud-cover %), and the **cross-model agreement ring** (vs. opening three models in three tabs). That framing should be visible in the UI — the metric picker is a list of questions ("Can I get in?", "Can I land?"), not a list of variable names.

Primary use case: *"where can I fly this weekend"* (discovery, no flight planned). Secondary: geographical context around a flight already being planned.

### PR 1 — the map

- `FlightListView`'s `selectedFlight` becomes a `SidebarSelection` enum (`.forecastMap | .flight(id)`); toolbar gains a Map button between `+` and `⋯`. **iPad from the start**: detail pane on regular width, `fullScreenCover` on compact (retrofitting the sidebar later is expensive).
- `MKMapView` via `UIViewRepresentable` — 619 SwiftUI `Annotation`s is a real jank risk and culling doesn't help at Europe zoom. **No clustering**, rejected on semantics: for "where can I fly" a cluster should show the *best* airport in it, for "what's dangerous" the *worst*. Rather than pick wrong, show every airport.
- Markers coloured from the **served** catalog (#419). Per-active-metric **agreement ring** as border — `weather-map.ts:304` calls `getAgreementForMetric(consensus, metric)`, so it means "the models disagree about the thing you're currently looking at". Category colour stays the primary read.
- Day/hour pickers **drawn from `/maps/forecast/days`**, never hardcoded — the grid is ragged by design (D+0…D+6; hours `06/09/12/15/18Z` except D+6 which is `06/12/18Z`; ICON absent on D+5/D+6). Port the "greyed but still tappable, explains itself" affordance: a model that can't reach the selected day must never read as agreement.
[...]

Issue #419, opened 2026-07-14
Forecast map: move the judgement server-side (consensus, colour thresholds, origin, day labels)


roznet:
Prerequisite for the iOS forecast map (see `designs/future/ios-forecast-map.md`). Moves the forecast map's *judgement* out of the web bundle and into the server, so the iOS client can be pure rendering and the two clients cannot silently drift.

**Acceptance criterion for the whole issue: the web forecast map behaves and looks exactly as it does today.** No iOS client is involved; this is deployable and verifiable on web alone.

### 1. Bake `consensus_majority` into the forecast payload

The server bakes only the **worst** consensus today; the web recomputes worst *and* majority client-side (`web/ts/visualization/weather-map-consensus.ts::computeConsensus`). Re-implementing that in Swift would mean hand-copying: ordinal max for categories, per-field worst-is-min-or-max rules, median-within-the-winning-category-pool for majority numerics, and circular mean for wind direction — all kept behaviourally identical to `analysis/airport_consensus.py`.

- Bake a `consensus_majority` block next to `consensus` in the `forecast_map:{day}:{hour}` cache entry.
- **Retire `computeConsensus` from the web** in the same PR — otherwise we've added a third implementation instead of removing one.

### 2. Map-metrics catalog: colours + thresholds as served data

`web/ts/visualization/weather-map-format.ts` hardcodes every colour ramp and threshold (`CAT_COLORS`, `RISK_COLORS`, `AGREEMENT_COLORS`, the wind/crosswind/headwind/ceiling/visibility/CAPE/cloud bands, metric labels, legends). The server has **zero** colour knowledge. Duplicating this in Swift means a threshold change needs a TS edit + a Swift edit + an App Store release.

These bands aren't really rendering anyway — "25 kt of crosswind is dark red" is a judgement about flying, same family as "ceiling below 1000 ft is IFR", which already lives server-side.

Follow the existing `api/help.py` pattern (one JSON file, imported by the web bundle *and* served ETag-cacheable to iOS, which already consumes it via `HelpCatalogResponse.swift`):

- Extract to `web/ts/data/map-metrics-catalog.json` — bands, hexes, labels, sub-labels, legend definitions.
- `weather-map-format.ts` **imports** it, so the web keeps one copy rather than gaining a second.
- Serve it (fold into `/api/help` or a sibling `/api/maps/metrics`).

### 3. Frequent airports endpoint

[...]

PR #423, opened 2026-07-15
feat(ios): forecast map — pan-European overview + airport card (#420)

roznet:
Port the web forecast map (maps.html) to the iOS/iPad app as a
pilot-framed weather map. Builds on the server-side judgement moved in
#419/PR #422: the client reads the served map-metrics catalog and the
baked worst/majority consensus, so it carries zero colour or consensus
logic and the two clients cannot drift.

The map (PR-1 scope):
- DTOs ForecastMapResponse / ForecastDaysResponse / FrequentAirportsResponse.
  ForecastMapResponse decodes with a plain decoder + explicit snake_case keys
  so `.convertFromSnakeCase` never rewrites the `agreement`/`models` dict keys.
- MapMetricsCatalog: Swift interpreter of the served catalog (categorical /
  threshold_asc / threshold_desc + m_to_sm / gray_ramp / alternate_needed,
  per-metric agreement key). HelpCatalogResponse widened with the `maps`
  section; bundled Resources/map-metrics-catalog.json baseline.
- Repository: forecastMap / forecastDays / frequentAirports (online-only).
- ForecastMapViewModel: day/hour/metric/model-mode/selection, (day,hour) LRU
  + adjacent-hour prefetch, cold-open centring from frequent airports.
- ForecastMapKitView: MKMapView via UIViewRepresentable, no clustering, 619
  markers with displayPriority=.required, zoom-scaled radius, per-active-metric
  agreement ring; metric/model switches are a pure recolour (no refetch/rebuild).
- FlightListView: selectedFlight → SidebarSelection enum + a Map toolbar button
  (iPad detail pane / iPhone fullScreenCover); day/hour/metric/model controls
  drawn from /maps/forecast/days (never hardcoded, greyed-but-tappable models).

The card + deep links (PR-2 scope):
- ForecastAirportCard: verdict header, alternate row, metric×model matrix,
  init-times footnote; row-tap switches the map's colour metric; duplicate hour
  stepper in the header. iPhone bottom sheet keeps the map pannable +
  live-recolouring via presentationBackgroundInteraction; iPad uses .inspector.
- Universal links: /maps.html → PendingNavigation.forecastMap(MapDeepLink) with
  the web's fc.* share state; /maps.html added to the AASA paths.

Tests: ForecastMapTests covers catalog colour evaluation (bands, m_to_sm,
gray ramp, null_color, categorical, alt-required aggregation), the agreement
key, consensus-block selection, model-mode tokens, and /maps.html routing.

Claude-Session: https://claude.ai/code/session_01M2TC1wErZ1QTDyeyM2M88U

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -26,6 +26,7 @@ debug
 .claude/worktrees
 .claude/projects
 .claude/scheduled_tasks.lock
+.claude/settings.local.json
 
 # Xcode
 xcuserdata/
--- a/app/flyfun-weather/flyfun-weather/App/AppState.swift
+++ b/app/flyfun-weather/flyfun-weather/App/AppState.swift
@@ -375,20 +375,42 @@ final class AppState {
 
     /// Parse an inbound Universal Link into a navigation target, if it names one.
     ///
-    /// Recognizes only `https://weather.flyfun.aero/briefing.html?flight=<id>` —
-    /// the same URL the web Smart App Banner carries as its `app-argument`. The
-    /// auth callback (`/auth/callback`), unknown hosts/paths, and a missing or
-    /// empty `flight` param all return nil so the caller routes them elsewhere.
-    /// Pure + `nonisolated` so the path/param logic is unit-testable off the
-    /// MainActor and can't silently regress.
+    /// Recognizes `https://weather.flyfun.aero/briefing.html?flight=<id>` (the
+    /// Smart App Banner's `app-argument`) and `https://weather.flyfun.aero/maps.html`
+    /// with the web's `fc.*` share state (#420) — a desktop-shared map link opens
+    /// the phone on the same day/hour/metric/airport. The auth callback
+    /// (`/auth/callback`) and unknown hosts/paths return nil so the caller routes
+    /// them elsewhere. Pure + `nonisolated` so the path/param logic is
+    /// unit-testable off the MainActor and can't silently regress.
     nonisolated static func navigationTarget(for url: URL) -> PendingNavigation? {
         guard let comps = URLComponents(url: url, resolvingAgainstBaseURL: false),
-              comps.host == "weather.flyfun.aero",
-              comps.path == "/briefing.html",
-              let flight = comps.queryItems?.first(where: { $0.name == "flight" })?.value,
-              !flight.isEmpty
-        else { return nil }
-        return .briefing(flightId: flight)
+              comps.host == "weather.flyfun.aero" else { return nil }
+        switch comps.path {
+        case "/briefing.html":
+            guard let flight = comps.queryItems?.first(where: { $0.name == "flight" })?.value,
+                  !flight.isEmpty else { return nil }
+            return .briefing(flightId: flight)
+        case "/maps.html":
+            return .forecastMap(mapDeepLink(from: comps.queryItems ?? []))
+        default:
+            return nil
+        }
+    }
+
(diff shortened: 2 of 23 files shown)


roznet:
Thanks — fixed the gust omission in `ee0a9f6`.

The card's `compactCell` now renders gusts, matching `airport-summary-card.ts`:
- wind → `dir/speedGgust` (e.g. `270/15G22`)
- crosswind / headwind → `value (gust)` (e.g. `12 (18)`)

Threaded `wind_gust_kt` / `gust_crosswind_kt` / `gust_headwind_kt` through `ForecastModelEntry.numericField` so the previously-dead decoded fields are read. The consensus block carries no gust field, so consensus-column cells stay steady-only (as on web). Added a `compactCell` parity test.

No other findings to address from this round. Note: the change is verified by reading + a unit test, but not compiled — there's no Xcode toolchain in this environment.

---

roznet:
Addressed the round-2 findings in `cc485a1` (two reviews arrived; the one still listing the gust gap had reviewed the pre-`ee0a9f6` commit — that's already fixed).

**Critical**
- **Zero markers on cold open** — annotation rebuild was gated on the `(day,hour)` string; the first `updateUIView` ran with `payload == nil` and consumed the guard. Now gated on a `payloadRevision` counter bumped on every payload assignment, so nil→populated and every slot swap repopulate; metric/model switches stay a pure recolour.

**Important**
- **LRU leak** in `store()` — the on-screen slot was dropped from `lruOrder` but kept in `slots` and never re-queued. Now re-appended.
- **Model rows `.disabled`** — made unreachable models tappable with a "no data this day" explanation (the design's "greyed but tappable, explains itself"); full-disable stays only on the day picker.
- **Re-entrant deep link dropped** — keyed the map view's `.id` to an open-token bumped on every `openForecastMap`, so a new link while the map is open re-creates the VM and applies the new state. Also now centres on the deep-linked airport.

**Minor** (batched since I was pushing): `consensus_majority` decodes with a graceful fallback to worst; literal-0 gust omitted (web parity); `isLoading` wired to a subtle slot-switch spinner; cold-open recentre guarded behind a gesture-detected user-interaction flag; dropped the duplicate `MapDeepLink`/`DeepLink` pair; guard-let the annotation dequeue; added an in-flight `Set` to avoid double-fetch; added a VM deep-link-at-init test.

Still verified by reading + unit tests, not compiled (no Xcode here) — flagging that the marker/gesture/`.id` behaviours are the kind best confirmed on-device.

---

roznet:
Fixed in `72d4c01`.

**Important** — `openForecastMap` no longer bumps `mapOpenToken` on every call. It bumps only when a *new* deep link arrives while the map is already open (`deepLink != nil && (selection == .forecastMap || showMapCover)`), so the re-entrant-deep-link fix is preserved but a plain toolbar re-open keeps the VM and its `(day,hour)` LRU cache instead of tearing down and re-fetching. Good catch — that was a regression from the round-2 `.id` fix.

**Minor** — expanded the comment on the gesture-based user-interaction detection, spelling out that it reads MapKit's private view hierarchy, both failure modes (silent stop → cold-open yank returns; false-positive → cold-open never applies), and the hardened alternatives, so a future maintainer has the context. Left the mechanism itself in place (a rewrite is riskier to land unverified than the documented trick).

I'm treating this as the last fix round on my side — the prior two rounds' correctness items are all confirmed landed, and what remains is well-understood and documented. Recommend merge once CI is green.

---

PR #425, opened 2026-07-15
Airport forecast overlay on the briefing route map (#424)

### LINKED ISSUE #424: Briefing map: airport forecast overlay at flight time
## Idea

On the briefing **map view**, overlay the per-airport forecast markers (the same ones the full forecast map shows) for the forecast time closest to the flight. This gives a geographic, at-a-glance picture of conditions across the region around the flight — without leaving the briefing.

## Controls (right-aligned on the existing map controls row)

- **On/off toggle** — show/hide the airport overlay. Default **on**.
- **Forecast-time label** — the actual snapshot valid-time used (may differ slightly from the flight time, since it snaps to the nearest available sample hour).
- **Metric dropdown** — category / wind / ceiling / visibility / CAPE / … (the forecast-map `FORECAST_METRICS` set). Independent of the route-segment color metric.
- **Open full forecast map** button — deep-links to `maps.html` seeded with the same day / hour / model / metric.

## Design decisions (finalized)

- **Airports:** show all ~620 watchlist airports. The point is a geographic picture; map zoom does the filtering.
- **Model:** the overlay follows the briefing map's existing **model selector** (individual model — gfs/icon/ecmwf). No worst/majority consensus here; that stays on the full forecast map (reachable via the button). Model keys already match between the two systems, so this is a pass-through.
- **Availability:** forecast snapshots only exist D-0…D-6. If the flight is **> 7 days out** (or the target day has no data), the overlay is simply not offered — toggle hidden, nothing drawn.
- **Legend:** reuse the forecast map's small over-the-map legend (`MAP_LEGENDS`), positioned so it doesn't collide with the route-segment legend.
- **Deferred:** when the flight time is effectively *now*, offer switching the overlay to current METAR instead of forecast — deferred to a later change; the `forecast_time` seam makes it a data-source swap, not a rewrite.

## Reuse (no backend changes)

- `weather-map-format.ts` — `getForecastColor`, `getConsensus`, `formatMetricValue`, `METRIC_LABEL`, `MAP_LEGENDS`, `FORECAST_METRICS` (catalog-driven, so overlay colors can't drift from the full map).
- `maps-adapter.ts` — `fetchForecastMap(day, hour)` + `fetchAvailableDays()`; the `/maps/forecast` endpoint is already built and cached.
- `RouteMapRenderer` — new `airportForecastGroup` layer group mirroring the existing `frontsGroup` overlay pattern.
- `maps-main.ts` `mapsUrlState` schema for the deep-link.

## Scope

Frontend-only, ~1 day. Backend (endpoint, cache, consensus, color catalog) is unchanged.

roznet:
## What & why

Overlays the per-airport forecast markers from the full forecast map onto the **briefing route map**, for the snapshot valid-time nearest the flight — a geographic, at-a-glance picture of conditions around the flight without leaving the briefing.

**Controls** (right-aligned on the existing map row):
- **On/off toggle** — show/hide the airport overlay. Default **on**.
- **Forecast-time label** — the snapshot valid-time actually used (nearest available sample hour to departure; may differ slightly from the flight time).
- **Metric dropdown** — category / wind / ceiling / visibility / CAPE / … (`FORECAST_METRICS`). Independent of the route-segment color metric.
- **Open full forecast map** — deep-links to `maps.html` seeded with the same day / hour / model / metric.

**Design decisions:**
- All ~620 watchlist airports are drawn — zoom does the spatial filtering.
- The overlay follows the briefing's existing **model selector** (individual model gfs/icon/ecmwf). Consensus (worst/majority) stays on the full forecast map, reachable via the button. Model keys already match between the two systems, so it's a pass-through.
- Outside the D-0…D-6 forecast horizon (flight > ~1 week out, or in the past) the overlay is simply **not offered** — the control cluster is hidden and nothing is drawn.
- The overlay legend is a small floating box inside the map (top-right), distinct from the route-segment gradient legend below the map.

**Reuse / architecture (no backend changes):**
- New `airportForecastGroup` layer group on `RouteMapRenderer`, mirroring the existing `frontsGroup` overlay pattern; drawn beneath the route so segments/waypoints stay on top.
- Colors, tooltips and legend come from the forecast map's served catalog via shared `getForecastColor` / `getForecastTooltip` / `forecastLegend` (extracted/exported), so the two views can't disagree.
- Snapshot payload holds every model per airport, so switching model or metric is a pure client recolour; only a new day/hour slice hits `/maps/forecast`, cached by slice.

**Deferred** (noted in the issue): when the flight time is effectively *now*, offer switching the overlay to current METAR instead of forecast — the `forecast_time` seam makes that a later data-source swap, not a rewrite.

## Issue linkage

Closes #424

## Testing / verification

- `tsc --noEmit` — clean for all touched files (only 2 pre-existing errors remain, in `ts/eval/label-panel.ts`, untouched here).
- `esbuild` bundles for `briefing` and `maps` build successfully.
- `vitest run` — 507/507 pass (includes the map-metrics-catalog validation the shared color path relies on).
- Standalone verification of the pure logic (17 assertions): UTC day computation (incl. month-boundary and past/future horizon), nearest-sample-hour snapping, `forecast_time` formatting, and deep-link URL building (individual model passed through; consensus falls back to the full map's default).

https://claude.ai/code/session_013JvgjJhiiN1qnPU76ppxVs

---
_Generated by [Claude Code](https://claude.ai/code/session_013JvgjJhiiN1qnPU76ppxVs)_


roznet:
Thanks for the review — all five findings addressed in 83986b2:

- **Critical (missing `id`)** — the forecast valid-time span now has `id="map-forecast-time"`, so `patchForecastTimeLabel()` resolves and the label updates from "…" to the real snapshot time on first load.
- **Important (unvalidated model)** — the overlay now draws only when the briefing's selected model actually has airport data for that day (the server-advertised `models` list). UKMO / MeteoFrance / ICON-EU — and ICON on a far day where it drops out — hide the overlay instead of rendering every marker muted-gray, matching how the fronts layer clears.
- **Important (hardcoded horizon/hours)** — removed `FORECAST_MAX_DAY` / `FORECAST_SAMPLE_HOURS`; the horizon and each day's offered hours now come from `fetchAvailableDays()`, per `designs/forecast-page.md`. This fixes the D+6 coarse-hour mismatch, and the fetch/cache key now uses the server-snapped hour.
- **Important (unvalidated persisted metric)** — `mapForecastMetric` is validated against `FORECAST_METRICS` and falls back to `flight_category`, so a future catalog rename resets gracefully instead of crashing the marker loop.
- **Minor (no tests)** — extracted the pure slot/hour/url/format logic into `route-map/forecast-overlay.ts` and added `tests/unit/forecast-overlay.test.ts` (16 cases: UTC day math, hour snapping against the server grid incl. the coarse far day, horizon gating, time formatting, deep-link model guard).

Full suite green (523 passed), `tsc` clean, briefing bundle builds.

One small UX note for a possible follow-up (not in this change): when an unsupported model is selected the markers simply disappear while the toggle stays on — honest, but a "no airport data for this model" hint could read better. Happy to add it if you'd like.

---

roznet:
Round-2 findings addressed in 2830209:

**Important**
- **Double-render / hot-path perf** — added a render signature guard: the ~620-marker teardown/rebuild is skipped when nothing about the overlay changed since the last draw (signature over show/metric/model + a data-version bumped only on a real reference change). `render()` and `updateForecastOverlay()` no longer both rebuild each pass, so altitude-slider drags (which full-re-render per input tick) don't churn the overlay.
- **Unsupported model not reflected in controls** — when the selected model has no airport data for the day, the cluster now shows a "No `<model>` airport data" note instead of the time/metric, so a blank overlay never reads as "all clear" (per `forecast-page.md`).

**Minor**
- **`.map-tooltip` leak** — the marker tooltip CSS is now a briefing-only class (`.wx-forecast-tooltip`); the full forecast map's own `.map-tooltip` markers are untouched again.
- **Duplicated model list** — added a single exported `FORECAST_INDIVIDUAL_MODELS` in `maps-adapter.ts`; the overlay and the two `maps-main.ts` copies now reference it.
- **Permanent "gave up" sentinel** — both fetches now back off 30s on failure instead of caching an empty grid / null slice forever, so a transient blip recovers on a later render (and doesn't hammer the endpoint on the slider hot path).

`tsc` clean, 523 tests pass, briefing + maps bundles build.

---

PR #427, opened 2026-07-15
fix(ios): repaint flight-list rows when a briefing updates on the server

### LINKED ISSUE #426: iOS flight list rows don't repaint when a briefing updates on the server (fixed by restarting the app)
## Symptom

When a flight's briefing is updated on the server, the flight-list card in the iOS app does **not always** update its traffic-light badge / advisory chips. Force-quitting and reopening the app shows the correct, fresh state — which is the classic "the data is fine, the view didn't repaint" signature.

## Root cause — stale row (primary)

`FlightCardView` renders its status badge from `flight.latestBriefing` (assessment / outlook / advisory summary), and `/api/flights` inlines that summary into every `FlightResponse`. So a server update means: same flight `id`, **new** `latestBriefing`.

`FlightResponse` conforms to `Hashable` with **id-only** equality (`Views/Flights/FlightListView.swift`):

```swift
extension FlightResponse: Hashable {
    static func == (lhs: FlightResponse, rhs: FlightResponse) -> Bool { lhs.id == rhs.id }
    func hash(into hasher: inout Hasher) { hasher.combine(id) }
}
```

This id-only identity is **intentional and required** for stable `List` selection and `ForEach` identity (a flight whose briefing changes while selected must stay selected). But SwiftUI also uses that same `==` when diffing the `FlightCardView` row: two versions of the same flight — old stale badge vs. new RED/AMBER/GREEN — compare **equal**, so SwiftUI concludes the row's inputs are unchanged and **skips re-rendering it**. `FlightListViewModel.loadFlights()` fetches the fresh data and updates `state`, but the visible card keeps the old badge. A cold restart rebuilds the view tree from scratch, so it always looks correct afterward.

### Why it's intermittent ("not always")

The card also takes `hasCachedData` / `isOffline` / `isRefreshing`. When the active-refresh poll happens to observe the whole refresh cycle, `isRefreshing` toggles `true → false` for that row, and that *incidental* input change forces a re-render that pulls in the fresh briefing. When the poll misses the cycle (see below), nothing forces the row and it goes stale.

## Secondary cause — edge-triggered completion detection

`FlightListViewModel.pollActiveRefreshesLoop()` triggers a reload only when a flight **leaves** the `/api/refresh/active` set:

```swift
let finished = refreshingFlightIds.subtracting(newIds)
if !finished.isEmpty { await loadFlights() }
```

A refresh whose entire active lifetime falls between two 5 s polls (fast refresh, or one triggered on the web / another device / auto-refresh that finished before we observed it as active) is never seen, so no reload fires at all until the next foreground / pull-to-refresh.

## Fixes

1. **Content-based row repaint** — make `FlightCardView` conform to `Equatable` over exactly the fields it draws and render it via `.equatable()`, so a changed briefing repaints the row while `FlightResponse`'s id-only identity stays for selection. Requires `Equatable` on `BriefingStatusInfo` / `AdvisorySummary` / `AdvisoryChip` / `CoveragePending`.
2. **Level-triggered reconcile** — decode the `fetch_timestamp` the server already sends in the briefing summary, and add a periodic reconcile in the poll loop that re-fetches the list and repaints only when a flight's latest pack timestamp advanced. Robust even when a refresh cycle is missed by the active-set edge.

## Affected files

- `app/flyfun-weather/flyfun-weather/Views/Flights/FlightCardView.swift`
- `app/flyfun-weather/flyfun-weather/Views/Flights/FlightListView.swift`
- `app/flyfun-weather/flyfun-weather/Models/API/FlightResponse.swift`
- `app/flyfun-weather/flyfun-weather/ViewModels/FlightListViewModel.swift`

roznet:
## What & why

The iOS flight-list card wasn't reliably updating its traffic-light badge / advisory chips when a flight's briefing changed on the server — force-quitting and reopening the app showed the correct state. Two causes, one primary and one secondary:

**1. Stale row (primary).** The card renders from `flight.latestBriefing`, which `/api/flights` inlines into each `FlightResponse`. `FlightResponse` is `Hashable` by `id` **only** — deliberately, so `List` selection stays stable when a flight's briefing changes. But SwiftUI reuses that same id-only `==` when diffing the row, so a same-id / new-briefing update compared *equal* and SwiftUI skipped re-rendering `FlightCardView`. `loadFlights()` fetched the fresh data, but the visible card kept its old badge until the view tree was rebuilt (relaunch). It was *intermittent* because the active-refresh poll toggling `isRefreshing` sometimes forced an incidental re-render that pulled in the fresh briefing.

**2. Edge-triggered completion detection (secondary).** `pollActiveRefreshesLoop()` reloaded only when a flight *left* the `/api/refresh/active` set. A refresh whose whole active lifetime fell between two 5 s polls (fast refresh, or one started on the web / another device / by auto-refresh) was never observed, so no reload fired at all until the next foreground / pull-to-refresh.

### Fixes

1. **Content-based row repaint.** `FlightCardView` now conforms to `Equatable` over exactly the fields it draws and is rendered via `.equatable()`, so a changed briefing repaints the row while the id-only identity stays for `List` selection. Adds `Equatable` to `BriefingStatusInfo` / `AdvisorySummary` / `AdvisoryChip` / `CoveragePending`.
2. **Level-triggered reconcile.** Decode the `fetch_timestamp` the server already sends in `latest_briefing`, and add a periodic reconcile (~30 s) in the active-refresh poll that repaints only when a flight's latest pack timestamp advanced — catching the missed-between-polls case the active-set edge can't see. Quiet by design: one cheap list fetch, no spinner, no-op when nothing changed.

## Issue linkage

Closes #426

## Testing / verification

Added unit tests (Swift Testing, via the injected `MockBriefingRepository`, no network):
- `FlightCardEquatableTests` — same-id / different-briefing cards compare **unequal** (while the underlying `FlightResponse` stays id-equal), identical content is equal, and the `isRefreshing` flag breaks equality.
- `reconcileRepaintsWhenPackTimestampAdvances` / `reconcileIsNoopWhenPackUnchanged` — the reconcile repaints only on a real pack-timestamp advance.
- `BriefingStatusDecodeTests` — `fetch_timestamp` decodes onto `fetchTimestamp` via the shared snake-case decoder, and a legacy payload without it decodes as `nil`.

Note: this is an iOS/Xcode target, which can't be built or run in the web environment — the tests above are written against the existing `flyfun-weatherTests` harness but were not executed here. Worth a local `xcodebuild test` (or a simulator pass on the flight list) before merge.

https://claude.ai/code/session_01NadGSq5EEvte2RyLUvDqNk

---
_Generated by [Claude Code](https://claude.ai/code/session_01NadGSq5EEvte2RyLUvDqNk)_

### DIFF
--- a/app/flyfun-weather/flyfun-weather/Models/API/FlightResponse.swift
+++ b/app/flyfun-weather/flyfun-weather/Models/API/FlightResponse.swift
@@ -157,7 +157,8 @@ enum FlexibilityMode: String, Codable, Sendable, CaseIterable {
 /// Weather-coverage status for a flight saved beyond the forecast horizon.
 /// Present on `FlightResponse.coverage` only while no model reaches the flight
 /// date yet; the UI shows a neutral pending state instead of an assessment.
-struct CoveragePending: Codable, Sendable {
+/// `Equatable` so `FlightCardView` can diff on it (see its `==`).
+struct CoveragePending: Codable, Sendable, Equatable {
     /// ISO date (yyyy-MM-dd) — first (early-outlook) briefing appears.
     let availableDate: String
     /// ISO date — full GRIB briefing, if resolved. nil when unresolved.
@@ -201,7 +202,10 @@ struct AircraftInfo: Codable, Hashable, Sendable {
 /// each card can render a status tag from the single `/api/flights` response
 /// (mirrors the web `BriefingStatusInfo`). Scalar fields are optional so a
 /// partial/legacy payload decodes rather than failing the whole flight.
-struct BriefingStatusInfo: Codable, Sendable {
+/// `Equatable` so `FlightCardView` diffs on the briefing content (the card's
+/// only source of truth) rather than `FlightResponse`'s id-only identity, which
+/// would otherwise make a same-id/new-briefing update look unchanged (#426).
+struct BriefingStatusInfo: Codable, Sendable, Equatable {
     /// GREEN / AMBER / RED traffic-light verdict (short-range, within the GRIB
     /// horizon). nil when only a long-range `outlook` is available.
     let assessment: String?
@@ -213,10 +217,17 @@ struct BriefingStatusInfo: Codable, Sendable {
     let outlookReason: String?
     let hasAdvisories: Bool?
     let advisorySummary: AdvisorySummary?
+    /// The latest pack's fetch timestamp (ISO 8601), as the server sends it in
+    /// `latest_briefing.fetch_timestamp`. Used by the flight list's
+    /// level-triggered reconcile to detect that a briefing advanced even when
+    /// the active-refresh poll never observed the refresh (#426). Optional (no
+    /// default — a `let` with a default would be dropped from the synthesized
+    /// `Codable`, never decoding the field); absent on older servers → nil.
+    let fetchTimestamp: String?
 }
 
 /// Compact RED/AMBER advisory breakdown for the flights-list card chips.
-struct AdvisorySummary: Codable, Sendable {
+struct AdvisorySummary: Codable, Sendable, Equatable {
     let red: Int
     let amber: Int
     /// Severity-ordered named concerns, capped at 3 server-side.
@@ -224,7 +235,7 @@ struct AdvisorySummary: Codable, Sendable {
 }
 
 /// One named advisory concern for the summary chips.
-struct AdvisoryChip: Codable, Sendable {
+struct AdvisoryChip: Codable, Sendable, Equatable {
     let status: String  // "RED" | "AMBER"
     let name: String
 }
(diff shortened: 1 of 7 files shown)


roznet:
Thanks — fixed the reconcile/load race in `9a088f1`.

**What was fixed:** `reconcileLatestPacks()` now guards on a monotonic `loadGeneration`. Every authoritative commit (a completed `loadFlights()` or a reconcile write) bumps it; the reconcile captures the value before its `repository.flights()` await and drops its write if it changed. So a slow background reconcile can no longer clobber a fresher `loadFlights()` that completed during its suspension — the exact revert-to-stale scenario you described. `loadGeneration` and `state` only mutate on the main actor between the reconcile's suspension points, so the compare-and-commit is race-free.

I went with the "detect-and-drop-itself" option (over sharing `isLoading`) so a user-initiated pull-to-refresh always wins over the passive background reconcile, rather than being silently suppressed while a reconcile is in flight.

Added a regression test (`reconcileDropsStaleWriteWhenLoadFlightsWins`) that parks the reconcile mid-fetch via the `TestGate`, lets a newer `loadFlights()` commit, then releases the reconcile and asserts it returns `false` and leaves the fresh (RED) briefing in place.

Nothing deferred.

---

PR #389, opened 2026-07-12
Add model-specific advisory evidence and focus

downle:
Closes #223

## Summary

- Adds the model-specific advisory evidence, data-state, provenance, and representative-model contract while keeping legacy packs additive and safe.
- Fixes the evidence-gated missing-data, aggregation, route-cell geometry, SFIP display/weight, and cloud-overlap defects documented in the audit.
- Adds model-isolated focus overlays, emphasis, Compare/DD-NWP/airport/fronts actions, and backend-sourced method badges without recalculating meteorology in the browser.
- Preserves old-pack behaviour without guessed geometry and reuses the Skew-T linkage shipped in #309.
- This is a standalone issue #223 implementation.

## Verification

- `GOOGLE_CLIENT_ID=test-client GOOGLE_CLIENT_SECRET=test-secret venv/bin/pytest -q`
  — 3520 passed, 12 skipped, 16 deselected; 485 warnings in 116.16s.
- `cd web && npx vitest run` — 36 files, 613 passed.
- `cd web && npx tsc --noEmit` — only the two known baseline errors at `ts/eval/label-panel.ts:233` and `:242`; no new issue #223 TypeScript errors.
- `cd web && npx playwright test` — 59 passed in 25.0s.
- Focused Advisory evidence actions Playwright group — 27 passed.
- Post-fix manual acceptance — **PASS**: standalone manual acceptance
  `3/3 passed`; focused Advisory evidence actions Playwright group
  `27/27 passed`; combined `30/30 passed`; visual inspection `6/6` refreshed
  screenshots acceptable.
- `git diff --check` — clean. Tracked worktree state is clean; the pre-existing untracked `docs/superpowers/plans/CLAUDE.md` is intentionally excluded.
- The Swift Codable fixture includes `representative_model`, `data_state`, `primary_method_id`, and `evidence_regions` while leaving the existing Swift models unchanged. Neither `xcodebuild` nor `swift` is available on this Linux host, so XCTest was not run locally and remains a macOS CI check.

Manual acceptance covered model replacement without cross-model union, explicit Compare attribution, map-halo ordering, partial/legacy/unavailable states, airport-drawer close/pack-switch behaviour and focus restoration, fronts fitting, keyboard operation, reduced motion, and forced-colour usability.

## Meteorology / metrics review

[...]

PR #438, opened 2026-07-16
feat(web): ribbon-hover tooltip surfaces reason_code — why, not just how bad (#412)

### LINKED ISSUE #412: Ribbon hover tooltip — surface reason_code (why is it flagged here, not just how bad)
The last follow-up carried over from #223 (the rest of that issue's "unscheduled follow-ups" list is deliberately dropped — we won't do them).

## What

Hovering the verdict ribbon should say **why** the advisory is flagged at the cursor's x-position, not just what colour it is.

The ribbon is a gapless partition of `[0, total_nm]` — it already tells you *that* the route is amber here and red there. What it can't tell you is that two identically-coloured stretches can mean completely different things.

## Why this is worth doing now

When #223 listed this, the data didn't exist. It does now: `HighlightRegion.reason_code` (shipped in 5fe6b571) records why each region fired, and it is only emitted where the colour and the shape genuinely cannot say:

- **`icing_escape`** — a RED band is either `no_escape` (the freezing level is *below terrain*: you cannot descend out of the ice) or `no_escape_unknown` (we don't know the freezing level or the terrain). An AMBER band is either `tight_margin` (there is an escape, but it's marginal) or `warm_escape` (there's above-freezing air below — you have an out).
- **`fiki_icing`** — a RED band is `sld` (**your FIKI certification does not cover SLD** — not a severity call, a legality wall), `severe_icing` (beyond capability), or `thick_transit` (you *can* transit it; the exposure is just long). AMBER splits `icing_at_cruise` (ice at your cruise level — you'd be *loitering* in it, the one thing FIKI cannot do) from `transit_exposure` (thickness on the climb/descent only).
- **`convective`** — `active_track` (the model's own convective scheme sees the storm) vs `thermo_floor` (the model is quiet, but the thermodynamics floored it up). On real packs `thermo_floor` fires 9 RED and 129 AMBER — those are precisely the "models under-fire" cases.

Today all of these render as an identical coloured stretch of ribbon. The pilot is told *how bad*, never *why*, and in several of these pairs the correct action is different.

Every other evaluator deliberately emits `reason_code = None` — their reason IS the advisory's definition, or it's already carried by `kind` / `severity`. See the note on `HighlightRegion` in `models/advisories.py`. **Do not "complete" them for tidiness** — a reason that restates the definition is noise, and the tooltip should simply show nothing for those.

## Shape

- Hover the ribbon → tooltip at the cursor's x: advisory + verdict +, when present, the human phrasing of `reason_code`.
- Needs a localized label per code (en/de/es/fr), the same way method ids would have been labelled. Unknown/absent code → show just the verdict, no invented text.
- The regions already carry `metric_id` too (`cloud_cover`, `icing`, `convective_risk`, `cat_risk`, …), so the tooltip could also offer a jump to the layer that explains the region — optional, and probably a second pass.

## Prior art

PR #389 (closed, unmerged) built a ribbon-adjacent hover treatment; worth a look for the interaction, not the data model — ours is different and better suited (distance-space regions with provenance).

roznet:
## What & why

Hovering the verdict ribbon in the cross-section now says **why** the advisory is flagged at the cursor's x, not just what colour it is. The ribbon is a gapless partition of `[0, total_nm]` — it already tells you *that* the route is amber here and red there, but two identically-coloured stretches can mean completely different things. The data to tell them apart (`HighlightRegion.reason_code`, shipped in 5fe6b571) already reaches the client; it just had no consumer.

The tooltip shows **advisory name + verdict + — when the flagged region under the cursor carries one — the human phrasing of `reason_code`**:

- **`icing_escape`** — `no_escape` / `no_escape_unknown` (RED: can't descend out of the ice) vs `tight_margin` / `warm_escape` (AMBER: an escape exists).
- **`fiki_icing`** — `sld` (outside FIKI certification — a legality wall) / `severe_icing` / `thick_transit` vs `icing_at_cruise` (loitering in ice) / `transit_exposure` (climb/descent only).
- **`convective`** — `active_track` (the model's own scheme sees the storm) vs `thermo_floor` (model quiet, thermodynamics floored it up).

Every other evaluator deliberately emits `reason_code = None` — their reason IS the advisory's definition — so the tooltip shows just the verdict there. An unknown/absent code → verdict only, never invented text.

**Design:** geometry-free lookups (`ribbonSeverityAt`, `reasonCodeAt`, `reasonLabelKey`) in `advisory-highlights.ts` — the ribbon owns the verdict at a distance, the flagged regions own the reason. The hover itself is handled inline in `interaction.ts` (distance-space, like `current-conditions`), gated on the ribbon actually being drawn. Per-code localized labels (`advisories.reason.<code>`) added in en/de/es/fr, plus verdict labels; aviation abbreviations (SLD, FIKI) left untranslated. The advisory display name is plumbed to the cross-section via a new optional `VizRouteData.advisoryHighlightName`, set alongside `advisoryHighlights`.

**Scope:** web-only for now (mouse hover). An iOS tap-based equivalent and the optional `metric_id` layer-jump (issue's "second pass") are follow-ups. Prior art PR #389 was reviewed for the interaction; the data model here is different (distance-space regions with provenance).

## Issue linkage

Closes #412

## Testing / verification

- New unit tests for the pure lookups + label resolution (boundary handling, overlap most-severe-wins, no-code regions, unknown/absent codes → null).
- Full web suite green: **546 passed**.
- `tsc --noEmit` clean for all touched files (only 2 pre-existing errors remain, in unrelated `ts/eval/label-panel.ts`).
- `npm run build:briefing` builds clean.

---
_Generated by [Claude Code](https://claude.ai/code/session_01JWMsxHPH41BYAyschqwnsF)_


roznet:
Thanks — the boundary finding was a real one. Fixed in d25ba9b.

**What changed:** `reasonCodeAt` now takes an optional `severity` and only considers regions of that severity; `showRibbonTooltip` passes the already-resolved ribbon verdict in. So at a touching RED/AMBER boundary (`no_escape` run meeting a `tight_margin` run) the reason always describes the verdict actually displayed — the boundary point resolves to `amber` on the ribbon and to `tight_margin` for the reason, never the contradictory `Amber · <RED reason>`. When no matching-severity region sits under the cursor, the tooltip shows the verdict alone rather than a mismatched reason.

Added a regression test for the touching-boundary case (both interiors + the shared boundary + the no-matching-severity fallback), which the previous "prefers most severe" overlap test didn't cover. Full web suite green (547), typecheck + `build:briefing` clean.

The other observations (i18n key parity, the `advisoryHighlightName` fallback being defensive-only, compare-mode scope, hit-band vs `DISTANCE_LABEL_DY`) match my intent — no changes there.

---

PR #439, opened 2026-07-16
ICON-EU: decode rain_con so the convective firing gate can evaluate ICON towers

### LINKED ISSUE #421: ICON-EU: decode rain_con so the convective firing gate can evaluate ICON towers
## Problem

The NWP convective track (#283) grades on the model's own convective scheme. Two of its safeguards need a **realization signal** — evidence that the scheme actually fired, not just that it drew a tower:

- **Firing gate** (`_apply_firing_gate`, `analysis/sounding/convective.py:664`) — a MODERATE+ tower is held down one level when the scheme is demonstrably dry (`convective_precip_mm_h <= 0.1` or `convective_cover_pct <= 15`).
- **Native corroboration** (`_CORROB_PRECIP_MM_H = 0.5`) — a realized MODERATE+ cell with strong native precip/indices is bumped up one level.

ICON-EU can be evaluated by **neither**. It exposes no `convective_cover_pct` (DWD has no convective-cloud-cover field) and we do not decode any convective precip. The gate is deliberately missing-data-safe (we only hold down on *positive* evidence of no firing), so **every ICON tower keeps its full tower-top tier, forever** — a deep-but-dry `htop_con` (the capped / elevated case) reads at full native risk with nothing able to contradict it. GFS and ECMWF both get held down here; ICON does not. This is recorded as a known calibration gap in `designs/convective-analysis.md` ("Known gap (ICON-EU)") and in the code comment at `fetch/grib/icon_eu_fetch.py:112`.

The fix is to decode DWD's `RAIN_CON`.

## What DWD publishes

`RAIN_CON` is available on ICON-EU open data as a single-level field at every step we already fetch (verified live against `https://opendata.dwd.de/weather/nwp/icon-eu/grib/<cycle>/rain_con/`; steps are hourly 0–78h then 3-hourly to 120h, matching `_snap_to_icon_eu_grid`). A single file at +006 is **~250 KB** compressed — for comparison `CAPE_ML`, which we already fetch, is ~730 KB. Cost is negligible next to the model-level sounding fetch (levels 35–74 per variable).

### Semantics vs ECMWF `cp`

`rain_con` is the near-equivalent of ECMWF's `cp`, with two differences that matter for the implementation:

| | ECMWF `cp` | ICON `rain_con` |
|---|---|---|
| Accumulation | since init | since init — **same**, needs step-differencing |
| Units | m water equivalent (`× 1000` → mm) | kg/m² ≡ **mm already — no ×1000** |
| Phase coverage | total convective precip (liquid + frozen) | **convective rain only**; convective snow is a separate `SNOW_CON` field |

So the de-accumulation logic is a direct port of the ECMWF path, but the unit conversion is *not*, and `rain_con` alone under-reads a cold-air-mass convective snow shower. See "Scope decision: `snow_con`" below.

### 1. Fetch the variable — `fetch/grib/icon_eu_fetch.py`

Add `rain_con` to `ICON_EU_CLOUD_DIAG_VARIABLES` (line ~115) and delete the now-obsolete comment above it explaining why it was deferred.

### 2. Fetch the predecessor step

This is the only genuinely new bit of design. ECMWF gets de-accumulation nearly free: the whole run is already on local disk (rsync'd from ECPDS), so `_enrich_ecmwf_grib` walks *every* a1 step in the flight window **±3h** (`ECMWF_FLIGHT_WINDOW_MARGIN`) in step order and carries `prev_cp_per_point` / `prev_a1_valid_utc` across the loop (`fetch/grib/__init__.py:2145-2214`). Because of the 3h margin, the step preceding the flight window proper is normally already in the loop, so in-window steps get a real rate.

ICON has no such margin — we download exactly the flight-window hours from DWD on demand, so **the first window hour has no predecessor to difference against**.

Proposed approach — add a leading step to the *single-level* fetch only (the model-level sounding fetch must not grow):

- Add a helper `icon_eu_previous_step(fhour: int) -> int | None` in `icon_eu_fetch.py`, respecting the grid (`fhour - 1` at or below 78h, `fhour - 3` above; `None` for `fhour == 0`, where accumulation is 0 by definition and no rate exists).
- Extend the forecast-hour list used for the cloud-diag fetch with `icon_eu_previous_step(min(forecast_hours))`, keeping the model-level list untouched.
- Iterate the cloud-diag loop in **sorted step order**, carrying `prev_rain_con_per_point: list[float | None]` and `prev_valid_utc`, exactly mirroring the ECMWF loop.
- Applying the leading step's diagnostics is a harmless no-op — `_matches_valid_time` will not match any `hourly` row outside the flight window — so it needs no special-casing beyond the state carry.

Both fetch paths need the extra step: `_prefetch_icon_eu_data_inner` (`fetch/grib/__init__.py:2951-2963`, `_fetch_diag`) and the on-demand fetch inside the enrichment loop (`fetch/grib/__init__.py:3143-3157`).

### 3. ⚠️ Cache-key invalidation (easy to miss)

The ICON single-level blob is cached under one key for all variables: `cache_key(fhour, "ICON_EU_CLOUD_DIAG")`. Adding `rain_con` changes the *content* of that blob, but **not the key** — so any already-cached run would decode without `rain_con` and silently keep the current (gate-less) behaviour, with no error. Bump the key (e.g. `"ICON_EU_CLOUD_DIAG_V2"`) so existing cached blobs are re-fetched.

### 4. Decode — `fetch/grib/decode.py`

Add to `_ICON_CLOUD_DIAG_FIELD_MAP` (line ~115):

```python
"rain_con": "conv_rain_kg_m2",  # kg/m² ≡ mm, ACCUMULATED since init
```

Leave `build_icon_cloud_diagnostics` (line 1285) unaware of the rate — like ECMWF, the raw accumulated value comes out of the decode and the *rate* is computed in the enrichment loop, which is the only place that knows the previous step.

### 5. De-accumulate and inject — `fetch/grib/__init__.py`

In the ICON cloud-diag enrichment loop, mirror `fetch/grib/__init__.py:2154-2183`:

```python
rate_mm_h = max(0.0, (rain_con - prev_rain_con) / window_h)   # NOT × 1000 — already mm
diagnostics[i] = diag_i.model_copy(
    update={"convective_precip_mm_h": rate_mm_h}
)
```

Compute `window_h` from the two valid times rather than assuming 1h — ICON drops to 3-hourly past +78h, just as ECMWF does past +90h. Rebuild via `model_copy` rather than mutating, per the `NWPCloudDiagnostics` immutability contract (review #284).

Once it lands on the diagnostics object it rides the existing forward-fill (`fill.py:_fill_cloud_diagnostics`) and spatial-interpolation (`spatial_interpolation.py:228`) paths for free — both already carry `convective_precip_mm_h`.

### 6. Nothing to change in analysis

`NWPCloudDiagnostics.convective_precip_mm_h` already exists (`models/analysis.py:160`), and `_convection_realized_nwp` / `_apply_firing_gate` / the corroboration bump all already read it. ICON starts being gated the moment the field is populated. **This is a behaviour change**, and the point of the issue: some ICON MODERATE+ towers will drop one level.

Note ICON will *not* take the `nwp_precip` ladder path (`method="nwp_precip"`) in practice — that fires only when there is no tower top and no cover, and ICON emits `htop_con` whenever its scheme is active. ICON stays `method="nwp_hybrid"`.

## Scope decision: `snow_con`

`SNOW_CON` (convective snow, also accumulated, also ~250 KB/step) would make the signal a true `cp` equivalent. Recommendation: **defer it**, and note it in the code. Rationale: it doubles the single-level file count for a case (convective snow showers) that is rare in the flights we brief, and omitting it is safe by construction — the firing gate holds down only on positive dry evidence, so a winter ICON tower with `rain_con ≈ 0` but real convective snow would simply keep its current un-held-down tier, i.e. no regression, just an unclosed corner. If we do add it, sum the two rates before assigning `convective_precip_mm_h`.

## Testing

- Unit: `_ICON_CLOUD_DIAG_FIELD_MAP` maps `rain_con`; `build_icon_cloud_diagnostics` round-trips it.
- Unit: `icon_eu_previous_step` — 1h grid below 78, 3h grid above, `None` at 0.
- Unit: the de-accumulation arithmetic — monotonic accumulation → correct mm/h; **no ×1000** (the highest-value regression to pin, since it's the one place the ECMWF port is wrong by 1000×); a 3h window divides by 3; a decreasing accumulation (new run / GRIB glitch) clamps to 0 rather than going negative; first step with no predecessor → `None`, not 0.0 (`None` = unknown, which the gate treats as missing-data-safe; `0.0` would actively hold the tower down — the two are **not** interchangeable).
- Integration: an ICON pack with a deep dry tower is held down one level; one with a wet tower is not. `tests/test_convective.py:672-780` already covers the gate/ladder logic given a populated field — these tests should confirm the field arrives populated.
- Manual: refresh a convective-day flight in dev, confirm `convective_precip_mm_h` is non-null on ICON `nwp_cloud_diagnostics`, and diff the ICON convective advisory tier against a pre-change pack.

## Acceptance criteria

- [ ] ICON `NWPCloudDiagnostics.convective_precip_mm_h` is populated on GRIB-anchored hours (null only for the first hour of a window when the predecessor step is unavailable).
- [ ] Rate is in mm/h (no ×1000), correct across the 1h/3h cadence change, clamped at 0.
- [ ] Cache key bumped, so existing cached runs refetch.
- [ ] Firing gate and native corroboration now evaluate for ICON.
- [ ] `designs/convective-analysis.md` "Known gap (ICON-EU)" removed; per-model tables updated to show ICON convective precip.
- [ ] `fetch/grib/icon_eu_fetch.py:112` deferral comment removed.

## Out of scope (noted while surveying the DWD listing)

Two other native ICON convective fields we don't use, worth their own issues if we want them:

- `cape_con` — the convection scheme's own CAPE (as distinct from the `cape_ml` we already take).
- `lpi_con_max` — lightning potential index. Arguably a *stronger* "did it really fire" gate than precip rate, since it keys on electrification directly.

roznet:
## What & why

The NWP convective track grades on the model's own convective scheme, and two of its safeguards — the **firing gate** (`_apply_firing_gate`) and **native corroboration** — need a *realization* signal, `convective_precip_mm_h`. ICON-EU exposed neither convective cloud cover nor convective precip, so the gate (deliberately missing-data-safe — it holds down only on *positive* dry evidence) could never fire for ICON. Every ICON MODERATE+ tower kept its full tower-top tier forever: a deep-but-dry `htop_con` read at full native risk with nothing able to contradict it, while GFS and ECMWF both got held down.

This closes that calibration gap by decoding DWD's `RAIN_CON`.

- **Fetch** `rain_con` on the single-level cloud-diag path. Introduce a shared `ICON_EU_CLOUD_DIAG_CACHE_KEY` (bumped to `..._V2`) referenced by all four cache-key call sites (prefetch, enrichment, precache, standalone verification) so adding a variable to the one-key-for-all-variables blob forces warm caches to re-fetch instead of silently keeping the gate-less behaviour.
- **Predecessor step.** ICON is fetched on-demand exactly on the window hours with no ±margin (unlike the locally-mirrored ECMWF run), so the first window hour has nothing to difference against. Added `icon_eu_previous_step` (respects the 1h≤78h / 3h grid, `None` at 0) and prepend one leading single-level step to the cloud-diag fetch only — the model-level sounding list is untouched.
- **De-accumulate & inject.** The ICON cloud-diag merge loop now walks steps in sorted order carrying the previous accumulated value + valid time and injects the mm/h rate via `model_copy`, a direct port of the ECMWF `cp` path. Crucial unit difference: `rain_con` is kg/m² ≡ mm (**already mm — no ×1000**, unlike ECMWF `cp` which is m water equivalent). Rate clamped at 0; `None` (not `0.0`) when no predecessor exists (`None` = missing-data-safe, `0.0` would actively hold a tower down). Extracted `icon_eu_conv_rain_rate_mm_h` as a pure, directly-tested helper.
- **No analysis change.** The field rides the existing forward-fill + spatial-interpolation paths, so the firing gate and native corroboration start evaluating ICON the moment it's populated. This is a behaviour change: some ICON MODERATE+ towers will now drop one level. `SNOW_CON` deferred (safe by construction — the positive-dry-evidence gate simply keeps the tier for a rare convective-snow tower).

Docs updated (`convective-analysis`, `weather-engine-specs`, `meteorology-decisions`): "Known gap (ICON-EU)" removed, per-model tables now show ICON convective precip, and the deferral code comment removed.

## Issue linkage

Closes #421

## Testing / verification

- **Unit** (`tests/test_grib.py`): `_ICON_CLOUD_DIAG_FIELD_MAP` maps `rain_con`; `ICON_EU_CLOUD_DIAG_VARIABLES` includes it; cache key is `..._V2`; `build_icon_cloud_diagnostics` stays unaware of the rate (leaves `convective_precip_mm_h` None); `icon_eu_previous_step` across the 1h/3h grid and `None` at 0; de-accumulation arithmetic — **no ×1000**, 3h window divides by 3, negative clamps to 0, and missing inputs return `None` not `0.0`.
- **Integration** (`tests/test_grib.py`): drives `_enrich_icon_eu_cloud_diagnostics` end-to-end with a mocked decode — the leading step seeds state, the first window hour gets a real mm/h rate, subsequent steps difference against their predecessor, and a run-reset accumulation clamps to 0.
- **Prefetch/precache** (`tests/test_icon_prefetch.py`, `tests/test_grib_precache.py`): updated for the V2 key and the leading-step diag fetch.
- Affected suites green locally: `test_grib`, `test_icon_prefetch`, `test_grib_precache`, `test_convective`, `test_grib_fill`, `test_spatial_interpolation`.

https://claude.ai/code/session_01AjaCg8x3iE1Gbokj1CxKhJ

---
_Generated by [Claude Code](https://claude.ai/code/session_01AjaCg8x3iE1Gbokj1CxKhJ)_

### DIFF
--- a/designs/convective-analysis.md
+++ b/designs/convective-analysis.md
@@ -45,8 +45,9 @@ Tracked in `NWP_CAPE_TYPE` dict (`variables.py:84`), stored as `nwp_cape_type` o
 | **Convective cloud cover %** | ✓ (TCDC on convectiveCloudLayer) | ✗ | ✗ | ✗ |
 | **Convective cloud base** | ✓ (PRES → ft via std atm) | ✓ (hbas_con, m → ft) | ✗ | ✗ |
 | **Convective cloud top** | ✓ (PRES → ft via std atm) | ✓ (htop_con, m → ft) | ✓ (hcct, m → ft) | ✗ |
+| **Convective precip rate** | ✗ | ✓ (rain_con, mm, de-accum) | ✓ (cp, m we ×1000, de-accum) | ✗ |
 
-Stored in `NWPCloudDiagnostics.convective_{cover_pct,base_ft,top_ft}`.
+Stored in `NWPCloudDiagnostics.convective_{cover_pct,base_ft,top_ft}` and `convective_precip_mm_h`. ICON `rain_con` is already mm (kg/m² ≡ mm) so it is **not** ×1000, unlike ECMWF `cp` (m water equivalent); both are accumulated since init and de-accumulated in the enrichment loop.
 
 ### MetPy-Derived Thermodynamic Indices (all models with pressure levels)
 
@@ -67,7 +68,7 @@ These are derived from 13–28 pressure levels (model-dependent resolution). Ava
 | **GFS** | Open-Meteo | SB | ✓ | ✓ | Full (cover + base/top) | All | Complete |
 | **Best Match** | Open-Meteo | SB | ✓ | ✓ | Full (via GFS) | All | Complete |
 | **ECMWF** | Open-Meteo | MU | ✗ | ✗ | None | All | Good (MU-CAPE + MetPy) |
-| **ICON-EU** | Open-Meteo | ML | ✗ | ✗ | Partial (base/top only) | All | Good (base/top + MetPy) |
+| **ICON-EU** | Open-Meteo | ML | ✗ | ✗ | Partial (base/top + conv precip, no cover) | All | Good (base/top + conv precip + MetPy) |
 | **UKMO** | Open-Meteo | ? | ✓ | ✗ | None | All | Fair (MetPy-only risk) |
 | **MétéoFr** | ✗ | — | ✗ | ✗ | None | All | MetPy-only |
 | **GEM** | ✗ | — | ✗ | ✗ | None | All | MetPy-only |
@@ -88,6 +89,7 @@ GFS GRIB2 ──→ decode_cloud_diag_per_point()
 
 ICON-EU GRIB2 ──→ decode_icon_eu_cloud_diag_per_point()
               ──→ NWPCloudDiagnostics.convective_{base_ft,top_ft}  (no cover_pct)
+              ──→ rain_con (accumulated) → convective_precip_mm_h (de-accum, #421)
 ```
 
 ### Stage 2: Temporal Forward-Fill
@@ -142,7 +144,7 @@ ICON-EU GRIB2 ──→ decode_icon_eu_cloud_diag_per_point()
   - **Primary scale — convective tower top** (`convective_top_ft`, the one native field common to GFS/ECMWF/ICON): `_CONV_TOP_FL_THRESHOLDS` (FL380 EXTREME / FL280 HIGH / FL200 MODERATE / FL120 LOW / present-but-shallow MARGINAL).
   - **Cover modifier** (GFS, where `convective_cover_pct` is present): numerous cover (≥35%) bumps the top-derived tier up one level, capped at HIGH. Cover-only (no top) sets a depth-unknown tier capped at MODERATE. The MODERATE cap is on the *base* tier only — Phase-2 corroboration (below) can still raise a strongly-realized cover-only MODERATE to HIGH when native precip/indices are strong. This can't fire today (GFS exposes no native `k_index`/`total_totals` and ECMWF takes the tower-top path), but the cap is intentionally not a hard ceiling against corroboration.
   - **Precip-rate fallback (Phase 3, method `"nwp_precip"`):** when the model emits *no* tower top and *no* cover fraction but `convective_precip_mm_h > 0.1` — the ECMWF marine / elevated-convection case, where `cp` lands but `hcct` is sentinel/absent — the tier comes from a convective-precip-rate ladder (`_CONV_PRECIP_MM_H_THRESHOLDS`: ≥2.0 → MODERATE, ≥0.5 → LOW, ≥0.1 → MARGINAL) instead of the old false NONE. Tower top stays primary whenever present; this is the geometry-absent fallback only. Depth is unknown from rate alone, so the ladder is capped at MODERATE. Realized by construction, so it skips the firing gate hold-down and the precip corroboration (which would double-count `cp`), and the strong-CIN suppression below (penalising a demonstrably-firing scheme on surface/ML CIN is circular). Rate is resolution-dependent → synoptic-scale v1.
-  - **Firing gate (Phase 2):** a MODERATE+ tower is only kept there when the scheme *realized* convection (`convective_precip_mm_h > 0.1` OR `convective_cover_pct > 15`); a deep-but-dry tower is held down one level. Missing-data-safe — held down only on positive dry evidence. **Known gap (ICON-EU):** ICON exposes neither `convective_cover_pct` nor (yet) `convective_precip_mm_h`, so realization can't be evaluated and every ICON tower is kept at full native tier — no dry-tower suppression until `rain_con` is decoded. Tracked as a calibration gap.
+  - **Firing gate (Phase 2):** a MODERATE+ tower is only kept there when the scheme *realized* convection (`convective_precip_mm_h > 0.1` OR `convective_cover_pct > 15`); a deep-but-dry tower is held down one level. Missing-data-safe — held down only on positive dry evidence. All three GRIB models can now be evaluated here: ICON gained a `convective_precip_mm_h` signal in #421 (DWD `RAIN_CON`, de-accumulated), so a deep-but-dry ICON tower is held down like GFS/ECMWF rather than kept at full native tier. Its only residual blind spot is a cold-air-mass **convective snow** shower (`RAIN_CON ≈ 0` but real `SNOW_CON`), which is safe by construction — the gate holds down only on positive dry evidence, so such a tower simply keeps its current tier (`SNOW_CON` deferred, see the code note in `icon_eu_fetch.py`).
   - **Native corroboration (Phase 2):** a realized MODERATE+ cell with strong native `k_index`/`total_totals`/conv-precip bumps up one level (capped HIGH). Thresholds (`_CORROB_K_INDEX=35`, `_CORROB_TOTAL_TOTALS=50`) are North-American severe-convection defaults; European environments (lower lapse rates, more modest moisture) often need TT ≥55, so this may fire readily over Europe — flagged for eval-digest calibration.
   - **Strong-CIN suppression** kept (prefers the model's own `ml_cin_jkg`, else DD `cin_surface_jkg`) — but *not* applied on the `"nwp_precip"` path (see Phase 3 above).
 
@@ -159,7 +161,7 @@ ICON-EU GRIB2 ──→ decode_icon_eu_cloud_diag_per_point()
 | Model | NWP Convective Result | Notes |
 |-------|----------------------|-------|
 | **GFS** | `method="nwp"` — risk from tower top + cover modifier; cover/base/top from GRIB | Only model with convective cover % |
-| **ICON-EU** | `method="nwp_hybrid"` — risk from tower top; GRIB base/top; `cape_ml`/`cin_ml` native | — |
+| **ICON-EU** | `method="nwp_hybrid"` — risk from tower top; GRIB base/top; `cape_ml`/`cin_ml` + `rain_con`→conv-precip native | Firing gate + native corroboration now evaluate ICON (#421) |
 | **ECMWF** | `method="nwp_lcl_top"` when `hcct` present; `method="nwp_precip"` (Phase 3) when `hcct` sentinel but `cp` fires — risk from the precip-rate ladder; `cp`/`kx`/`totalx`/`mlcape`/`mlcin` native | `hcct` is sparsely delivered (sentinel over marine/elevated convection), so `cp` is often the only firing signal |
 | **Others** | None | No diagnostics at all |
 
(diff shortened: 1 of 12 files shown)


roznet:
Thanks — both findings are non-blocking (as noted, neither affects correctness). Addressed in a526328:

**1. precache warm-gap.** Added a code comment rather than the leading-step fetch. The airport-profile hour set is gappy across day boundaries, so a single `min()` leading step gives no parity anyway, and per-block predecessor warming isn't worth the complexity for a warm-path-only win — a briefing starting on a boundary hour just fetches that one predecessor live in `_enrich_icon_eu_cloud_diagnostics`, gate stays missing-data-safe. The comment prevents mistaking this for full prefetch parity.

**2. real-GRIB validation.** Good catch. I tried to close this empirically with a live DWD pull, but this CI/sandbox env's network policy rejects `opendata.dwd.de` (403 on CONNECT), so I can't decode a real `RAIN_CON` file here. Tempered the "RESOLVED" note in `meteorology-decisions.md` with an explicit caveat: the product/URL was verified live in the issue and the de-accumulation is unit + mocked-decode tested, but the cfgrib decoded shortName still wants a one-time real-GRIB confirmation. It does follow the same lowercased-shortName convention as the sibling single-level fields (`ceiling`, `clcl`, …) that decode correctly in prod, so risk is low — but flagging it for a manual check when convenient.

---

roznet:
Thanks — traced both cited paths; no correctness issue, and addressed the completeness gap in 6c5555a.

**Finding 1 (nwp_precip reachability for ICON).** Verified: in `assess_native_convective`, the `base is not None and top is not None → "nwp_hybrid"` branch is evaluated *before* the `nwp_precip` branch (which requires `top is None`). Per issue #421's domain note, ICON emits `htop_con` whenever its scheme is active, so a firing ICON point (rain_con > 0) always carries tower geometry → `nwp_hybrid`, and never reaches `nwp_precip` in practice. In `convective_character.py`, `native_cp` only changes `realized` when `has_geom` is False — which isn't the case for a firing ICON point (`realized = … or has_geom`). So no behavioural change on that path for ICON.

Closed the gap anyway:
- Added ICON-labeled tests — a deep-but-dry ICON tower (base+top, rate 0.0) is held down HIGH→MODERATE by the firing gate (the actual #421 behaviour), and both dry/wet ICON towers assert `method == "nwp_hybrid"`, pinning that a populated `convective_precip_mm_h` never reroutes ICON onto the `nwp_precip` ladder.
- Clarified the ICON row in `convective-analysis.md` accordingly.

(Note the existing `test_nwp_firing_gate_*` tests already exercise the base+top = ICON shape with/without precip; the new tests add the explicit `method` assertion and the ICON framing.)

**Finding 2 (live-GRIB shortName).** Agreed it shouldn't be deferred indefinitely. I can't run it in this sandbox — the network policy rejects `opendata.dwd.de` (403 on CONNECT) — so it needs a manual/dev-env decode of one real `RAIN_CON` file to confirm the cfgrib shortName. Recommend doing it immediately post-merge (a `refresh` on a convective-day ICON flight, confirming `convective_precip_mm_h` populates on `nwp_cloud_diagnostics`). If the shortName differs, the field map silently returns None and the feature no-ops safely — so it's a "confirm it took effect" check, not a correctness risk.

No blockers from either review round. Recommending merge.

---

roznet:
Good catches — both fixed in 1ebeeec.

**1. De-accumulation desync on a missing `rain_con` step (data-integrity).** Confirmed and fixed. The state-carry now assigns `prev_rain_con_per_point[i] = raw.get("conv_rain_kg_m2")` unconditionally, so a step lacking `rain_con` (failed DWD file for that hour, or an uncovered point) resets the predecessor to `None`. The next step then differences against `None` → `None` (missing-data-safe) instead of dividing a multi-step accumulation delta by a single-step window and inflating the rate. Added a regression test (`test_missing_rain_con_step_invalidates_predecessor`): an intermediate step with geometry but no `rain_con` → that step's rate is `None` and the following step's rate is `None`, not the inflated `(4.5 − 1.0)/1h = 3.5`. You're right it was safe-direction only, but it broke the exact None-vs-0.0 invariant the rest of the path is built around, so worth closing.

The sibling ECMWF `cp` loop does share this carry pattern (`prev_a1_valid_utc` advances while `prev_cp_per_point` only updates on a reading) — left out of scope here as you noted; happy to open a follow-up issue covering both if you'd like.

**2. Dead doc pointer.** Fixed — `convective-analysis.md` pointed at a `SNOW_CON` "code note in `icon_eu_fetch.py`" that didn't exist. Added the note at the fetch site documenting the #421 scope decision to defer `SNOW_CON`, so the pointer now resolves.

No blockers outstanding. The only open item remains the one-time live-GRIB `rain_con` shortName confirmation (can't run in this sandbox — DWD blocked). Recommending merge.

---

Issue #441, opened 2026-07-16
Backend GRIB decode audit: 7 correctness findings + 5 efficiency items (externally grounded) — ECMWF/GFS/ICON


roznet:
Backend GRIB decode audit (ECMWF / GFS / ICON) — 7 correctness findings + 5 efficiency items, externally grounded.

Origin: external LLM review (GPT-5.6). **Every claim below was independently re-verified against the code**, and every *convention/interpretation* claim (sign, datum, statistical processing) was **grounded on ground truth** — decoded from our own cached GRIB via eccodes + cross-checked against the provider parameter database. This issue is the paper trail; sub-PRs reference it.

## How claims were verified (provenance)

Two independent kinds of ground truth were used:

1. **The GRIB message's own self-description** — eccodes keys (`name`, `units`, `stepType`, `typeOfStatisticalProcessing`, `paramId`, `missingValue`) + the **actual decoded value distribution** from cached runs. This is authoritative for what our pipeline actually receives. eccodes ships ECMWF's official parameter tables, so the reported `name`/`units` *are* the param-DB content.
2. **Provider parameter documentation** — ECMWF Parameter DB, ECMWF Confluence Wiki, DWD ICON database description, NCEP GFS product inventory.

Files decoded (2026-07-16):
- ECMWF: `/Users/brice/tmp/ecmwf/data/brg_a1_ifs-ens-cf_od_oper_fc_20260716T120000Z_..._24h`
- ICON-EU: `data/.cache/grib/icon-eu/20260716_15z/f024_ICON_EU_{CLOUD_DIAG_V2,U,W}.grib2`

**Grounding status legend:** ✅ decoded-confirmed · 📖 external-doc-confirmed · 🧩 code-logic (self-evident, no external ref needed) · ⏳ open

---

## Correctness findings

| # | Finding | Severity | Grounding |
|---|---|---|---|
| 1 | ICON zeros valid negative U, V, W | Critical | ✅🧩 |
| 2 | ECMWF/ICON model CIN uses opposite sign convention → strong cap never suppresses | Critical | ✅📖 |
| 3 | AGL vs MSL heights mixed in one `ceiling_ft`/height field | High | ✅📖 |
| 4 | ICON CLC cloud layers reused across forecast hours | Medium (bounded fallback-fill) | 🧩 |
| 5 | GFS instantaneous cover treated as time-averaged | High | 🧩📖 |
[...]

PR #444, opened 2026-07-17
fix(grib): AGL/MSL airport ceiling datum + vectorise ICON interp (#441)

roznet:
Second wave of the GRIB decode audit (#441): the AGL/MSL datum fix, the ICON interp vectorisation, and the resolution of the ECMWF multi-grid question. Validated against real GRIB and a 422-pack corpus replay.

**Addresses #441** (the remaining deferred items). Two sibling datum surfaces stay open (see below).

### `fix(airport): convert ceiling to AGL for flight category` — #3 (behavioural)
Airport flight-category used AGL thresholds (LIFR<500 … MVFR<3000 ft AGL) against a ceiling that was **MSL** for GFS/ICON and the sounding, **AGL** only for ECMWF, with no terrain subtraction — and `reconcile_ceiling`'s `min()` mixed datums before the threshold. At elevated airports this over-stated the ceiling by the field elevation → **optimistically-clear categories**.

Fix (datum-aware, at the single airport consumer): `reconcile_ceiling` gains optional `field_elevation_ft` + `model`, converts sounding (MSL) and NWP ceiling (MSL for GFS/ICON, AGL for ECMWF) to a common **AGL** datum, then takes the lower. Field elevation comes from the airport's **published nav.db elevation** (`get_airport_elevations`) — exactly what METAR AGL references — plumbed via `compute_airport_conditions` from `advise.py` and the `packs.py` recompute paths. Without elevation it keeps the legacy `min()` (unmigrated callers unaffected). `AirportModelCondition.ceiling_ft` is now AGL.

**Corpus replay (422 packs, 226 GFS/ICON airport-conditions with known elevation):**
- **20 (8.8%) change flight category — all more restrictive** (VFR→MVFR, MVFR→IFR, IFR→LIFR); zero more-permissive → the old behaviour was optimistically wrong.
- Concentrated at elevated fields: EDNY 1367 ft, LJBL 1654 ft, EDDS 1276 ft, LIPB 789 ft. Sea-level airports ≈ unchanged.
- e.g. EDNY 4016 ft MSL − 1367 ft field = 2649 ft AGL → correctly MVFR, not false VFR.

### `perf(grib): vectorise ICON model-level interpolation` — efficiency #2
Replaced the per-level xarray `.interp()` loop with the batched-bilinear `_frac_grid_indices` machinery already used for ECMWF/GFS. **Verified numerically identical** on real cached ICON (U/V/W/T/QV/P/QC/QI, 40 levels): worst Δ ~1e-13, zero None-mismatches. Single-thread gain is modest (~1.1×, cfgrib open dominates); the real benefit is releasing the GIL under parallel fhour decode.

## #5 ECMWF multi-grid first-wins — resolved, no code change
Investigated real deliveries: the "overlapping" grids in the (var,level) index are **disjoint geographic domains at identical 0.25° resolution** — (99,229)=Europe, (47,151)=Scandinavia, (114,229)=N.America. A route point is in exactly one, so first-wins never competes, and "prefer highest resolution" is moot (all same res). Nothing to fix.

## Testing
`3604 passed, 8 skipped, 0 failed` (full suite). New datum tests cover ICON MSL→AGL conversion, ECMWF no-double-subtract, below-field clamp, and legacy-unchanged.

## Deferred (tracked in #441)
Same datum defect on `airport_consensus.best_ceiling` (verification/scoring) and `map_queries._alt_required` (forecast-map category) — they shift separate metrics that need their own behaviour diff.

https://claude.ai/code/session_01225CifyZ4VTSQyUDJJ6GMq

### DIFF
--- a/designs/alternates.md
+++ b/designs/alternates.md
@@ -100,12 +100,18 @@ TS implementations are pinned identical by `tests/fixtures/consensus_vectors.jso
 operating on a **plain snapshot dict** whose keys equal `AirportForecastSnapshotRow`
 column names (NOT the ORM row):
 
-- `best_ceiling(snap)` — `min(sounding_ceiling_ft, nwp_ceiling_ft)` when either
-  primary estimate is present (the conservative reconciliation shared with
-  `analysis.airport_conditions.reconcile_ceiling`), else falls back to
-  `cloud_base_ft` then `lcl_ft`
-- `flight_category(snap)` — `classify_flight_category(ceiling, visibility_mi)`
-- `snap_to_dict(snap)` — column-keyed → lightweight per-model dict
+- `best_ceiling(snap, *, field_elevation_ft=None)` — `min(sounding_ceiling_ft,
+  nwp_ceiling_ft)` when either primary estimate is present (the conservative
+  reconciliation shared with `analysis.airport_conditions.reconcile_ceiling`),
+  else falls back to `cloud_base_ft` then `lcl_ft`. When `field_elevation_ft` is
+  given, each estimate is converted to **AGL** first (sounding/GFS/ICON are MSL
+  → subtract elevation; ECMWF NWP and `lcl_ft` are already AGL) via the shared
+  `to_agl_ceiling` helper, so the ceiling matches the AGL flight-category
+  thresholds and METAR — the model for the ECMWF exception is read from
+  `snap["model"]`. Callers (`map_queries`, `alternates`) pass the airport's
+  published `get_airport_elevations` value. (#441 finding #3)
+- `flight_category(snap, *, field_elevation_ft=None)` — `classify_flight_category(ceiling, visibility_mi)`
+- `snap_to_dict(snap, *, field_elevation_ft=None)` — column-keyed → lightweight per-model dict
 - `enrich_wind(d, runway_ends)` — `compute_runway_winds`, best runway, gusts
 - `consensus(per_model, mode=<aggregation>)` — default `"majority"`; the mode is
   the user's advisory aggregation preference (see "Consensus" above)
--- a/src/weatherbrief/airports.py
+++ b/src/weatherbrief/airports.py
@@ -246,3 +246,25 @@ def get_runway_ends(icao_codes: list[str], db_path: str) -> dict[str, list[Runwa
         result[icao] = ends
 
     return result
+
+
+def get_airport_elevations(icao_codes: list[str], db_path: str) -> dict[str, float | None]:
+    """Get published field elevation (ft AMSL) for given airports.
+
+    Used to convert MSL model/sounding ceilings to AGL for aviation
+    flight-category classification — METAR ceilings are AGL above this
+    published field elevation. (#441 finding #3)
+
+    Returns:
+        Dict mapping ICAO code to elevation in feet AMSL, or None if unknown.
+    """
+    model = _load_airport_model(db_path)
+    result: dict[str, float | None] = {}
+    for icao in icao_codes:
+        airport = model.airports.get(icao)
+        result[icao] = (
+            float(airport.elevation_ft)
+            if airport is not None and airport.elevation_ft is not None
+            else None
+        )
+    return result
--- a/src/weatherbrief/analysis/airport_conditions.py
+++ b/src/weatherbrief/analysis/airport_conditions.py
@@ -140,9 +140,22 @@ def _ceiling_from_sounding(sounding: SoundingAnalysis) -> float | None:
     return min(ceilings) if ceilings else None
 
 
+def _nwp_ceiling_is_agl(model: str | None) -> bool:
+    """Whether a model's NWP ceiling is already AGL (height above ground).
+
+    ECMWF ceil/cbh/hcct are AGL; GFS (geopotential) and ICON (CEILING/HBAS_CON)
+    are MSL. Unknown/aggregate models (best_match, …) are treated as MSL — the
+    majority case, and the conservative direction at elevated airports. (#441)
+    """
+    return (model or "").lower() == "ecmwf"
+
+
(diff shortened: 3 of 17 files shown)


roznet:
Thanks — triaged. **No blockers**; not pushing (doc/test/citation-only would cost a full review round for no correctness gain). Disposition:

**Important 1 — `scoring.py:173` + `route_weather.py:435` (deferred, now listed).** Both call the same `reconcile_ceiling`, correct catch. But migrating them shifts **verification** (`category_match`/`ceiling_delta_ft`) and **obs-model comparison** metrics at elevated fields — the same "moves a separate metric, needs its own behaviour diff" reasoning I applied to `best_ceiling`/`map_queries`, just reached via the shared function rather than a reimplementation. Adding them to the deferred set: **datum cleanup for the verification & comparison surfaces** (scoring, route_weather obs-comparison, `best_ceiling`, `map_queries`) — each migrated + diffed together in a follow-up, not silently in this PR.

**Important 2 — non-issue (mistaken premise).** euro_aip's `Airport` dataclass exposes **only `elevation_ft`** (`dataclasses.fields` confirms). The `_resolve_airport_coords` fallback to `elevation_feet`/`elevation_m`/`elevation_meters` is dead code — those attributes don't exist on the model, so `getattr` always returns None. `get_airport_elevations` reading `elevation_ft` is therefore complete; there's no airport where a different name would be populated. Will add a debug-log when it resolves None (so the legacy fallback isn't silent) + optionally a shared helper for dedup — on main.

**Minor 3 — ICON MSL claim, now evidenced.** Confirmed on the raw GRIB: `HBAS_CON`/`HTOP_CON` eccodes names are literally *"…above msl"* (same DWD cloud-geometry family as `ceil`), and the corpus replay moved categories only optimistic→conservative — both consistent with ICON ceiling = MSL. Will cite this in `_nwp_ceiling_is_agl`.

**Minor 4/5 — regression test + docstring.** Both worth doing; folding into the same main commit after merge.

Merging + fixing 2–5 on main rather than triggering another review round for non-blocking items.

roznet:
### Round 2 — extended the datum fix to the whole compare/classify class

Rather than defer, migrated **every** site that compares/classifies a model ceiling against AGL thresholds or AGL METAR (Important #1 expanded from 2 sites to the full class):

- **Group A** (call the same `reconcile_ceiling`): `scoring.py::_score_model_vs_metar` (verification `category_match`/`ceiling_delta_ft`), `route_weather.py::run_observation_comparison` + both callers (`pipeline.py`, refresh).
- **Group B** (the shared `airport_consensus` engine behind the forecast map + alternates): `best_ceiling`/`snap_to_dict` now AGL-convert and read the model from the snapshot; `map_queries.py` (elevation cache, resilient), `alternates.py::_assess`/`_build_alternate` (FAA/EASA triggers).
- **Dedup (Important #2):** extracted one shared `to_agl_ceiling()` helper; `reconcile_ceiling` and `best_ceiling` both use it. Verified euro_aip `Airport` exposes **only** `elevation_ft` (the fallback the review cited is dead code), so `get_airport_elevations` is complete; the map elevation lookup now degrades to datum-naive on failure rather than erroring.
- **Minor #3:** cited the ICON-MSL evidence (eccodes names read "…above msl" + DWD doc) in `_nwp_ceiling_is_agl`.
- **Minor #4:** added a synthetic-dataset regression test for the vectorised ICON interp vs the per-level reference (CI-safe, monkeypatched cfgrib).
- **Minor #5:** documented the `airport_elevations` parameter.

All new params default to the legacy datum-naive path when elevation is absent. Full suite: **3612 passed, 0 failed**.

Note: this shifts verification/forecast-map/alternate metrics at elevated fields the same way the airport panel moved (optimistic→conservative) — the mechanism is identical to the corpus-validated 8.8%.

Nothing outstanding from round 1. One same-class read remains for a follow-up: `api/airport_profile.py:330` reads raw `nwp_ceiling_ft`/`sounding_ceiling_ft` (MSL) for the airport-profile chart series.

roznet:
Round 2 fix pushed. **Critical (lcl_ft):** confirmed — lcl_ft is the Espy surface approximation, already AGL; converting it with `source_is_agl=True` (unchanged), + regression tests for the lcl_ft and cloud_base_ft fallback rungs. **Minor:** split the pipeline.py elevation lookup into its own try/except so it can't discard good runway data. Full suite 3614 passed, 0 failed.

roznet:
### Round 3 fixes pushed

**Important #1 (blocker — class completeness):** confirmed — `standalone_verification._score_cycle` was the last un-migrated `_score_model_vs_metar` caller. Now threads `field_elevation_ft` (via `get_airport_elevations`, wrapped resiliently), so the standalone verification path scores on the same AGL datum as the flight-linked `score_flight`.

**Important #2 (dedup):** extracted the per-dataset bilinear corner/weight math into a shared `_bilinear_grid_weights` helper used by both `_decode_pressure_vars_from_datasets` and the vectorised ICON path — the ~25 duplicated lines are gone; a future numerical fix lives in one place. Validated by the full grib suite + the equivalence regression test.

**Minor #4/#5:** `designs/alternates.md` now documents the `field_elevation_ft` parameter + AGL behaviour; the alternates elevation import folded into the top-level import.

**Minor #3 (PR-body correction):** you're right — `best_ceiling` and `map_queries._alt_required` are **fixed** in this PR (via the now-AGL `best_ceiling`/`snap_to_dict`), not deferred. The corrected remaining follow-up under #441 is: `standalone_verification.py:1208-1212` (`cloud_base_delta_ft`/`lcl_delta_ft`) compares raw MSL `cloud_base_ft`/`lcl_ft` against AGL METAR — pre-existing, not touched here, tracked for #441.

Full suite: **3614 passed, 0 failed**.

PR #449, opened 2026-07-18
perf(standalone): fix 40-min cache rebuild + cycle-aware post-tasks + verification indexes (PR A of #448)

roznet:
## Summary

PR A of #448 — removes the deterministic ~40-minute post-cycle cache rebuild and the smaller measured inefficiencies. Expected: forecast cycle **~110 → ~72 min**, light cycle **~40 → ~4 min**, ~4.5 h/day less MySQL load. (PR B will parallelize the sounding analysis that dominates the remaining fetch phase.)

## Changes

**1. `COUNT(DISTINCT)` query-plan fix (the 38-minute item)**
`get_activity_summary`'s two distinct-count queries let MySQL pick `ix_verif_scores_source_model_days` with `ref=const` on `source` alone — scanning all 3.8M standalone rows regardless of time window (7d and 30d each took ~19 min; 24h got the range plan and took 2 s). Verified on prod: `FORCE INDEX (ix_verif_scores_source_time)` → ~6 s. The hint is MySQL-only (`dialect_name="mysql"`); SQLite renders unchanged (compile-tested both dialects).

**2. Cycle-aware `run_post_cycle_tasks`**
Forecast cycles create no scores → skip rollup + stats/leaderboard caches, rebuild only the forecast map (~24 s). Light cycles keep skipping the map. `full` (CLI/legacy) rebuilds everything. `rebuild_all` gains `include_score_stats` mirroring the existing `include_forecast_map`.

**3. Migration 080 — two query-shape indexes**
- `verification_daily_stats (source, model, days_out, date)`: the bias-leaderboard filters equality columns with a date range; existing indexes lead with `date`/`icao`, so 30d/90d keys were 985K-row full scans (~9.5 s × 18 keys per rebuild).
- `taf_verification_scores (source, observation_time)`: TAF pseudo-model aggregates run at query time from raw and were 287K-row full scans.

**4. Bulk snapshot insert** — `_store_snapshots` now does chunked Core `executemany` instead of building 15–20K ORM objects (~1–2 min/cycle).

**5. Inert EDR pass removed** — the #221 accumulator iterated every derived level of ~56K soundings/cycle while collecting nothing (`analyze_sounding_lite` never computes Richardson). Plumbing kept; #221 stays parked.

[...]

Issue #448, opened 2026-07-18
Standalone forecast cycle: ~110 min → ~30 min (40-min cache rebuild + single-core sounding analysis)


roznet:
## Problem

The twice-daily standalone forecast cycle (07:15/19:15 UTC) takes ~1h50–2h. Log mining of the two 2026-07-17 cycles + prod DB probing attributes it precisely (full analysis: `scratch/standalone-cycle-debrief/ANALYSIS_optimization_review.md`):

| Phase | Time | Root cause |
|---|---|---|
| GFS+ICON+ECMWF legs | ~68 min | 85–92% is **single-core** MetPy sounding analysis (~56k `analyze_sounding_lite` calls, ~57 ms each on the droplet); GRIB decode is only 3–4 min total |
| Cache rebuild | ~40.5 min | **~38 min = two `COUNT(DISTINCT)` queries** in `get_activity_summary`: MySQL picks the `source`-prefix index and scans all 3.8M standalone rows, window-independent (7d≈30d≈19 min; 24h gets the range plan, 2 s). `FORCE INDEX (ix_verif_scores_source_time)` → 6 s (verified on prod). Attribution via `verification_cache.computed_at` timeline — the rebuild logs nothing internally |
| Daily rollup | ~1 min | fine |

The same rebuild runs after the 5 daily light cycles too (the whole reason they take ~40 min) ≈ 4.5 h/day of MySQL grinding against a **128 MB default buffer pool** (~2.7 GB data+indexes on `verification_scores` alone).

Secondary findings (verified): `verification_daily_stats` indexes lead with `date`/`icao` so the 30d/90d bias-leaderboard queries full-scan 985k rows (~9.5 s × 18 keys); `taf_verification_scores` has no `(source, observation_time)` index (287k-row scans); `_store_snapshots` builds 15–20k ORM objects row-by-row; the parked EDR accumulator (#221) still iterates every derived level for zero collected data.

## Plan

**PR A — cycle cost (code + one migration)**
1. `FORCE INDEX` hint on the two `COUNT(DISTINCT)` queries (MySQL-only, SQLite unaffected)
2. Cycle-aware `run_post_cycle_tasks`: forecast cycle → forecast-map rebuild only (it creates no scores); light cycle → rollup + stats/leaderboards only (already skips map)
3. Migration: indexes `verification_daily_stats (source, model, days_out, date)` + `taf_verification_scores (source, observation_time)`
4. Bulk-insert forecast snapshots
[...]

PR #450, opened 2026-07-18
perf(standalone): pooled sounding analysis in the cycle child (PR B of #448)

roznet:
## Summary

PR B of #448 — parallelizes the cycle's dominant cost: ~56K GIL-bound `analyze_sounding_lite` calls (~55–60 single-core minutes per forecast cycle, measured in the #448 analysis). Expected: fetch phase ~68 → ~25–30 min, **forecast cycle ~72 → ~30–35 min** (light cycles unaffected — no sounding work). Completes the #448 plan together with the merged PR A (#449) and the buffer-pool config.

## Design: same pool code, separate pool instance

The disposable cycle child now runs its **own** bounded pool using the existing decode-pool + dispatcher machinery (spawn lifecycle, per-job timeout, fault recovery, dead-letter — all reused). The web app's 3-worker pool is untouched; the child's workers are reniced (`--background`), so interactive work preempts them and a worker OOM kills only the disposable child. This is the deliberate, bounded version of the second pool that #236 avoided by accident-prevention — justified now that #448 measured the cycle as CPU-starved, not memory-starved (peak_rss ~1.2 GB / 6 GiB).

## Changes

- **`analysis/sounding/snapshot_fields.py` (new)** — single shared implementation for inline + pooled paths: `compute_snapshot_sounding_fields` (profile → the 7 snapshot scalars, fail-silent), `build_sounding_payload` / `analyze_sounding_batch_items` (plain-dict IPC serialisation). `_enrich_with_sounding` delegates to it, so the two paths cannot drift.
- **`decode_worker.analyze_sounding_batch`** — first non-decode worker entry point: ~100-profile batches, pure/idempotent (dispatcher may retry freely), lazy MetPy import so the web app's pool boot doesn't pay it.
- **Call sites** — gfs/icon chunks and each ECMWF step collect payloads during the loop, batch-dispatch via `_dispatch_decode_parallel`, merge fields back by index. Dispatch failure degrades to surface-only snapshots (identical contract to inline failure — scoring skips sounding comparisons). ECMWF merges per step so pending payloads stay bounded at ~619.
- **Opt-in gating** — `pool_soundings=True` is passed only by `run_standalone_cycle`; the interactive alternates path (which shares these fetchers, `tasks/alternates.py`) stays inline — zero behavior change for briefings.
- **Scheduler** — the child's `GRIB_DECODE_WORKERS` now comes from **`STANDALONE_ANALYSIS_WORKERS` (default 2 — no `.env` change needed; `0` restores the old inline behaviour as a rollback switch)**. Side effect: the child's ECMWF/gfs/icon GRIB decodes use the pool too.
- **CLI** — tidy `shutdown_decode_pool(wait=False)` at exit so a hung worker can't block the child from exiting.
- `decode_pool_enabled()` public helper; `designs/grib-decode-dispatcher.md` + `designs/metar-taf-accuracy.md` updated.

## Evidence & testing

- Local benchmark (2 workers, 3×100 profiles): 1.44× — exactly the theoretical bound for 3 jobs on 2 workers (one worker takes 2 jobs); evens out to ~1.9–2× at cycle scale (30+ batches/chunk). IPC overhead negligible.
- 141 tests pass across the affected suites, including new `tests/test_sounding_batch.py`: batch-path ≡ inline-path equivalence, per-item fail-silence, merge-by-index, dispatch-failure degradation, opt-in gating, and a **real spawn-pool round-trip** (catches pickling/worker-import regressions; `@pytest.mark.slow`, verified passing).
- Subprocess command test updated: child env now `GRIB_DECODE_WORKERS=2` by default, `0` honored as override.

## Rollout / validation

- Watch the first forecast cycle after deploy: `analyzed N snapshots in Xs (pooled)` per chunk, `ECMWF GRIB leg: … (pooled)`, and `peak_rss_mb` on the cycle row (expect roughly +300–600 MB over the ~1.2 GB baseline for 2 workers).
- After a few clean cycles, bumping to 3 workers is a one-line `.env` change (`STANDALONE_ANALYSIS_WORKERS=3`) worth ~8–10 more minutes.
- Deferred (not worth it yet): ECMWF decode-ahead — decode is only ~2.5 min serialized; revisit if the post-deploy timings say otherwise.

Part of #448 — merging this completes the plan; the issue can close on deploy.


roznet:
Review round 1 addressed in the latest push:

- **Important 1 — priority not threaded**: fixed. `_analyze_soundings_pooled` now takes `priority`; the gfs/icon chunk path passes `DecodePriority.BACKGROUND` explicitly and `fetch_ecmwf_grib_snapshots` forwards its existing `priority` param (the in-scope oversight the review spotted). New test pins the forwarding.
- **Important 2 — teardown not in finally**: fixed. `cmd_standalone` wraps the cycle + rollup in `try/finally`; the `wait=False` teardown now runs on the failure path, with a comment explaining the atexit blocking-join hazard it prevents.
- **Minor — whole-batch failure blast radius**: fixed rather than commented (push already triggered a round). `_dispatch_decode_parallel` gains `return_exceptions=True` (gather-style; the legacy rollback path represents its shared-fate failure as the exception in every slot), and the merge now skips only the failed batch's ~100 snapshots with a count in the log.

Nothing deferred. 138 tests + the slow real spawn-pool round-trip pass.

roznet:
Review round 2 addressed in ccc049b6, batched with a teardown hardening from an external (codex) review so this is a single push:

**Round-2 findings — all fixed:**
- **Important 1 (fallback shares web-app pool)**: `pool_soundings` is now an explicit `run_standalone_cycle` parameter, default False; only the standalone CLI passes True. The `STANDALONE_SUBPROCESS=0` fallback stays inline. Gating pinned by new tests (`TestPoolSoundingsGating`) plus an assertion in the CLI mapping test.
- **Important 2 (unbounded chunk accumulation)**: gfs/icon chunks now flush at `_SOUNDING_FLUSH_THRESHOLD` (5× batch size), after the snap append so indices stay valid — bounded pending like the ECMWF per-step merge, plus workers start while the chunk is still parsing.
- **Important 3 (pool RSS through cache rebuild)**: the CLI shuts the pool down gracefully right after `run_standalone_cycle`, before `run_post_cycle_tasks`; the `finally` teardown remains as the failure-path net (idempotent).
- **Minor 4**: `standalone_grib.py` docstring updated for the new child default.
- **Minor 5**: EDR-divergence note added at the harvest site in `snapshot_fields.py`.

**Hardening (same push):** `shutdown_decode_pool(force=True)` — TERM → join(2 s) → KILL of every known worker, including orphans from replaced pools via a new accumulating `_ALL_DECODE_WORKERS` registry (the per-pool registry is cleared on rebuild and can't reach them). Rationale: `wait=False` only *abandons* a wedged worker; at interpreter exit the concurrent.futures atexit hook joins forever, hanging the child until the 3 h supervisor timeout. The CLI `finally` now uses `force=True`. The web-app half (force on dispatcher timeout recovery + supervisor process-group kill) is deliberately out of scope here — tracked in #451, since it changes a documented trade-off in the #171 design.

Tests: 171 passed across the affected suites + the slow real spawn-pool round-trip; new coverage for hung-worker force-kill (<15 s teardown, workers verified dead), registry survival across pool rebuild, and no-pool no-op.

PR #452, opened 2026-07-19
feat(offload): portable snapshot artifacts — emit (P2) + ingest (P3)

roznet:
## What

Stage 1 of the standalone-cycle **compute-offload** work: two generic, provider-agnostic capabilities that let the forecast cycle be computed off the serving box and shipped in as a self-contained file. **No deployment topology lives in the repo** — this PR only adds the plumbing; where compute runs and how artifacts move is intentionally out of tree.

- **P2 — emit** (`standalone --emit-artifact PATH [--region eu|us|all]`): after a cycle, export the fresh `airport_forecast_snapshots` to a portable, versioned **SQLite artifact** (mirror table + a `_manifest` row with a content checksum, per-model init times, row count, source host).
- **P3 — ingest** (`ingest-artifact PATH`): validate an artifact (row-count + checksum) and **upsert** its rows into this DB by natural key `(icao, model, model_init_time, forecast_hour)`, recording a `verification_cycles` row (`source='standalone_forecast_imported'`).

## Why it's safe to merge

- **Additive & dormant.** Prod's scheduler runs `standalone --forecast-only --with-rollup --background` and the web app never touches these paths — the new flag/subcommand are no-ops until explicitly invoked.
- **No migration.** Nothing touches the DB schema. The `region` column is deferred to the US-expansion work; columns are derived from the ORM model so `region` is carried automatically once it lands.
- **Dialect-agnostic import.** Reads the artifact with stdlib `sqlite3` and upserts through SQLAlchemy Core (**no `ATTACH`**), so it behaves identically on dev SQLite and prod MySQL. Idempotent by natural key; rejects a corrupt/truncated/manifest-less artifact **before writing any row** (no partial import).

## Tests (`tests/test_snapshot_artifact.py`)

Round-trip equivalence vs a direct-DB cycle, latest-init-per-model selection, ingest idempotency, two-disjoint-artifacts no-loss, and corrupt-row / truncated / missing-manifest rejection. 8 new + 89 neighborhood tests (standalone_verification, map_queries, cache_builder) green. Verified the real `ingest-artifact` CLI round-trips idempotently (3 inserted, 0 on re-run).

## Deferred (follow-ups, not gaps)

- True **stage→swap** on ingest (v1 uses idempotent natural-key upsert — behavior-identical to the direct cycle).
- **`region`** filter is wired but inert until the US-expansion migration adds the column.
- A **MySQL ingest** check before any production cutover (these tests prove logic on SQLite).

### DIFF
--- a/src/weatherbrief/db/models.py
+++ b/src/weatherbrief/db/models.py
@@ -780,6 +780,24 @@ class AirportForecastSnapshotRow(Base):
     sounding_convective_risk: Mapped[str | None] = mapped_column(String(16), nullable=True)
 
 
+def snapshot_natural_key(
+    icao: str, model: str, model_init_time: datetime, forecast_hour: datetime,
+) -> tuple:
+    """Canonical natural key for `airport_forecast_snapshots` (`uq_afs_key`).
+
+    Strips tzinfo so tz-aware and SQLite's naive datetimes compare equal. Single
+    source of truth for dedup, shared by the standalone cycle's inserter and the
+    artifact importer so their idempotency logic can't drift apart.
+    """
+    init_naive = (
+        model_init_time.replace(tzinfo=None) if model_init_time.tzinfo else model_init_time
+    )
+    fhour_naive = (
+        forecast_hour.replace(tzinfo=None) if forecast_hour.tzinfo else forecast_hour
+    )
+    return (icao, model, init_naive, fhour_naive)
+
+
 class VerificationMonthlyStatsRow(Base):
     """Pre-aggregated monthly verification accuracy per airport/model/days_out.
 
--- a/src/weatherbrief/tasks/snapshot_artifact.py
+++ b/src/weatherbrief/tasks/snapshot_artifact.py
@@ -53,8 +53,10 @@ from weatherbrief.db.models import (
 # Bump when the on-disk artifact layout changes incompatibly.
 ARTIFACT_SCHEMA_VERSION = 1
 
-# verification_cycles.source tag written by import_snapshots.
-IMPORT_CYCLE_SOURCE = "standalone_forecast_imported"
+# verification_cycles.source tag written by import_snapshots. MUST fit
+# VerificationCycleRow.source (String(24)) — prod MySQL rejects/truncates a
+# longer value and, sharing the import transaction, would roll the import back.
+IMPORT_CYCLE_SOURCE = "standalone_imported"
 
 _SNAPSHOT_TABLE = "airport_forecast_snapshots"
 _MANIFEST_TABLE = "_manifest"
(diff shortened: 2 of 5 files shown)


roznet:
Addressed all 9 findings in `6348c553` (pushed). Summary for the next round:

**Blockers fixed**
- **#1** `--no-verify-checksum` now skips *only* the checksum — row-count is always validated in `_validate(verify_checksum=...)`, so a truncated artifact is still rejected (matches the flag's help text). New test: `test_no_verify_checksum_still_rejects_truncated`.
- **#2** `export_snapshots` writes to a temp file then `os.replace` — re-runnable against a fixed path (no "table already exists") and crash-safe (no partial artifact at the destination). New test: `test_export_overwrites_existing_path` (also asserts no stray temp files).
- **#3** `fetched_at` is restamped at ingest time, so ship latency can't shrink the replica's 10-day retention window and prune fresh long-lead rows. Round-trip test now compares all-columns-except-`fetched_at` and asserts the restamp.
- **#6** `ArtifactManifest.from_json` raises `ArtifactValidationError` (not bare `TypeError`) on a malformed manifest → clean CLI rejection. New test: `test_malformed_manifest_rejected`.

**Batched (since the round was already paid for)**
- **#4** imported `verification_cycles` row records this import's *own* elapsed time.
- **#5** shared `snapshot_natural_key()` in `db/models.py`, used by both `_store_snapshots` and the importer — one dedup technique, drift eliminated.
- **#7** `read_manifest` guards `os.path.isfile` first → a mistyped path errors cleanly, no stray empty `.sqlite`. New test: `test_missing_file_rejected`.
- **#8** clarifying comment on `--emit-artifact` + `--forecast-only`/`--light` semantics.
- **#9** `verification_cycles.airports` counts only rows actually inserted (no-op re-ingest records 0). Idempotency test asserts `airports == [1, 0]`.

Nothing deferred. 93 tests green locally (artifact + standalone_verification + map_queries + cache_builder).

roznet:
Round 2 addressed in `09a9da7a`.

**Blocker fixed:** `IMPORT_CYCLE_SOURCE` shortened `"standalone_forecast_imported"` (28) → `"standalone_imported"` (19) so it fits `VerificationCycleRow.source` `String(24)` — good catch that our SQLite-only tests couldn't see. Added `test_import_cycle_source_fits_column`, a dialect-independent guard that asserts the tag length against the column length so a future over-long tag fails in CI.

**Minor batched:** `cmd_standalone` now prints a warning when `--region` is passed but the snapshot table has no `region` column yet (unfiltered export).

**Deferred (non-blocking, noted for follow-up):**
- `_select_rows` loading all ORM rows before `latest_per_model` — Core/window-function query when this becomes a scheduled flow (CLI-only today).
- `from_json` reporting "malformed manifest" instead of "schema newer" for a future added manifest field — forward-compat wrinkle for whoever adds the next field.

PR #455, opened 2026-07-20
feat(ios): simpler add-PIREP flow — drop flight-window gate, add entry points, one-shot GPS pre-fill

roznet:
## What & why

The in-flight "Report PIREP" button was gated on `isInFlightWindow && pirepCanPublish`, and GPS pre-fill only worked while route tracking was actively running. Combined with the admin-only publish flag, the "+" never appeared even mid-flight (the plane was visible on the cross-section, proving tracking worked — the blocker was the hidden `pirepCanPublish` flag). This reworks the iOS flow toward **less gate, more smart pre-population**:

- **Drop the in-flight-window condition** from the briefing toolbar button — a PIREP is a position report and the form fetches its own GPS fix. Shown whenever `pirepCanPublish`.
- **PIREPs tab**: reachable by publishers (`pirepCanView` **or** `pirepCanPublish`) and now carries a permanent **"Report a PIREP"** action (empty-state button + persistent bottom bar).
- **Flight list**: new **"Add PIREP"** context-menu item (when `pirepCanPublish`) that opens the reporting sheet directly.
- **`FlightTrackingService.requestOneShotLocation()`**: a single `CLLocationManager.requestLocation()` fix with no route projection, so `PirepReportingView` pre-fills lat/lon/altitude without the pilot tapping "Start". No-op while a live track is already feeding positions.

No server change: the `flight_id` PIREP query already matches in-window reports without a `pack_id` (`storage/pireps.py`), so own reports still surface in the tab. Publishing stays gated by `pirep_can_publish` (server-enforced 403).

Design docs (`ios-app-overview.md`, `ios-app-ui.md`) updated to match the as-built behavior.

**Operational note:** `pirep_can_publish` must be enabled on the tester's account (admin endpoint) for the flow to work end-to-end — the server 403s submission otherwise.

## Issue linkage

No linked issue — direct working change.

## Testing / verification

- Static review only: this container is Linux with no Swift toolchain, so the SwiftUI/CoreLocation changes are hand-verified but **not compiled**. Needs a build on macOS before merge.
- No backend changes; existing `flight_id` PIREP query path is unchanged.

---
_Generated by [Claude Code](https://claude.ai/code/session_01EdHLwawGZ533oC5YVYv4MP)_


roznet:
Thanks — addressed all three in `e964ed9`:

- **Critical (stale altitude race):** `requestOneShotLocation()` now clears `currentLocation` at the start of the non-tracking path, so the synchronous `onAppear` pre-fill can't inherit the previous fix — the fresh one-shot fix wins. This fixes it at the service level, which covers all three entry points (they share a `FlightTrackingService`/`pirepTrackingService` instance) and the across-flights case in `FlightListView`. The live-tracking path is untouched (`guard !isTracking` returns before the reset).
- **Important (offline gate):** dropped the `!isOffline` condition on the flight-list "Add PIREP" item. `PirepViewModel.submit` already detects a transient-network failure and enqueues via `PirepOfflineStore`, so this now matches the toolbar + PIREPs-tab entry points and the documented offline-queue flow. (This also makes the `ios-app-ui.md` "gated on `pirepCanPublish` only" line accurate again — no doc edit needed.)
- **Minor (stale copy):** reworded the no-GPS submit error to "No GPS position yet — wait a moment for a location fix, or enable location access in Settings".

Note: still not machine-compiled (Linux session, no Swift toolchain) — needs a macOS build before merge.

---

roznet:
Round-2 fixes in `357d93e`:

- **Important ("Submit Another" stale position):** confirmed regression — `resetForm()` neither cleared `reportedAltitudeFt` nor re-requested a fix, and the one-shot `requestLocation()` auto-stops after one delivery, so subsequent PIREPs in a session reused the first report's lat/lon/altitude. Fixed both ends: `resetForm()` now clears `reportedAltitudeFt`, and "Submit Another" re-calls `trackingService.requestOneShotLocation()` (no-op while tracking). Design doc updated.
- **Minor (list "Add PIREP" not `coverage`-gated):** left intentionally ungated. Adding a coverage gate would reintroduce exactly the kind of gate this PR set out to remove, and the toolbar's `coverage == nil` check is incidental (the whole briefing toolbar is hidden before a pack loads), not a deliberate PIREP rule. A PIREP is a position report about *current* conditions, independent of which flight's menu opened it.

Thanks for the positive verification notes on the round-1 fixes. Still pending a macOS compile (no Swift toolchain in this session).

---

roznet:
Round-3 fixes in `771d632` (both Criticals verified against the code first):

- **Critical (publish-only tab 403):** confirmed — `loadPireps()` ran on every briefing open and 403s for `can_view=false`, so `LoadingStateView` rendered its error screen and the "Report a PIREP" action (previously buried in the loaded-and-empty branch) was unreachable. Fixed: the container now skips `loadPireps()` unless `pirepCanView`, and `PirepListView` renders the add bar **outside** the load-state switch (present in every state) with a "filing enabled, viewing not yet" message instead of an error for publish-only accounts.
- **Critical (out-of-window orphaning):** confirmed — `submit()` sends no `pack_id`, so linkage depends entirely on the `observed_at` window (`storage/pireps.py`). Full `pack_id` plumbing would need a cross-stack server change (the iOS `PackMetaResponse` doesn't carry the numeric id, and `_meta_to_response` has 10+ call sites) that can't be iOS-compiled here — disproportionate. Instead the **flight-list "Add PIREP" is now gated on the flight's tracking window**, which is the entry point with zero temporal awareness and the failure scenario you described (long-pressing a months-away flight). The two in-briefing entries stay permanent; a report filed there far outside the window is retained as standalone community data — now documented in `ios-app-overview.md` / `ios-app-ui.md` rather than being silent.
- **Minor:** collapsed the duplicated `canPublish`/`onAdd` guard into the single `addBar` property.

Still not machine-compiled (no Swift toolchain in this session).

---

roznet:
Round-4 fixes in `d8995bf`:

- **Important (subscriber orphaning):** good catch — confirmed the linkage predicate is `observed_at`-in-window **AND** `user_id == flight.user_id` (owner). Since `pirep_can_publish` is account-level, a subscriber can file on a shared flight and orphan regardless of timing. Added `flight.isEditable` (owner) to the flight-list "Add PIREP" gate, and documented that in-briefing reports on a subscribed flight are always standalone.
- **Important (duplicated window formula):** hoisted the ±2h window into a single `FlightResponse.isInTrackingWindow(now:)`; `BriefingContainerView.isInFlightWindow` and the flight-list gate both delegate to it (also matches the server's window in `storage/pireps.py`).
- **Minor (missing test):** added `FlightTrackingWindowTests` pinning the boundaries — start edge, departure, end edge, and the two just-outside instants.

Thanks for the clean bill on the one-shot-location state machine / reset-and-refetch flow. As before, not machine-compiled here (no Swift toolchain), so the test is written to match the existing Swift Testing pattern but runs on your macOS build.

---

PR #461, opened 2026-07-20
ICON-D2 short-range upgrade: source the icon slot from ICON-D2 on domain + 48h fit

### LINKED ISSUE #456: ICON-D2 short-range upgrade: source the ICON slot from ICON-D2 when the whole route fits domain + 48h window
## Goal

When a route fits entirely inside the ICON-D2 domain **and** the flight window is within the D2 forecast horizon, source the ICON model's GRIB enrichment (full sounding replacement + cloud diagnostics) from **ICON-D2** (2.2 km, convection-permitting) instead of ICON-EU (6.5 km). Otherwise, use ICON-EU exactly as today.

This is an **in-place upgrade of the existing `icon` model slot** — NOT a 7th model, and NOT a mixed source:

- **All-or-nothing gate**: every route point inside the D2 domain AND flight-window end within the selected D2 run's horizon (48 h from init). If either fails → pure ICON-EU for the whole route. Never mix D2 and EU data within one briefing/cross-section (explicit product decision: keep each model column internally consistent from a single source).
- The existing ICON-EU whole-route domain check (`icon_eu_fetch.py`, bbox 29.5–70.5°N / 23.5°W–62.5°E, silently skip if outside) is the right shape — the D2 gate is the same test with a different bbox plus the horizon condition.

## Why it's cheap

ICON-D2 is essentially a config-level clone of the ICON-EU decode path:

- **Same server, same file layout**: `https://opendata.dwd.de/weather/nwp/icon-d2/grib/{HH}/{var}/…` with per-variable bz2 files, regular-lat-lon products.
- **Same variable set, same names**: model-level `t, qv, u, v, p, w, qc, qi, clc`; single-level `ceiling, hbas_con, htop_con, clcl/clcm/clch/clct, cape_ml/cin_ml, rain_con` (verify each exists on the D2 feed; expected to).
- **Same grid type** (regular lat/lon) → existing bilinear `_interpolate_per_point` works unchanged.
- **Same machinery reused whole**: log-pressure interpolation from model levels via `P` (`icon_eu_levels.py`), QV→RH derivation, W→omega (−ρ·g·w), chunked per-variable decode, hypsometric altitude fallback, `NWPCloudDiagnostics` mapping, rain_con de-accumulation with the prepended previous step, fill semantics (forward-fill diagnostics, linear-interp soundings).
- **Files are the same size as ICON-EU** (verified 2026-07-20: model-level T f006 = 1.01 MB for D2 vs 0.99 MB for EU; ceiling 0.69 vs 1.15 MB). The 2.2 km grid over the small domain has ~906 k points ≈ ICON-EU's ~905 k over Europe. No new download-volume, memory, or decode-time concern.

## What actually differs (the work)

1. **Domain bbox**: ~43.18–58.08°N, 3.94°W–20.34°E (verify exact corners from a decoded D2 file or DWD docs before hardcoding — same style as the ICON-EU constant). Covers Germany, Alps, Benelux, most of France, SE England; excludes Brittany edge cases, Scotland, Spain — partial-coverage routes are common, which is exactly why the gate is all-or-nothing.
2. **Levels**: ICON-D2 has **65 model levels** (bottom = 65) vs ICON-EU's 74 (we fetch 35–74). Need a D2-specific aviation-range level slice covering the equivalent altitude band (surface to ~FL250+), analogous to the EU 35–74 table. Check DWD's D2 level heights when picking the range.
3. **Filename casing/format quirks**: observed D2 filenames use lowercase variable suffix (`…_60_t.grib2.bz2` vs EU's `…_60_T.grib2.bz2`) and single-level files carry a `_2d_` segment (`…_006_2d_ceiling.grib2.bz2`). Parametrize the URL builder rather than assuming EU conventions.
4. **Cycles/horizon**: 8 runs/day (00–21z every 3 h), hourly steps to **48 h** (vs EU's 120 h). Publication delay shorter than EU (~1–2 h — measure and set in the freshness registry). Data retention ~1 day, same as EU.
5. **Run/gate interaction**: horizon check must be against the *selected run*: `flight_window_end ≤ run_init + 48 h`. If the freshest complete D2 run doesn't cover the window but an older one would, prefer falling back to ICON-EU over using a stale D2 run (keep it simple; EU is always available as fallback).
6. **Wiring**:
   - `cache.py` `MODEL_TTL_SECONDS` entry + cache dir `data/.cache/grib/icon-d2/{date}_{cycle}z/`.
   - `fetch/freshness/registry.py`: new source entry (e.g. `icon_d2:dwd`) with its schedule, so staleness/auto-refresh understand D2 cadence when it was used.
   - Pack metadata must record **which source was used** for the icon slot (`icon_eu` vs `icon_d2`) — `grib_init_times` key can stay `icon`, but add the source name so the freshness bar can badge it.
   - Freshness bar / model badge: show e.g. `ICON (D2)` when D2 supplied the run, so the upgrade is visible, not silent.
7. **Fallback robustness**: if the D2 download/decode fails mid-enrichment, fall back to ICON-EU for the whole route (never leave a half-enriched D2 pack) — mirror the existing graceful-degradation pattern.

## Out of scope

- Per-point D2/EU mixing or blending seams (explicitly rejected).
- D2's 15-minute sub-hourly output.
- ICON-D2-EPS ensemble.
- Any UI beyond the source badge.

## Testing

- Unit tests: domain gate (inside / one-point-outside / horizon-exceeded), level-slice mapping, URL builder for both model-level and single-level (`_2d_`) forms.
- Decode test against a downloaded sample D2 file (same style as existing ICON-EU decode tests).
- Integration: a Germany-internal route (e.g. EDDM→EDDH, departure < 48 h) enriches from D2 and badges it; the same route at D-4, and a route to Spain, both fall back to ICON-EU.

## References (in-repo context)

- `src/weatherbrief/fetch/grib/icon_eu_fetch.py`, `icon_eu_levels.py`, `decode.py` (`decode_icon_eu_per_point_chunked`), `fill.py`, `cache.py`
- `designs/weather-engine-specs.md` (§C DWD ICON-EU) — update it (and `fetch.md`) as part of this issue.

roznet:
## What & why

In-place upgrade of the existing `icon` model slot (issue #456). When a route fits **entirely** inside the ICON-D2 domain **and** the flight window is within a D2 run's 48h horizon, the ICON GRIB enrichment (full sounding replacement + cloud diagnostics) is sourced from **ICON-D2** (2.2 km, convection-permitting) instead of **ICON-EU** (6.5 km). Otherwise ICON-EU is used exactly as before. This is a source-variant of the same slot — **not** a 7th model, and **never** a per-point mix of D2 and EU within one briefing.

Since ICON-D2 is a config-level clone of the ICON-EU decode path (same server layout, same variable names, same regular-lat-lon grid type), the work is a parametrization rather than a copy:

- **`IconVariant` config object** (`ICON_EU` / `ICON_D2`) captures everything that differs — domain bbox, cycles, publication delay, model-level slice, filename conventions, cache slug, and freshness source key. Every `icon_eu_*` helper (URL builders, domain gate, run finder, window/level/horizon math) now takes `variant=ICON_EU`, so all existing callers and tests are unchanged.
- **D2 filename quirks** handled: model token `icon-d2`, region token `germany`, **lowercase** variable suffix (`…_60_t`), and a `_2d_` segment on single-level files (`…_006_2d_ceiling`).
- **All-or-nothing gate** in `_prepare_icon_eu`: every route point inside the D2 domain AND `flight_window_end ≤ selected_run_init + 48h`. It resolves the D2 run inline while gating, so it never picks D2 without a usable run; if the freshest complete D2 run doesn't cover the window it falls back to ICON-EU rather than a stale D2 run.
- **Fallback robustness**: on a total D2 failure (feed hiccup / decode error) the icon slot re-runs cleanly on ICON-EU — never a half-D2 pack.
- **Source attribution**: `enrich_forecasts` now returns `grib_sources` (model → source key used) so the pack records `model_sources["icon"]` = `icon_eu:dwd` or `icon_d2:dwd`; the freshness bar badges **`ICON (D2)`** when D2 supplied the run.
- **Wiring**: cache TTL (`icon-d2`, 6h), freshness registry entry (`icon_d2:dwd`) + readiness check (`icon_d2_dwd`), design-doc updates (`weather-engine-specs.md` §C.2, `fetch.md`).

**Note on two values that need live validation** (the sandbox has no route to `opendata.dwd.de`): the D2 domain corners and the model-level slice `25–65` are taken from the documented ICON-D2 `germany` regular-lat-lon product (1215×746 ≈ 906k points, matching the issue). `level_min=25` is a deliberately conservative top — the log-p interpolation clips any target level above the fetched column, so over-fetching a few thin upper levels is cheap insurance against truncating the sounding. Both are single well-commented constants; worth confirming against DWD's D2 HHL table once the feed is reachable.

## Issue linkage

Closes #456

## Testing / verification

- New `tests/test_icon_d2.py`: D2 URL builders (model-level lowercase + `_2d_` single-level), all-or-nothing domain gate (Germany-internal inside; Barcelona / Glasgow / Brest one-point-outside → EU), 48h horizon on every cycle, hourly temporal grid, out-of-range classification, `_prepare_icon_eu` variant selection (picks D2 on domain+run fit; falls back to EU when route outside D2 or D2 run uncovered; `force_variant` bypass), cache-TTL + registry + readiness-dispatch wiring.
- Updated `tests/test_freshness_registry.py` for the new `icon_d2:dwd` source.
- ICON-EU back-compat preserved: existing helpers keep `variant=ICON_EU` defaults; existing EU URL/domain/window tests are untouched.
- Verified the pure-logic modules (URL builders, domain/horizon/window math, registry config) standalone in-sandbox; frontend `briefing-ui.ts` typechecks and bundles. **The full pytest suite could not run in-sandbox** (the private `euro-aip` dep requires Python ≥3.12; sandbox is 3.11) — the integration-level tests are written to run in CI.

---
_Generated by [Claude Code](https://claude.ai/code/session_01EmbDUYzFjU7jtVbEoNsrqo)_

### DIFF
--- a/designs/fetch.md
+++ b/designs/fetch.md
@@ -192,7 +192,7 @@ if not status.fresh:
     schedule_refresh()
 ```
 
-Nine tracked source/model pairs (`SOURCE_REGISTRY` in `fetch/freshness/registry.py`): 3 direct GRIB (`ecmwf:direct`, `gfs:noaa`, `icon_eu:dwd`) + 6 Open-Meteo republishes (`gfs/ecmwf/icon/meteofrance/ukmo/gem:openmeteo`). Each `SourceConfig` carries schedule (cycles, delivery_offset, horizon) plus descriptive metadata (model/provider label, role, resolution, coverage, pressure_levels) feeding `/api/data-sources` and the help-page table.
+Ten tracked source/model pairs (`SOURCE_REGISTRY` in `fetch/freshness/registry.py`): 4 direct GRIB (`ecmwf:direct`, `gfs:noaa`, `icon_eu:dwd`, `icon_d2:dwd`) + 6 Open-Meteo republishes (`gfs/ecmwf/icon/meteofrance/ukmo/gem:openmeteo`). Each `SourceConfig` carries schedule (cycles, delivery_offset, horizon) plus descriptive metadata (model/provider label, role, resolution, coverage, pressure_levels) feeding `/api/data-sources` and the help-page table.
 
 The legacy `fetch/model_status.py` module (`fetch_model_metadata`, `check_freshness`, `compute_next_update`) is no longer consumed by the freshness endpoint's per-request path. `fetch_model_metadata` survives as a direct Open-Meteo init-time probe: it backs the Open-Meteo marker population in `fetch/freshness/sources.py` (background loop, not per-call) and is called directly by `tasks/alternates.py`, `tasks/standalone_verification.py`, `hewson/precompute.py`, `frontal/` (grid + CLI), and pack building in `api/packs.py`.
 
@@ -219,7 +219,7 @@ grib_init_times, grib_skip_reasons = enrich_forecasts(
 |--------|---------|
 | `gfs_idx.py` | Parse `.idx` files, plan HTTP byte ranges for CLMR/ICMR and cloud diagnostic variables |
 | `grib_fetch.py` | Find latest GFS run, bracket forecast hours, download via HTTP Range from S3 |
-| `icon_eu_fetch.py` | Find latest ICON-EU run, download model-level (QC/QI/P) and single-level diagnostics from DWD. Per-variable download for memory-efficient chunked decode |
+| `icon_eu_fetch.py` | Find latest DWD ICON run + download model-level (QC/QI/P) and single-level diagnostics. Parametrized by `IconVariant` (`ICON_EU` / `ICON_D2`) — one code path, two variants differing in domain, cycles, horizon, level slice, filename conventions, cache slug and freshness source key (#456). Per-variable download for memory-efficient chunked decode |
 | `icon_eu_levels.py` | Log-pressure interpolation from ICON-EU model levels to pressure levels |
 | `ecmwf_fetch.py` | Parse ECPDS filenames, scan delivery directory, find latest run. No HTTP — files land on local disk via ECPDS push |
 | `decode.py` | cfgrib → xarray decode, bilinear interpolation to route points. Chunked ICON-EU decoder (`decode_icon_eu_per_point_chunked()`) processes one variable at a time with explicit `gc.collect()` between — peak ~270MB vs ~800MB. ECMWF decoders handle multi-grid files (first-wins per point) |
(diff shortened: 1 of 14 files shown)


roznet:
Review follow-up pushed in 7b38ca74 (both findings verified against the live DWD feed, which the implementation sandbox couldn't reach):

1. **Level slice 25–65 → 16–65.** D2 level numbering isn't comparable to ICON-EU's — decoded DWD's HHL height field: D2 level 25 ≈ 6,300 m (FL207), which would have truncated every D2 sounding at ~20,000 ft; level 16 ≈ 9,460 m (FL310) matches ICON-EU's level-35 (~300 hPa) top.
2. **Variant-specific single-level diagnostics.** D2 has no deep-convection scheme: `hbas_con`/`htop_con` 404 on the feed, and `rain_con` is near-zero in explicit storms (rain lands in `rain_gsp`/`prg_gsp`). All three dropped for D2 → `convective_base/top_ft` and `convective_precip_mm_h` are cleanly **None** (missing-data semantics; `icon_eu_conv_rain_rate_mm_h` already returns None-not-0). The rain_con lead-step prefetch is skipped for variants that don't carry it.

Convection-permitting replacements (dbz_cmax, echotop, lpi, uh_max, prg_gsp — all verified live with units/stepType) are specced in follow-up #462.

Tests: 211 passed locally (test_icon_d2, test_freshness_registry, full test_grib) against main's venv.

Issue #442, opened 2026-07-16
Convective advisory: replace the DD thermo floor with an NWP-native grade + DD-trigger amber cap


roznet:
## Problem

The convective advisory currently grades each model at `graded_risk = max(NWP-scheme tier, DD-thermo tier)` — a full **DD floor** (`convective.py`, `floored_by_thermo`). In practice the DD/CAPE thermo track is conservative, so a loaded-but-uncorroborated environment where the model's own convective scheme is quiet can still be floored all the way to **RED** on DD alone. That's the main false-alarm source today (see the EGTF→LFAT→LFQA 2026-07-17 investigation: GFS graded amber purely off DD MODERATE while its NWP scheme was quiet; with a deep-but-dry tower, DD could push red).

We don't want to simply drop DD either — a genuine "DD loaded, NWP green" divergence must not go unnoticed.

## Decision (approved)

Grade the color from the **NWP-native tier only** (no full floor), with **one** asymmetric exception:

> If the NWP tier is **green** (below `min_risk`) **and** the DD-vs-NWP cross-check returns `dd_not_corroborated` (DD MODERATE+ while the model scheme is quiet), **upgrade to AMBER only** — never red — with `reason_code = "dd_trigger"`. Cap the displayed peak at MODERATE for these DD-only ambers.

New invariant: **RED ⟺ the model's own NWP scheme fires HIGH.** DD can only ever raise green→amber.

The amber trigger is coupled to the *exact* condition that already emits the `convective_cross_check` note, so color, note, and reason can never diverge.

### Truth table (2026-07-17 test flight)

| Model | NWP | DD | Today | After |
|---|---|---|---|---|
| ECMWF | HIGH (wet) | LOW | red | **red** (own scheme) |
| ICON | MODERATE | MODERATE | amber | **amber** (NWP) |
| GFS | quiet | MODERATE | amber (DD floor) | **amber**, reason = DD not corroborated |

## Consequences (accepted)

- **Reverses the advisory-level safety asymmetry** in meteorology-decisions §4 (reasoning 2) — a quiet-NWP + loaded-DD point is now AMBER, not RED. DD's voice is preserved as amber + the cross-check note, not silenced. Documented in the new meteorology-decisions §18.
- **NWP-amber + DD-HIGH stays amber** (the upgrade only lifts *green* NWP). Consistent with "red ⟺ NWP HIGH".
[...]

PR #464, opened 2026-07-21
ICON-D2 explicit-convection track for convective advisories (#462)

### LINKED ISSUE #462: ICON-D2 native convective diagnostics: reflectivity, echo top, LPI, updraft helicity for the icon slot (follow-up to #456)
## Context

Follow-up to #456 / PR #461 (merged). ICON-D2 is convection-permitting: it runs **no deep-convection parameterization**, so the parameterized-scheme diagnostics we consume from ICON-EU either don't exist on the D2 feed or silently change meaning. #461 therefore ships D2 with `hbas_con`/`htop_con`/`rain_con` deliberately unfetched — `convective_base/top_ft` and `convective_precip_mm_h` are **None** on D2-sourced packs.

This issue builds the replacement: an **explicit-convection track** that makes D2 the highest-resolution deterministic explicit-convection signal in the stack (excellent structural detail; exact cell placement grows uncertain with lead time — coarse models stay valuable for synoptic environment and timing). Deep convection in D2 lives in explicit storm fields (reflectivity, echo tops, lightning potential, updrafts, graupel) — a different *kind* of signal than the parameterized diagnostics, and the design below keeps that distinction explicit end-to-end.

Spec revised 2026-07-21 across two external (Codex) review rounds — field semantics, sub-hourly message structure, corridor sampling, completeness, handoff, and the v1 decision table below supersede earlier drafts.

## Verified field inventory

All verified against `opendata.dwd.de` (HEAD + cfgrib decode of live `_2d_` single-level messages, 2026-07-21). D2 filename convention per #461: lowercase var, `_2d_` segment.

| Variable | Decoded metadata | Meaning (per DWD ICON DB manual) | Role |
|---|---|---|---|
| `dbz_ctmax` | dBZ, stepType=**max** | Column-max simulated reflectivity, **max over previous hour** | **Required core** — the firing signal |
| `dbz_cmax` | dBZ, stepType=instant, floor −150 | Column-max reflectivity, instantaneous | Optional — pair with echotop if needed |
| `echotop` | shortName `min_pres`, **Pa**, stepType=**min**, sentinel −999; **4×15-min messages per file** (see below) | Lowest pressure (= highest altitude) where reflectivity > **18 dBZ**, per 15-min window | Depth/character AFTER firing — **never** cloud-top geometry |
| `lpi_max` | J/kg, stepType=max | Lightning Potential Index, max over previous hour | Electrification corroborator |
| `w_ctmax` | m/s, stepType=max | Max updraft (0–10 km) over previous hour | **Required** storm-strength corroborator |
| `uh_max` | m²/s², stepType=max, **signed** | Updraft helicity 2–8 km AGL, max amplitude over previous hour | Narrative/character only in v1 (rotation signal) |
| `grau_gsp` | kg/m² ≡ mm, stepType=**accum** (since init); on-the-hour message + 3 quarter-hour messages per file | Accumulated grid-scale graupel → **hourly graupel accumulation** after de-accumulation (like ICON-EU `rain_con`) | Mixed-phase-core corroborator |
| `tcond10_mx` | available | Column condensate **above the −10 °C isotherm** (the "10" is −10 °C, NOT 10 g/kg), interval max | **Deferred from v1** — no payload field; revisit with calibration |
| `cape_ml` / `cin_ml` | fetched since #461 | Mixed-layer CAPE/CIN, instantaneous | Unchanged — same meaning as EU |

Not usable: `hbas_con`/`htop_con` (404), `rain_con` (≈0 in explicit storms), `echotopinm` (404 — only the pressure form exists), `prr_con` (404). Prefer accumulated `grau_gsp` over instantaneous `prg_gsp` (a brief core between hourly samples would be missed).

### Sub-hourly message structure — VERIFIED (eccodes inspection of live f012 files, 2026-07-21)

D2 single-level convective files contain **multiple GRIB messages per file**, and the decoder MUST select/aggregate by `stepRange`/`validityTime` — never trust cfgrib to merge four valid-times into one field, and never let a concatenated blob double-count:

- `echotop` fXXX: **four** 15-min `min_pres` messages — at f012: stepRanges `705-720m`, `720-735m`, `735-750m`, `750-765m` (valid :00, :15, :30, :45). The message ending on the hour covers only the FINAL 15 minutes of `(H−1, H]`.
- `dbz_cmax`: four instantaneous quarter-hour messages (if fetched at all).
- `grau_gsp`: one on-the-hour since-init accumulation (`11-12`-style hour boundary: `0-12`) **plus** three quarter-hour since-init accumulations (`0-735m` etc.) — de-accumulation must use the **on-the-hour** messages only.
- `dbz_ctmax`: a single full-hour message (`stepRange 11-12`) — this is why it's the core field.

**Hourly echo top** must therefore be CONSTRUCTED, aligned with `dbz_ctmax`'s interval: for valid hour H, take the **minimum pressure across the four 15-min windows ending at H−45, H−30, H−15, H** — three messages from file f(H−1) + the first message from f(H). This is the highest 18-dBZ echo over `(H−1, H]`. Fetching f(H−1) for the first flight-window hour reuses the existing lead-file prepend pattern (rain_con predecessor). If any quarter is missing, degrade `echo_top` completeness — do not present a partial min as the hourly value.

## Hard semantic rules

1. **Echo top is not a cloud top.** An 18 dBZ echo top sits *below* the physical storm top (anvil ice reflects weakly). It must NOT feed the consumers `htop_con` fed — in particular the overfly-clearance filter in `analysis/advisories/convective.py` (`top_ft + clearance <= cruise_ft`), where it would err in the dangerous direction ("safe to overfly" under a higher anvil). Field name: `echo_top_18dbz_ft`. D2 cells render with **unresolved vertical geometry** for clearance purposes.
2. **Never linearly interpolate dBZ** (logarithmic). Interval-max fields are held over their interval, not interpolated; if instantaneous reflectivity ever needs blending, convert to linear Z first. Corridor-max extraction (below) makes this mostly moot.
3. **Interval maxima attach to `(valid_hour − 1, valid_hour]`.** stepType=max/min fields describe the *previous* hour, not the timestamp instant. Fill: hold over the covering interval (ECMWF `10fg` gust precedent, #441). `echotop` is the exception: each message covers only 15 minutes (verified) — never hold a single quarter across the hour; use the constructed hourly min described above.
4. **Graupel ≠ hail.** Wording everywhere: "graupel / strong mixed-phase core", never "hail".
5. **None ≠ 0** (house rule, #421): a channel that failed to fetch/decode is unknown, and unknown never suppresses or fakes evidence.

## Payload model

Nested explicit-convection object — do NOT add these to the parameterized fields on `NWPCloudDiagnostics`:

```python
class NWPExplicitConvectiveDiagnostics(BaseModel):
    source: Literal["icon_d2"]
    reflectivity_hour_max_dbz: float | None = None   # dbz_ctmax, corridor max
    reflectivity_instant_dbz: float | None = None    # dbz_cmax (optional)
    echo_top_18dbz_ft: float | None = None           # hourly min over 4 quarters, via column P→z
    lightning_potential_hour_max_jkg: float | None = None  # lpi_max
    updraft_hour_max_ms: float | None = None         # w_ctmax
    updraft_helicity_2_8km_hour_max_m2s2: float | None = None  # uh_max, SIGNED (see corridor rule)
    graupel_hour_mm: float | None = None             # grau_gsp hourly accumulation (de-accumulated)
    # Completeness is per-tier and means "valid (unmasked) corridor cells
    # decoded", NOT merely "file downloaded": an all-masked corridor is
    # UNAVAILABLE; a valid −150 dBZ floor is genuinely QUIET.
    detection_complete: bool = False   # dbz_ctmax valid over the corridor this hour
    strength_complete: bool = False    # lpi_max + w_ctmax + graupel all valid
    echo_top_complete: bool = False    # all 4 quarter windows present
```

`source` doubles as provenance for the advisory: the evaluator currently only sees model `"icon"` and has no access to `model_sources` — the presence of this payload IS the explicit-convection mode signal. Advisory modes: `parameterized_scheme` (EU path, unchanged) vs `explicit_convection` (payload present).

## Corridor-maximum extraction (new decode path)

Per-point bilinear sampling defeats a 2.2 km model: a storm cell between 10 NM route points — or displaced a few km by timing error — vanishes, making D2 *less* likely to fire than coarse models. For the extrema fields:

- Extract **max within a route buffer** (configurable, start ~10 NM kernel) at the grid level: `dbz_ctmax`, `lpi_max`, `w_ctmax`, graupel; for UH select **`argmax(abs(uh_max))` but store the signed value** (keeps the payload's signed promise); **min valid pressure** (highest echo) for `echotop` in the same buffer.
- This is a different code path from `_interpolate_per_point` — decode keeps the 2D field and reduces over a neighborhood mask per route point. Do NOT reuse `analysis/spatial_interpolation.py` unchanged for these extrema.
- Record in the payload/docs that values are **corridor maxima**, not centreline point values.
- `cape_ml`/`cin_ml`/cloud covers keep normal per-point interpolation.

## Analysis → advisory handoff

The convective evaluator consumes per-point-hour assessments (via `SoundingAnalysis`), not raw diagnostics. Wiring:

- Payload lives beside the existing pattern: `HourlyForecast.explicit_convective_diagnostics` (sibling of `nwp_cloud_diagnostics`; exact name aligned at implementation).
- A separate **`assess_convective_explicit()`** turns it into the standard convective-assessment contract with `method="nwp_explicit"`, **`top_ft=None`** (so the existing overfly-clearance filter structurally CANNOT consume echo top as a cloud top), provenance (`source="icon_d2"`), and the completeness flags. `echo_top_18dbz_ft` travels as a character/detail field only.
- **Incomplete detection data** (`detection_complete=False`) produces an **"explicit assessment unavailable"** state — it must NOT become: NWP `NONE`; "scheme quiet" in the DD cross-check (#178); or a CAPE-only fallback presented as D2's explicit verdict. Other models and the independent DD track remain fully usable.

## Firing and severity — v1 decision table (calibration starting points, not physical constants)

Corroborator set C (each counts 1 when its channel is complete AND over threshold): `lpi_max ≥ 1 J/kg` · `w_ctmax ≥ 10 m/s` · `graupel_hour_mm ≥ 0.5` · `cape_ml ≥ 500 J/kg`. (`|uh_max| ≥ 25 m²/s²` is narrative/character only in v1 — HRRR UH thresholds are NOT portable, 2–8 km vs 2–5 km layer.) `lpi_max ≥ 5` counts as 2.

| Corridor `dbz_ctmax` | |C| = 0 | |C| = 1 | |C| ≥ 2 |
|---|---|---|---|
| < 35 dBZ | no fire | no fire | no fire (environment tracks unaffected) |
| 35–44 dBZ | **no fire** — detail note "echo present, likely stratiform/melting-band" | MARGINAL | MODERATE |
| 45–49 dBZ | MODERATE | MODERATE | HIGH |
| ≥ 50 dBZ | HIGH | HIGH | HIGH |

Rules: map tiers onto the existing convective-assessment vocabulary at implementation. **Incomplete corroborator channels never downgrade below the dbz-alone row and never upgrade** (|C| counts only complete channels). Zero on one channel never suppresses positive evidence on another. `detection_complete=False` → unavailable (handoff section), never "quiet". Severity narrative after firing draws on echo-top depth, `w_ctmax`, stronger LPI, graupel, and the rotation note. Log the table + rationale + rejected alternatives as a dated `designs/meteorology-decisions.md` entry before shipping; revisit thresholds against the validation cases. DD cross-check (#178) unchanged — this is the NWP side of the same reconciliation.

## Aggregation — OPEN QUESTION, decide in #442's framework

A lone D2 explicit cell against quiet coarse models aggregates GREEN under majority rules. Proposal on the table (from external review): high-confidence D2 explicit convection floors the aggregate at AMBER with wording like "ICON-D2 explicitly develops a cell; coarser models do not agree." This changes aggregation semantics → decide it inside the #442 grade rework (where DD-only-caps-amber was decided) with its own meteorology-decisions entry. Do not implement silently here; do make the per-model D2 RED visible in details regardless.

## Domain gate hardening (carried from #461 review)

The D2 regular-lat-lon files carry ~16.7% masked corners (native domain is not a lat/lon rectangle; measured 17% NaN on live decodes). A route can pass the #461 bbox gate yet clip masked cells. The delivered files are `gridType=regular_ll`, template 0, with **no rotated-pole metadata** (verified via eccodes) — so the pole cannot be read from the GRIB. Exact fix: a **cached validity mask** built once from an invariant regular-grid D2 field (e.g. HHL) or from any message's bitmap, and require the route's **entire corridor buffer** (not just the centreline) to lie in valid cells. Route fails → ICON-EU, same all-or-nothing rule. (Fallback alternative: hardcode DWD's documented rotated-pole constants and test the native rectangle in rotated coordinates — but the mask tests the actual delivered product and is preferred.)

## Wiring checklist

- Variant single-level list: add the new fields for D2 (`ICON_D2_CLOUD_DIAG_VARIABLES`).
- **Bump the D2 cloud-diag cache key** `ICON_D2_CLOUD_DIAG_V2` → `_V3` (adding variables changes blob content, not key — #421 precedent; warm caches would silently lack the new fields).
- `grau_gsp` de-accumulation: reuse the rain_con machinery, selecting the **on-the-hour** messages only. Generalize the lead-step gate (currently `"rain_con" in variant.cloud_diag_variables`) into a variant-level needs-predecessor flag — D2 now needs f(H−1) for BOTH graupel de-accumulation and the three echo-top quarters.
- Decode selects messages by `stepRange`/`validityTime` (see sub-hourly structure) — do not feed a multi-message blob to cfgrib and trust the merge.
- `echotop` Pa→ft via the same hour's decoded column (P→z exists in the sounding path); sentinels: `≤ −998` Pa and dbz `≤ −100` → None.
- Fill (`fill.py`): interval-max fields hold-over per rule 3; register spatial handling explicitly (corridor extraction happens at decode, so these fields SKIP generic spatial interpolation).
- Carried from #461 review: add the total-D2-failure → EU re-run test (mock `_decode_and_merge_icon_eu` → `(None, None)` first call, assert second call is `ICON_EU`); required-field completeness feeds the `detection_complete`/`strength_complete`/`echo_top_complete` flags.

## Validation

Replay + controls, with observed radar/lightning where available (spatial+temporal tolerance, not exact-point):

- Convective hits: 2026-06-27 EGTF→LFAT→LFQA (fully in D2 domain; NWP-early/DD-late split), 2026-06-21 LFMD→EGTF (in-domain legs). Same-config replay method (advisory behaviour diff).
- **Quiet controls:** stratiform-rain days with 30–40 dBZ bright-band (must NOT narrate thunderstorms), winter graupel-shower days (must not over-grade).
- Domain-edge/masked-corner routes; required-field download/decode failure cases (must say "unavailable", never GREEN).
- Semantics tests: interval maxima attach to `(h−1, h]`; hourly echo top is the min over exactly the four quarters ending in `(h−1, h]` (and degrades completeness when a quarter is missing — never a partial min presented as complete); graupel de-accumulation uses on-the-hour messages only; corridor extraction catches a synthetic cell placed between route points.
- Acceptance is NOT "fires at least as early as EU's rain_con track" alone — that rewards false alarms; hits AND the quiet controls must both pass.

## Out of scope

- ICON-D2-EPS probabilistic convection (future candidate — new signal class).
- 15-minute sub-hourly output for D-0 onset timing (future candidate).
- Radar observation blending; changes to the DD satellite/obs side.
- The #442 aggregation decision itself (flagged above, decided there).

## Sources

- Live feed verification + message decodes (units/stepType/sentinels in the table): 2026-07-21 against `https://opendata.dwd.de/weather/nwp/icon-d2/grib/00/` (HEAD sweep + cfgrib decode of one `_2d_` file per variable; HHL decode for level heights in #461; **eccodes per-message inspection of f012 echotop/dbz_ctmax/grau_gsp** confirming the sub-hourly message structure and `regular_ll` template-0 grid).
- DWD ICON database reference manual (field definitions: ECHOTOP 18 dBZ threshold, LPI formulation, UH 2–8 km layer, TCOND10 −10 °C definition): https://www.dwd.de/DWD/forschung/nwv/fepub/icon_database_main.pdf
- DWD lightning/LPI report: https://www.dwd.de/EN/ourservices/reports_on_icon/pdf_einzelbaende/2022_10.pdf
- External (Codex) spec review 2026-07-21 — origin of the echo-top clearance hazard, DBZ_CTMAX preference, LPI floor, corridor sampling, mask-aware gate, and validation-matrix items adopted above.
- Prior art in-repo: #441 (gust hold-over), #283 (native firing gate), #442 (grade framework), #178 (DD cross-check), #421 (None≠0, cache-key bump), `designs/meteorology-decisions.md` §18.

## Dependencies / ordering

- #461 is merged — this builds directly on `IconVariant` and the D2 fetch path.
- Independent of #442, but the firing gate must be phrased in whichever framework lands first; the aggregation floor is #442's call.

roznet:
## Summary
- Adds an ICON-D2 **explicit-convection** path (`dbz_ctmax`, LPI, updraft, graupel, constructed hourly echo top) with corridor-max sampling, so D2 storms are no longer invisible after #461 dropped parameterized `hbas_con`/`htop_con`/`rain_con`.
- Keeps echo top as character only (`top_ft=None` on `nwp_explicit`) so overfly clearance cannot treat 18 dBZ echo tops as cloud tops.
- Corroboration treats LPI as the primary storm-process signal (w/graupel substitute only when LPI is absent); CAPE is environment/narrative only so 35–44 dBZ bright-band + CAPE cannot fake a thunderstorm. Incomplete detection stays unavailable (not quiet, not CAPE-as-D2).

## Test plan
- [x] `pytest tests/test_icon_d2_explicit.py tests/test_icon_d2.py tests/test_grib_fill.py tests/test_convective.py`
- [ ] Replay in-domain convective hit (e.g. EGTF→LFAT→LFQA 2026-06-27) and confirm D2 fires with unresolved vertical geometry / ghost column
- [ ] Quiet control: stratiform 30–40 dBZ day does not narrate thunderstorms
- [ ] Domain-edge route falls back to ICON-EU via delivered mask / total-failure path
- [ ] Confirm aggregation of a lone D2 cell vs quiet coarse models is still intentional (#442 open)

Closes #462

Made with [Cursor](https://cursor.com)

### DIFF
--- a/designs/meteorology-decisions.md
+++ b/designs/meteorology-decisions.md
@@ -1958,3 +1958,46 @@ follow-up:
 - Watch for any all-quiet-NWP sounding that *did* produce observed convection
   (lightning/METAR-TS) — that is the case the removed floor was protecting, and
   the one that would argue for a narrower cap.
+
+---
+
+## 19. ICON-D2 explicit convection: storm-process vs environment corroborators (#462)
+
+**Date:** 2026-07-21
+**Status:** Implemented (#462).
+**Context:** After #461 sourced the `icon` slot from convection-permitting
+ICON-D2 (dropping parameterized `hbas_con`/`htop_con`/`rain_con`), #462 builds
+the replacement track from explicit storm fields (`dbz_ctmax`, `lpi_max`,
+`w_ctmax`, `grau_gsp`, constructed hourly `echotop`). The issue's v1 decision
+table counted LPI, updraft, graupel, **and** ML-CAPE as independent
+corroborators. Two meteorology problems:
+
+1. **LPI already embeds updraft² × graupel** (Lynn & Yair / DWD LPI). Counting
+   `lpi_max`, `w_ctmax`, and `grau_gsp` as three votes triple-counts one
+   mixed-phase updraft.
+2. **CAPE is environment, not storm process.** A 35–44 dBZ bright band over a
+   primed sounding would upgrade to MARGINAL on CAPE alone — exactly the
+   stratiform false-alarm the quiet-control matrix is meant to catch.
+
+### The decision
+
+- **Firing core:** corridor-max `dbz_ctmax` over `(valid_hour−1, valid_hour]`.
+- **Storm-process corroborators (|C|):** LPI is primary (≥1 → 1, ≥5 → 2). When
+  LPI is missing/below floor, `w_ctmax ≥ 10` and `graupel ≥ 0.5 mm` may each
+  *substitute* (count 1). When LPI is present, w/graupel are narrative detail
+  only — not additive.
+- **CAPE / UH:** narrative only after (or beside) a fire — never upgrade the
+  35–44 band alone. Wording: "graupel / mixed-phase core", never "hail".
+- **Echo top:** `echo_top_18dbz_ft` is character only; `top_ft=None` on
+  `method="nwp_explicit"` so overfly clearance cannot treat it as a cloud top.
+- **Incomplete detection:** assessment with `native_data_complete=False` stays
+  in the NWP slot (not CAPE fallback, not "scheme quiet" for #178/#442
+  cross-check).
+
+Rejected alternatives: (a) issue draft's four-way |C| including CAPE; (b)
+mapping `echotop` into `htop_con` consumers; (c) centreline bilinear sampling
+of 2.2 km extrema (cells between route points vanish).
+
+Aggregation when a lone D2 cell faces quiet coarse models remains an open
+#442 decision — do not silently majority-vote it away; make the per-model D2
+RED visible in details regardless.
--- a/src/weatherbrief/analysis/advisories/convective_character.py
+++ b/src/weatherbrief/analysis/advisories/convective_character.py
@@ -381,6 +381,11 @@ class ConvectiveCharacterEvaluator:
                         (showers is not None and showers >= showers_mm)
                         or (cover is not None and cover >= CHAR_COVER_REALIZED_PCT)
                         or has_geom
+                        or (
+                            nwp is not None
+                            and nwp.method == "nwp_explicit"
+                            and nwp.risk_level != ConvectiveRisk.NONE
+                        )
                     )
                     embedded = _point_embedded(sounding, cruise_ft)
                     # Below-base avoidability geometry (#298): is the layer from
--- a/src/weatherbrief/analysis/sounding/__init__.py
+++ b/src/weatherbrief/analysis/sounding/__init__.py
@@ -234,6 +234,7 @@ def analyze_sounding_lite(
         detect_cloud_layers,
     )
     from weatherbrief.analysis.sounding.convective import (
+        assess_convective_explicit,
         assess_convective_nwp,
         assess_convective_thermo,
     )
@@ -311,9 +312,16 @@ def analyze_sounding_lite(
     convective = assess_convective_thermo(
         indices, omega_700_pa_s=_omega_near_700(derived_levels)
     )
-    convective_nwp = assess_convective_nwp(
-        indices, hourly.nwp_cloud_diagnostics if hourly else None,
-    )
+    if hourly and hourly.explicit_convective_diagnostics is not None:
+        convective_nwp = assess_convective_explicit(
+            indices,
+            hourly.explicit_convective_diagnostics,
+            hourly.nwp_cloud_diagnostics,
+        )
+    else:
+        convective_nwp = assess_convective_nwp(
+            indices, hourly.nwp_cloud_diagnostics if hourly else None,
+        )
 
     # Compute ceiling from NWP-reclassified cloud layers
     sounding_ceiling_ft = compute_sounding_ceiling_ft(
--- a/src/weatherbrief/analysis/sounding/convective.py
+++ b/src/weatherbrief/analysis/sounding/convective.py
@@ -15,6 +15,7 @@ from weatherbrief.models import (
     ConvectiveRegime,
     ConvectiveRisk,
     NWPCloudDiagnostics,
+    NWPExplicitConvectiveDiagnostics,
     ThermodynamicIndices,
 )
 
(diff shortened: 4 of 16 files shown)


roznet:
Closing unmerged — #462 shipped via #463. Recording the comparison factually since all three PRs were independent implementations of the same issue.

## Why #463 was picked over this one

**1. Decode approach — the deciding factor.** #462 states explicitly: *"Decode selects messages by `stepRange`/`validityTime` — do not feed a multi-message blob to cfgrib and trust the merge."* This PR adds the 7 explicit variables to `ICON_D2_CLOUD_DIAG_VARIABLES`, and `fetch_icon_eu_single_level` concatenates every variable's file into one blob per fhour **in `as_completed` order**. That blob then goes to `cfgrib.open_datasets`, and fields are looked up by name (`{"dbz_ctmax"}`, `{"lpi_max"}`, `{"min_pres","echotop"}`, `{"tgrp","grau_gsp"}`).

The repo already documents why that's unsafe — `decode.py:134-141`: DWD's `RAIN_CON` decodes under cfgrib shortName **`crr`**, not `rain_con`, "so the key MUST be `crr` for the field to be picked up at all." DWD local params routinely decode as `unknown`. If `dbz_ctmax` does, this fails **silently**: `detection_complete=False` at every point, no exception, no log line — a convective-safety feature that quietly does nothing. Nothing in the diff tests the cfgrib selection path (`test_icon_d2_explicit.py` covers `_corridor_extrema`, `_deaccumulate_nonnegative_grid` and `_hourly_echo_min_pressure_grid` on synthetic numpy arrays only), and the PR's own test plan leaves the replay and quiet-control boxes unchecked.

Second-order: putting 7 multi-message variables into the *shared* cloud-diag blob also routes them through the existing `decode_icon_eu_cloud_diag_per_point`. If `cfgrib.open_datasets` chokes on that heterogeneous file, its `except` returns empty dicts and D2 loses ceiling/CLC/CAPE too — i.e. this regresses a working path to add a new one.

#463 uses per-variable cache blobs plus an eccodes message-level decode that parses `stepRange` as a string (`"705-720m"`), noting that reading `startStep`/`endStep` as ints truncates sub-hourly steps to hours. That's the detail the whole four-quarter echo-top construction hinges on.

**2. Incomplete detection grades green.** #462, hard rule: incomplete detection *"must NOT become NWP `NONE`."* Here `assess_convective_explicit` returns `ConvectiveAssessment(risk_level=NONE, native_data_complete=False)`. Traced downstream: `tasks/advise.py::_resolve_analyses` (L177-185) swaps that into the active `convective` slot and badges `convective_method_effective="nwp"`. Nothing outside `convective_cross_check` reads `native_data_complete`. So a D2 hour whose detection channel failed **grades GREEN and claims it graded on NWP** — the exact fake-quiet failure #462 exists to remove, and the opposite of this PR's own summary ("Incomplete detection stays unavailable (not quiet…)"). #463 and #465 both return `None`, which falls back to thermo and badges it honestly.

Same shape as the bot's `strength_complete` finding: flags computed, stored, never consumed.

**3. Forward-filling interval maxima.** `_fill_diag_hourly` hold-fills `explicit_convective_diagnostics` onto hours with no data, repeating an hour-max reflectivity forward. #462 rule 3 holds interval maxima only over *their covering interval*; a 1-hour max from a missing hour has no such interval (contrast the ECMWF `10fg` gust precedent, whose window spans the gap). #463 registers explicit no-fill skips in both `fill.py` and `spatial_interpolation.py`.

**4. Fail-closed mask gate + uncaught decode.** `route_in_icon_d2_product_mask` returns `False` on any download failure, so a 404 or transient hiccup on the hsurf URL silently disables D2 entirely. The bot's Critical stands: `decode_icon_d2_domain_coverage` has `try/finally` with no `except`, bytes are cached *before* validation, and neither `route_in_icon_d2_product_mask` nor `_prepare_icon_eu` catches — so one truncated download wedges every D2-domain briefing until the next run and takes GFS/ECMWF down with it, because `_prepare_icon_eu` runs ahead of the parallel fetch phase. #463 builds the mask from a delivered message's bitmap, caches it to disk as an invariant npz, and **fails open** to the bbox gate with the reasoning documented.

**5. `_severity_modifiers` can still say "hail".** `convective.py:102` emits `"high freezing level + CAPE: hail risk"` from freezing level + CAPE alone, and this PR passes it through unmodified on the explicit track — violating #462's graupel-≠-hail rule. (#465 has the same gap; #463 rephrases it and tests `severe_modifiers`.)

**6. Docs.** Only `meteorology-decisions.md` updated; `weather-engine-specs.md` §C.2 left self-contradicting (still lists `prg_gsp`, still calls 15-min sub-hourly output out of scope while this PR reconstructs exactly that). #463 updated 7 docs including `fetch.md`, the module doc for the new pipeline.

## What was better here, and what we took

**The corroborator argument is the best single idea across the three PRs, and it's right.** DWD's LPI is `∫ ε w² dz` weighted by mixed-phase content, so counting `lpi_max` + `w_ctmax` + `grau_gsp` as three independent votes triple-counts one updraft; and CAPE-as-corroborator lets a 35–44 dBZ bright band upgrade to MARGINAL on environment alone, which is the stratiform false alarm the quiet controls are meant to catch.

We did **not** take it silently — the v1 table survived two external review rounds on #462, so changing the firing algebra needs the validation matrix run under both schemes, not a swap at implementation time. Filed as **#466** with your proposal (LPI primary, w/graupel substitute only when LPI is absent, CAPE/UH narrative) as the candidate and the both-directions acceptance criteria.

Also good and noted, though not carried: the `ConvectiveCharacterEvaluator` realized-by-explicit-firing branch (#463 landed the equivalent, gated at MODERATE+ rather than any non-NONE), and the `nwp_explicit` doc-comment updates in `web/ts/`.

PR #465, opened 2026-07-21
feat(icon-d2): add explicit convection diagnostics

### LINKED ISSUE #462: ICON-D2 native convective diagnostics: reflectivity, echo top, LPI, updraft helicity for the icon slot (follow-up to #456)

roznet:
Closes #462. Follow-up to merged #461.

## Summary

- add a separate `NWPExplicitConvectiveDiagnostics` payload and `nwp_explicit` assessment path for ICON-D2; parameterized ICON-EU diagnostics remain unchanged
- fetch/decode hourly `dbz_ctmax`, optional `dbz_cmax`, four-quarter `echotop`, `lpi_max`, `w_ctmax`, signed `uh_max`, and de-accumulated `grau_gsp`
- reduce extrema over a 10 NM route corridor, with complete-cell checks at bitmap and outer-grid edges; gate D2 on the delivered HSurf validity mask
- align temporal semantics: select exact validity endpoints, build echo top from f(H-1)'s trailing three quarters plus fH's on-hour quarter, and difference graupel grid cells before corridor maximization
- convert echo pressure through the same point/hour D2 pressure-height column, but keep `top_ft=None` so an 18-dBZ echo top can never authorize overflight
- implement the accepted v1 reflectivity/corroborator decision table, incomplete-is-unavailable behavior, DD cross-check, and convective-character handoff
- retain existing cross-model aggregation; the proposed lone-D2 AMBER floor remains an explicit #442-framework question
- bump D2 diagnostics cache to V3 and add the total-D2-failure clean ICON-EU rerun test
- document the signal-class boundary, thresholds, rationale, rejected alternatives, and calibration obligation

## Meteorology notes

The implementation treats D2 as high-resolution deterministic storm-structure evidence, not as a drop-in replacement for a deep-convection scheme. Reflectivity supplies realization; LPI/updraft/graupel/ML-CAPE corroborate. Graupel is described as a mixed-phase-core signal, never hail. UH is narrative only in v1 because D2's signed 2–8 km product is not directly portable from HRRR 2–5 km thresholds. Coarser models retain their value for synoptic environment and timing, especially as exact cell placement loses skill with lead time.

## Verification

- `341 passed`: focused D2 fetch/decode, GRIB, convective assessment, and prefetch suite
- `305 passed`: D2 + advisory evaluator/provenance regression suite
- full suite: `3716 passed, 9 skipped, 17 deselected`
- process-pool tests rerun outside the macOS sandbox: `43 passed`
- the remaining Google OAuth redirect failure reproduces unchanged on `main` and is unrelated to this branch

Live DWD samples were also inspected during implementation to confirm the regular-lat/lon grid, message endpoints, sentinels, and accumulation semantics described in #462.


roznet:
Closing unmerged — #462 shipped via #463. Recording the comparison factually since all three PRs were independent implementations of the same issue. This was the strongest of the two alternatives: faithful to the spec, correct on the incomplete-detection semantics, and the docs were updated in lockstep. It lost on one axis, but that axis was the decisive one.

## Why #463 was picked over this one

**1. Decode approach — the deciding factor.** #462 states explicitly: *"Decode selects messages by `stepRange`/`validityTime` — do not feed a multi-message blob to cfgrib and trust the merge."* This PR adds the 7 explicit variables to `ICON_D2_CLOUD_DIAG_VARIABLES`, and `fetch_icon_eu_single_level` concatenates every variable's file into one blob per fhour **in `as_completed` order**. That blob goes to `cfgrib.open_datasets`, and fields are resolved by name (`{"dbz_ctmax"}`, `{"lpi_max"}`, `{"min_pres","echotop"}`, `{"tgrp","grau_gsp"}`).

The repo already documents why name resolution is unsafe here — `decode.py:134-141`: DWD's `RAIN_CON` decodes under cfgrib shortName **`crr`**, not `rain_con`, "so the key MUST be `crr` for the field to be picked up at all." DWD local params routinely decode as `unknown`. Your `_find_all`/`_select` is genuinely more robust than #464's first-match `_find` — it scans every matching variable across every dataset and selects the exact validity endpoint either way — but it still can't find a field cfgrib named `unknown`, and the failure mode is silent: `detection_complete=False` at every point, no exception, no log. The PR body confirms live samples were inspected for grid, message endpoints, sentinels and accumulation semantics — but not for the cfgrib shortNames the lookup depends on, and no test exercises the selection path (`test_icon_d2_explicit.py` covers `_corridor_extrema`, `_deaccumulate_nonnegative_grid`, `_hourly_echo_min_pressure_grid` and `_pressure_pa_to_column_altitude_ft` on synthetic arrays only).

Second-order: routing 7 multi-message variables through the *shared* cloud-diag blob also feeds them to the existing `decode_icon_eu_cloud_diag_per_point`. If `cfgrib.open_datasets` chokes on that heterogeneous file, its `except` returns empty dicts and D2 loses ceiling/CLC/CAPE as well — a regression to a working path in service of a new one.

#463 uses per-variable cache blobs plus an eccodes message-level decode that parses `stepRange` as a string (`"705-720m"`), documenting that reading `startStep`/`endStep` as ints truncates sub-hourly steps to hours. Its tests **encode real GRIB2 messages with eccodes** — minute-resolution step ranges, bitmaps, `stepType=max/min/accum` — and drive the actual decoder through them, including the case where a larger quarter-hour graupel accumulation sits in the same blob as the on-the-hour message and must be ignored.

**2. Fail-closed mask gate + uncaught decode.** `route_in_icon_d2_product_mask` returns `False` on any download failure, so a 404 or transient hiccup on the hsurf URL silently disables D2 entirely rather than degrading. And `decode_icon_d2_domain_coverage` has `try/finally` with no `except` — identical to #464, where the bot rated it Critical: bytes are cached *before* validation, neither the caller nor `_prepare_icon_eu` catches, and `_prepare_icon_eu` runs ahead of the parallel fetch phase, so one truncated download wedges every D2-domain briefing until the next run *and* takes GFS/ECMWF down with it. Your review only flagged the unbounded `_D2_DOMAIN_MASK_BYTES` cache, not the crash path. #463 builds the mask from a delivered message's bitmap, caches it to disk as an invariant npz (built once, ever — not per run), and **fails open** to the bbox gate with the reasoning written down.

**3. `_severity_modifiers` can still say "hail".** `convective.py:102` emits `"high freezing level + CAPE: hail risk"` from freezing level + CAPE alone, and this PR passes it through unmodified on the explicit track — violating #462's graupel-≠-hail rule. Reachable exactly where it matters (high freezing level + CAPE >1000 is plausible when dBZ reaches the ≥50 HIGH band). #463 rephrases it on the explicit track and asserts on `severe_modifiers`, not just drivers/suppressors.

**4. Corridor completeness is stricter than the payload contract.** `_corridor_extrema(require_complete=True)` returns `None` when *any* cell in the kernel is non-finite, so a corridor with one masked cell reads unavailable rather than "max over the valid cells". Defensible, and consistent with your corridor mask gate — but combined with the fail-closed gate it means domain-edge routes lose the channel twice over. #463 reduces over valid cells and marks `corridor_valid` from whether any exist, which matches #462's completeness wording ("valid (unmasked) corridor cells decoded").

Everything else was solid: the incomplete-detection→`None` semantics, the diff-then-reduce graupel ordering, the four-quarter echo-top construction, the `top_ft=None` structural guarantee, the outer-grid-envelope check in `_corridor_extrema` (a real edge #464 missed), and four design docs updated coherently.

## What was better here, and what we took

**Your echo-top Pa→ft conversion — taken and now on `main`** (7093e0b2). `_pressure_pa_to_column_altitude_ft` keeps a single datum per hour: log-pressure over the hour's own geopotential column, extrapolated along the nearest two levels when the echo sits above the aviation slice, instead of silently switching to ISA pressure altitude mid-field. #463 fell back to ISA whenever the target left the bracketed range, which means a metre-datum column and an ISA column can both appear within one route — a hundreds-of-feet inconsistency that makes echo tops incomparable between points. Ported into `_echo_top_pa_to_ft` with your extrapolation semantics, tests, and a meteorology-decisions §19 rule 1 + fetch.md note crediting the source.

The same commit picked up the zero-width-pressure-bracket guard your structure made obvious (that code runs inside the per-hour attach loop with no `try/except` above it, so a degenerate column would abort the entire enrichment pass, not just one hour).

Also noted, not carried: the `_find_all`/`_select` multi-dataset scan (moot under per-variable blobs), and the `ConvectiveCharacterEvaluator` realized-by-explicit-firing branch (#463 landed the equivalent, gated at MODERATE+ rather than any non-NONE).

Separately, #464's argument that LPI already embeds `w² × graupel` — so counting LPI, updraft and graupel as three votes triple-counts one updraft, and CAPE-as-corroborator lets a bright band upgrade on environment alone — was the best meteorological idea across the three PRs and is filed as **#466** for calibration against the #462 validation matrix. Your table follows the spec's four-way |C| exactly, which was the right call for an implementation PR; the recalibration belongs in its own issue.

PR #470, opened 2026-07-21
ICON-D2 fetch cost: per-level cache + ceiling-limited download + warm known flights (#469)

### LINKED ISSUE #469: ICON-D2 fetch cost: per-level GRIB cache + ceiling-limited download + warm known flights
ICON-D2 costs **2.6×** ICON-EU per forecast hour, and download is the dominant term in refresh wall time. This issue covers three changes, in dependency order: a per-level GRIB cache (prerequisite), a ceiling-limited fetch (uses it), and cache warming for flights we already know about (uses both).

All figures below are measured, not estimated — droplet→DWD, prod DB, and live HHL decodes on 2026-07-21.

## Measurements

**Volume**, per forecast hour, same route, same run (`20260721 09z`):

| | total | model-level | cloud diag | explicit conv |
|---|---|---|---|---|
| ICON-D2 | **732 MB** | 9 vars × 50 levels = 697 MB (95%) | 20 MB | 19 MB |
| ICON-EU | **278 MB** | 9 vars × 40 levels = 269 MB | 10 MB | — |

Per (variable, level, hour) file: **0.95 MB on the wire** (bz2), 1.548 MB decompressed as we cache it. D2 costs 2.07× EU *per level* at near-identical grid-point counts — a 2.2 km convection-permitting field simply doesn't compress like a 6.5 km one. This is structural; there is no configuration fix for it.

**Throughput**, droplet → opendata.dwd.de:

| parallel workers | MB/s |
|---|---|
| 4 | 17.3 |
| 8 (`MAX_DOWNLOAD_WORKERS` today) | 29.6 |
| **16** | **37.1** |
| 24 / 32 | 30.6 / 33.1 |

→ **14.5 s per forecast hour** today. A 5-hour flight window is **72 s** of pure D2 model-level download before anything else runs.

**Usage** (prod DB, 3,149 packs over 180 days = 17.5 briefings/day, ~7 distinct flights/day, 2.5 packs per flight):

| cruise altitude | share | | departure hour | share |
|---|---|---|---|---|
| ≤ 6,000 ft | 47.5% | | 05z–18z | **97.7%** |
| 6–10,000 | 40.5% | | night | 2.3% |
| 10–14,000 | 4.6% | | | |
| 14–18,000 | 4.6% | | **≤ 18,000 ft total** | **97.2%** |
| > 18,000 | **2.8%** | | | |

**Model-level heights** (decoded from DWD HHL, two points — these are terrain-following):

| level | at sea level | over Alps (terrain 9,700 ft) |
|---|---|---|
| 16 (current top) | 31,005 ft | 31,508 ft |
| **27** | **18,680 ft** | 20,549 ft |
| 30 | 16,096 ft | 18,524 ft |
| 38 | 10,354 ft | 14,592 ft |
| 65 (bottom) | 200 ft | 9,766 ft |

Note the binding constraint for a fixed MSL altitude is **low terrain, not high** — levels ride up over the Alps, so sea-level columns need the lower level index. **Level 27 is the domain-safe cut for 18,000 ft**: levels 27–65 = 39 of 50 = **22% less to download**.

## Phase 1 — per-level GRIB cache (prerequisite)

Today the cache key is `cache_key(fhour, "ICON_D2_T")` — **no level dimension**. All 50 levels are concatenated into one blob per (variable, hour).

That makes any partial-level fetch a **correctness hazard, not just an inefficiency**: write levels 27–65 into `f029_ICON_D2_T.grib2`, and the next briefing that needs FL250 calls `is_cached()`, gets `True`, and silently decodes a **truncated column**. Nothing downstream can detect it.

DWD already serves one file per (variable, level, hour) — `icon-d2_germany_regular-lat-lon_model-level_2026072109_029_16_t.grib2.bz2` — and `fetch_icon_eu_per_variable` already takes a `levels: list[int]`. So the fetch side needs no protocol change; only the cache layout does.

- Key becomes per level: `f029_ICON_D2_T_L27.grib2`
- Read path concatenates the requested levels (cheap — it is already concatenated bytes today)
- A briefing needing more levels than are cached fetches **only the missing files** — this is what makes the incremental top-up real
- File count: ~9,800 per warmed run. Shard by forecast hour if directory size becomes awkward
- Migration: follow the existing `legacy_ck` pattern (`{prefix}_QC_QI_P`), which already handles exactly this kind of cache-layout transition — treat a legacy whole-column blob as a hit for the full range only

Independently valuable even without phases 2–3: it makes any future level-range change safe.

## Phase 2 — ceiling-limited fetch

Derive the top model level from the flight's ceiling plus margin, per route, using the elevation profile we already fetch (levels are terrain-following, so a fixed level index is not a fixed altitude).

**Hard constraint: keep `t`, `qv`, `p` at full column.** MetPy CAPE is the buoyancy integral to the equilibrium level — the live ESMX→EKRK case had EL at **FL306** with cruise at 6,000 ft, and D2 echo tops reached **FL365**. Truncating those three variables at the flight ceiling would systematically under-compute CAPE and cap tower tops, degrading the convective track that #462/#466/#467 are actively working on. (`cape_ml`/`cin_ml` are single-level fields, so the *NWP* CAPE is unaffected — it is the sounding-derived one that would break.)

Limit only `u, v, w, qc, qi, clc` → ~19% saving; limiting all nine would be ~22% but is not safe.

Covers 97.2% of flights at ≤18,000 ft; the 2.8% above simply fetch the extra levels via the phase-1 top-up.

## Phase 3 — warm the cache for known flights

Not a speculative calendar precache. **Rejected: precaching the daylight band broadly.** Sized at 9.3 GB wire / 15.2 GB disk per run for a daylight, ≤18k window (28 forecast hours × 39 levels) → 37 GB/day at 4 runs, 75 GB/day at 8. Against ~11 D2 briefings/day each reading ~5 forecast hours, that is **2–4× more bandwidth than on-demand fetching (~12–18 GB/day)**, with roughly a 20% hit rate, taken from DWD's free public service. It only becomes worthwhile if D2 briefing volume grows by about an order of magnitude.

Instead, warm what we know will be asked for. Flights are in the DB ahead of time — **36% are created more than 48h out**, and each is briefed ~2.5 times. After each run publishes, warm the cache for flights departing in the next 24–48h, using their **actual routes, window hours, and their own ceilings** (so it composes with phase 2).

- Near-100% hit rate instead of ~20%
- A few GB/day rather than 37–75
- Naturally concentrates on Thu–Sat, serving the weekend departure peak, with no weekday logic
- Timing: aim to land before ~06z — **30% of all packs are generated 04z–08z** (peak 07z, 11.2%)

There is an existing pattern to follow in `precache.py`, which already pre-warms ICON-EU/GFS for the maps.

## Also worth doing (small, independent)

Raise `MAX_DOWNLOAD_WORKERS` from 8 → 16: measured 29.6 → 37.1 MB/s, ~25% for one constant. Caveat: it is shared across concurrent briefings, so a blanket raise multiplies open connections under load — wants a concurrency-aware cap and a sustained-load test (the samples above were sub-second).

## Rejected alternatives

- **Thinning the near-surface levels** (~15%): the boundary layer is where ceiling and cloud base come from, and `clc`/`qc`/`qi` are sharply layered there. Not worth risking a primary product.
- **Cutting the top of the column**: already binding. The target pressure list runs to 150 hPa but D2 truncates at ~300 hPa (level 16 = FL310, matched to ICON-EU's level 35 ≈ FL300 by design in #461). Measured convective tops in the live case: EL FL306, EU towers FL320, D2 echo tops FL365. `weather-engine-specs.md` already records that the original guess of level 25 (FL206) "would have truncated every D2 sounding".
- **Day-of-week special casing**: briefing load is flat (13.1%–15.1% Mon–Sun; Fri–Sun only 5% busier per day). Departures *do* skew weekend (Sat 19.0% vs Wed 11.2%) but 2.5 briefings per flight smear that peak across the preceding week.
- **Spatial subsetting**: DWD serves whole-domain bz2 files with no `.idx`/byte-range equivalent to GFS, so we download 906k grid points to use ~25 route points regardless.

## Constraints

- Prod disk: `/mnt/flyfun_data` 149 GB, **55 GB free** (icon-eu cache is already 18 GB, gfs 5 GB). A full 48h D2 run would be 33.3 GB decompressed — one run fits, two do not.
- DWD open data is a free public service; sustained tens of GB/day is worth avoiding where a targeted alternative exists.

Related: #462 (the explicit-convection track whose CAPE/EL depend on the full column), #467 (validation — the ESMX→EKRK figures quoted here come from that run).

roznet:
## What & why

ICON-D2 costs 2.6× ICON-EU per forecast hour and download is the dominant term in refresh wall time. This implements the three dependency-ordered changes from #469, one commit per phase, plus the small worker-count bump.

**Phase 1 — per-level GRIB cache (`per_level_cache=True`, D2 only).** The model-level sounding is now cached one file per (variable, level) — `f029_ICON_D2_T_L27.grib2` — instead of one whole-column blob per variable. This is a *correctness* prerequisite, not just efficiency: a whole-column cache key can't tell a complete column from a partial one, so a partial-level fetch would silently serve a truncated sounding to the next briefing. Per level, `is_cached` answers per level, so a deeper-column briefing tops up only the missing files. Whole-column blobs written by older code (`{prefix}_{VAR}` and the even-older `{prefix}_QC_QI_P`) stay valid as full-range hits — no cache flush. ICON-EU keeps its whole-column blob unchanged.

**Phase 2 — ceiling-limited fetch.** The wind/cloud variables (`u, v, w, qc, qi, clc`) are fetched only down to a domain-safe model-level cut derived from `flight_ceiling_ft` (~19% of a D2 run's volume at FL180). The cut maps ceiling → top level from measured DWD HHL anchors (level 27 = 18,680 ft, 30 = 16,096, 38 = 10,354); model levels ride higher over terrain, so the sea-level column is the binding case and a sea-level cut covers the ceiling everywhere in the domain. **Hard constraint honoured:** `t, qv, p` stay full column so MetPy CAPE (buoyancy integral to an equilibrium level far above the flight — live ESMX→EKRK had EL FL306 at 6,000 ft cruise, echo tops FL365) is untouched. Fail-safe: unknown/high ceiling or ICON-EU → full column. Composes with phase 1.

**Phase 3 — warm the cache for known flights.** After each D2 run publishes, `precache_icon_d2_flights` warms the D2 cache for flights already in the DB departing within the next 48h, reusing the briefing prepare+prefetch path verbatim so the warmed cache is byte-for-byte what a briefing reads. Near-100% hit rate at a few GB/day — the targeted alternative to the broad daylight precache the issue rejects (37–75 GB/day, ~20% hit rate). Wired into `run_grib_precache_loop` on every fresh `icon_d2:dwd` marker; inherits the `WB_GRIB_PRECACHE_ENABLED` gate.

**Also (small, independent).** Default DWD download workers raised 8 → 16 (37.1 vs 29.6 MB/s measured, ~25%), env-tunable via `MAX_DOWNLOAD_WORKERS` as the concurrency-aware safety valve the issue asks for. The cold-cache prefetch already bounds connections via `_POOL_MAXSIZE`, so this only raises the default for single fetch calls.

## Issue linkage

Closes #469

## Testing / verification

- New unit tests: `test_icon_d2_per_level.py` (cache-key labels, per-level download primitive keying, prefetch job-planning incl. top-up and both legacy-blob migration paths, decode path building per-level lists), `test_icon_d2_ceiling.py` (domain-safe cut, per-variable level split, prefetch honouring it, `_prepare_icon_eu` wiring), plus flight-warming tests in `test_grib_precache.py` and worker-default tests in `test_icon_d2.py`.
- Full suite green on the changes: 3807 passed, 12 skipped. The one failure (`test_auth.py::test_google_login_redirects` → 404 without `GOOGLE_CLIENT_ID`) is a pre-existing environment/OAuth-config issue unrelated to this diff.
- Not exercised here (no DWD access / GRIB fixtures in the dev sandbox): the live download + cfgrib decode of real D2 GRIB. The pure planning/selection logic is unit-tested; the fetch/decode seams mirror existing patterns (`_d2_corridor_mask_ok`) and fail open to the full column.
- Design doc `weather-engine-specs.md` updated for all three phases + the worker bump.

https://claude.ai/code/session_01UCK2sXfFZRRrefUUZr85cu

---
_Generated by [Claude Code](https://claude.ai/code/session_01UCK2sXfFZRRrefUUZr85cu)_


roznet:
Fixed the Critical finding in `6309840`.

- Added `timezone` to the `datetime` import in `precache.py` — confirmed the bug: production always omits `now`, so `datetime.now(timezone.utc)` was raising `NameError`, swallowed by `run_grib_precache_loop`'s outer `except`, leaving phase-3 warming silently dead.
- Added a regression test (`test_default_now_resolves_without_nameerror`) that calls `precache_icon_d2_flights` **without** `now` and with a non-empty `db_path` (mocked empty DB), so the previously-broken line actually executes — the gap the existing tests had.

Nothing deferred; the review had no other findings. Pushed as-is (no rebase); the branch is cleanly mergeable onto current `main`.

---

PR #471, opened 2026-07-21
ICON-D2 explicit-convection track: reflectivity firing, corridor extrema, echo top detail-only (#462)

downle:
Independent implementation of **#462**, built from the pre-#462 base (`faa66f97`, which includes #461's `IconVariant` groundwork).

**Status note:** #463 has already merged an implementation of this issue, so this branch conflicts with `main` in the same feature area — it is offered as an **independent alternative implementation for review/comparison**, not intended to merge as-is. The Files-changed view (three-dot diff from the merge-base `faa66f97`) shows exactly this implementation and nothing else. It implements the issue's design plus the deltas from an independent meteorological review ([details below](#user-content-review-deltas-vs-the-issue-spec)), every one of them re-verified against the live DWD feed on 2026-07-21.

## What it does

D2 is convection-permitting (no deep-convection scheme), so the parameterized diagnostics consumed from ICON-EU (`hbas_con`/`htop_con`/`rain_con`) don't exist or change meaning on its feed. This builds the replacement: an **explicit-convection track** sourced from D2's storm fields, kept strictly separate from the parameterized track end-to-end.

- **Variables** (all re-verified live against `opendata.dwd.de`, eccodes per-message inspection of 00z files, 2026-07-21, + DWD ICON DB manual): `dbz_ctmax` (firing signal, hour-max), `echotop` (`min_pres` Pa, 4×15-min messages; hourly product constructed over the 4 quarters ending in (H−1, H]), `lpi_max`, `w_ctmax`, `grau_gsp` (per-cell de-accumulation, on-the-hour messages only), and **`uh_max_med`** (2–5 km).
- **New eccodes decode path**: messages selected by validity; multi-message storm fields never reach cfgrib's merge. **Corridor extrema** (~10 NM disc) replace per-point bilinear sampling; `dbz_ctmax` also keeps the centreline value as a cell-vs-shield discriminator.
- **Payload** `NWPExplicitConvectiveDiagnostics` attached to **every** D2 hour (always-attach: decode failure = all-None + flags False, never "quiet").
- **`assess_convective_explicit`**: dbz × corroborator decision table; **`top_ft` always None** so the 18 dBZ echo top structurally cannot reach the overfly-clearance filter; no CIN suppression (a simulated echo IS realized convection); lead-time confidence wording; graupel-never-hail wording; own DD-vs-explicit cross-check.
- **`detection_complete=False`** → `convective_explicit_unavailable`: the advisory renders **UNAVAILABLE** (excluded from extent math), never GREEN; grading falls back to thermo badged truthfully (`convective_method_effective="thermo"`).
- **Domain-gate hardening**: cached bitmap validity mask requires the whole corridor unmasked, else ICON-EU (all-or-nothing, #456 rule).
- **Fill**: interval-max hold-over — gap hours inherit the NEXT anchor's (covering) interval (10fg gust precedent, #441).
- **Cache**: D2 cloud-diag key V2→V3 (#421 precedent); predecessor-step gate generalized to a variant flag (`grau_gsp` + `echotop` now need the lead file).
- **Aggregation floor for a lone D2 cell deliberately left to #442**; the per-model D2 status is visible regardless. Full rationale: `designs/meteorology-decisions.md` §19.

## Review deltas vs the issue spec

1. **Sentinel discipline fix** — the spec's "dbz ≤ −100 → None" contradicts its own "valid −150 dBZ floor is genuinely QUIET": the GRIB **bitmap** alone marks missing (~16.7% of cells); −150 dBZ and −999 Pa are real quiet data. Completeness is bitmap-driven per channel, so quiet nights read QUIET, never UNAVAILABLE.
2. **`uh_max_med` (2–5 km) instead of `uh_max` (2–8 km)** — the 2–5 km layer IS on the feed and is the literature/HRRR-calibrated layer, so the "thresholds not portable" caveat disappears. Still narrative-only in v1.
3. **"Unavailable renders GREEN" gap closed** — the evaluator previously mapped a missing convective assessment to GREEN; explicit-unavailable now renders UNAVAILABLE and is excluded from the extent denominator.
4. **Lead-time confidence bands** — ICON-D2 is radar-nudged (LHN), so wording is stratified: ≤6 h "positions meaningful", 6–24 h "placement approximate", >24 h "environment-level signal". Wording only, never a tier input.
5. **Centreline-vs-corridor discriminator** — widespread shield (point ≈ max) vs discrete cell (max ≫ point) feeds the 35–44 dBZ stratiform note.
6. **Echo-top Pa→ft via ISA** — single reproducible datum for a corridor extremum whose cell is not under the centreline; documented as character-only.

## Verification

- Decode smoke-tested against real downloaded D2 blobs: corridor catches a 45 dBZ cell ~10 km off the centreline near Berlin that bilinear sampling reads as 27 dBZ; masked SW corner yields all-None + incomplete (UNAVAILABLE, not quiet); echo-top hourly min + ISA conversion (63,495 Pa → 12,375 ft) verified.
- **56 new tests** in `tests/test_icon_d2_explicit.py`: corridor math, sentinel rules, quarter construction/degradation, graupel de-accum (incl. reset-clip), builder completeness, decision table, assessment contract (`top_ft` never set), cross-check, evaluator UNAVAILABLE, fill semantics, mask gate, and the total-D2-failure → EU re-run test carried from #461's review.
- Full suite: **no regressions** (6 pre-existing failures on the base in `test_icon_prefetch`/`test_auth`, unrelated; one assertion updated for the intentional V2→V3 cache bump).

🤖 Generated with [Kimi Code](https://www.kimi.com/code) — co-authored on the commit via `Co-authored-by`.

### DIFF
--- a/designs/meteorology-decisions.md
+++ b/designs/meteorology-decisions.md
@@ -1958,3 +1958,126 @@ follow-up:
 - Watch for any all-quiet-NWP sounding that *did* produce observed convection
   (lightning/METAR-TS) — that is the case the removed floor was protecting, and
   the one that would argue for a narrower cap.
+
+## 19. ICON-D2 explicit-convection track: reflectivity firing, corridor extrema, echo top as detail-only
+
+**Date:** 2026-07-21
+**Status:** Implemented (#462).
+**Context:** Follow-up to the ICON-D2 source upgrade (#456). D2 runs **no
+deep-convection parameterization**, so the parameterized diagnostics consumed
+from ICON-EU (`hbas_con`/`htop_con`/`rain_con`) either don't exist on the D2
+feed or silently change meaning; D2 packs ship those fields as ``None``. Deep
+convection in D2 lives in **explicit storm fields** — a different *kind* of
+signal — so it gets its own payload, its own assessment
+(``method="nwp_explicit"``), and its own cross-check, never blended with the
+parameterized track. All feed semantics below were verified against
+`opendata.dwd.de` (eccodes per-message inspection of live 00z files) and the
+DWD ICON database manual on 2026-07-21.
+
+### The variable set (and what was rejected)
+
+- ``dbz_ctmax`` (column-max simulated reflectivity, hour-max) — the **firing
+  signal**. One full-hour message per file; attaches to ``(H−1, H]``.
+  Caveat accepted: DWD's forward operator is a Rayleigh approximation over
+  QR/QS/QG/T, so absolute dBZ is approximate (bright-band/graupel biases) —
+  handled by the 35–44 dBZ corroboration row, not by lowering the tier.
+- ``lpi_max`` (Lynn & Yair 2010; w² integrated over the graupel-bearing
+  layer), ``w_ctmax`` (max updraft 0–10 km), ``grau_gsp`` (grid-scale graupel,
+  de-accumulated **per cell** from on-the-hour accumulations) — the
+  corroborator set. LPI physically requires the graupel scheme, so LPI and
+  graupel are correlated corroborators — |C| is not "independent
+  confirmations", and is treated as such in calibration.
+- ``uh_max_med`` (updraft helicity 2–5 km) — **chosen over ``uh_max``
+  (2–8 km)** because 2–5 km is the literature/HRRR-calibrated layer, keeping
+  rotation notes comparable to portable thresholds. Narrative-only in v1
+  either way; signed argmax-|uh| keeps the rotation sense.
+- ``echotop`` (delivered as ``min_pres`` in Pa, 4×15-min messages per file) —
+  the hourly product is **constructed** as the min pressure over the four
+  quarter windows ending in ``(H−1, H]`` (three from file f(H−1), one from
+  f(H)). It is a **storm-depth indicator, never a cloud top**: an 18 dBZ echo
+  top sits below the physical storm top (weakly-reflecting anvil ice), so it
+  is structurally barred from ``top_ft`` (the assessment always sets
+  ``top_ft=None``) and from the overfly-clearance filter, where it would err
+  in the dangerous direction.
+- Rejected/deferred: ``dbz_cmax`` (instant; sub-hourly detail not needed in
+  v1), ``tcond10_mx`` (overlaps the dbz signal), ``rain_con`` (≈0 without the
+  deep scheme), ``hbas_con``/``htop_con``/``prr_con``/``echotopinm`` (404),
+  ``uh_max`` (superseded by ``uh_max_med``, above), ICON-D2-EPS (future
+  probabilistic signal class), radar blending (out of scope).
+
+### Sentinel discipline: bitmap is the only "unknown"
+
+The GRIB **bitmap** alone marks missing data (the masked domain corners,
+~16.7% of cells on every field). Encoded quiet values are real data: **−150
+dBZ** is the genuine no-echo floor and participates in the corridor max;
+**−999 Pa** is "no ≥18 dBZ echo this quarter" and counts as a valid cell
+(excluded only from the pressure min). Completeness flags are driven by
+bitmap-valid corridor fractions (≥ 50 %), never by decoded values — so a
+quiet night reads QUIET, never UNAVAILABLE, and a masked corridor reads
+UNAVAILABLE, never QUIET. This is the #421 None ≠ 0 rule applied at grid
+level.
+
+### Firing table (v1 calibration starting points, not physical constants)
+
+| corridor ``dbz_ctmax`` | \|C\| = 0 | \|C\| = 1 | \|C\| ≥ 2 |
+|---|---|---|---|
+| < 35 dBZ | no fire | no fire | no fire |
+| 35–44 dBZ | no fire (stratiform/bright-band note) | MARGINAL | MODERATE |
+| 45–49 dBZ | MODERATE | MODERATE | HIGH |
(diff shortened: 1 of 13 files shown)


roznet:
## Comparative review vs. merged #463 + main follow-ups (record-keeping)

We implemented #462 via **#463** (merged) plus six follow-up commits across four review rounds. This PR is an independent single-shot implementation (Kimi) of the same issue, kept for comparison. We are **not merging it** — closing for that reason. Recording the review here since the comparison was genuinely useful.

Compared against `main` as it stands today (#463 + `6c581f9`, `7093e0b`, `7f9c45c`, `0aee175`, `1291d06`, `573dbad`).

### What was strong

On the hard, verifiable core this implementation was at least as correct as the merged+reviewed result, in one shot, with accurate self-documenting rationale. Notably, it **pre-empted the two most dangerous bugs #463 shipped and only caught later on live routes**:

- **0-360 longitude convention (main `0aee175`, critical).** DWD delivers `longitudeOfFirstGridPointInDegrees` in the 0-360 convention (western edge 3.94W → 356.06). #463's new eccodes path read that key raw, so *every* route point failed the corridor gate and the D2 slot was silently never selected until the fix. This PR reads eccodes' computed `latitudes`/`longitudes` value-arrays (already signed, −3.94…20.34°E) and documents exactly why — never hits the trap.
- **Corridor clipped by the outer grid boundary (main `7f9c45c`).** #463 originally envelope-checked only the centreline and let `searchsorted` truncate at array bounds, so an edge route reduced extrema over a partial buffer yet reported `detection_complete=True`. This PR carries a `clipped` flag (`i0<0 or j0<0 or i1>n_rows or j1>n_cols`) and gates every completeness flag + the mask on `not clipped`.
- **Per-cell graupel de-accumulation (main `6c581f9`).** diff-the-whole-grid-per-cell then corridor-max (diff-then-reduce) — correct, and cleaner than #463's cell-map intersection.
- **No-hail wording, decision table, corroborator set** — equivalent to main (identical `dbz × |C|` tiers; CAPE corroborator-only so a 35-44 dBZ bright-band + CAPE can't fake a thunderstorm; `"hail risk"` rephrased to graupel/mixed-phase on the explicit path).
- Structurally sidesteps the degenerate-column crash main guarded in `7093e0b` (closed-form ISA echo-top conversion, no log-pressure bracket).

### Legitimate design divergences (not defects — reasonable disagreement)

- **Echo-top Pa→ft via ISA everywhere** vs main's per-hour geopotential column. Both avoid the mid-field datum-flipping #463 had; different single-datum choice. This PR's rationale (the echo cell is a corridor extremum, not under the centreline, so a centreline column borrows the wrong datum) is sound, though it diverges from `meteorology-decisions.md` §19 rule 1 as written.
- **UH on the 2-5 km layer** (`uh_max_med`, HRRR-portable) vs main's 2-8 km (`uh_max`, what the spec named but §19 flags as non-portable). Narrative-only on both; arguably the better meteorology here.
- Per-channel valid-fraction completeness (≥50%) vs main's all-or-nothing; hold-over fill implemented (main deliberately rejected it); plus narrative-only extras (cell-vs-shield discriminator, lead-time confidence bands).

### What's still missing / weaker than main

- **Fake-quiet on fetch failure (main `6c581f9`, the most consequential gap).** The "attach to every D2 hour" claim holds at the per-point builder and the explicit-sub-decode-failed path, but the explicit decode was folded into the *shared cloud-diag loop*, which `continue`s before attaching on (a) a fetch/cache miss of the diag file and (b) an empty standard decode (`if not decoded_points: continue` — throws away an already-built explicit payload). A one-hour fetch hiccup therefore still falls through to the parameterized path and reads structurally quiet on D2. main split this into a dedicated `_enrich_icon_d2_explicit_convective` with no early-out.
- **UNAVAILABLE hides a real hazard.** The evaluator renders explicit-unavailable points as UNAVAILABLE and drops them from the extent denominator, short-circuiting *before* the thermo fallback it also computes. If every route point is D2-unavailable, the whole advisory goes grey while discarding a usable CAPE/DD thermodynamic assessment. main instead grades those points on the real thermo track badged `convective_method_effective="thermo"` — the "DD/thermo track stays usable" principle.
- **Tests mock the GRIB reader, so they don't defend the bugs the code got right.** `_read_d2_explicit_messages` is replaced with hand-built messages and every fixture uses −180..+180 longitudes — so the suite would stay green even if the decode *had* the 0-360 or boundary-clip bug, and the graupel fixture (`prev=0.5`, `cur=3.0`) can't distinguish diff-then-reduce from the buggy order. main round-trips real GRIB2 bytes with explicit regressions for 0-360 (prime-meridian straddle), boundary clip, prefetch, and full f011/f012/f013 enrichment.
- **Boundary-clip still emits the partial-buffer extremum** (flagged incomplete) where main returns `None` — weaker guarantee if a consumer ever reads the value while ignoring the flag.
- **No design-doc sync.** main updated seven docs (§19 + INDEX/advisories/analysis/data-models/fetch/weather-engine-specs); this PR wrote only §19 and omits the 0-360 and boundary-clip history.

### Takeaway

Excellent on the meteorological/algorithmic spec — matched or beat the merged result on the risky decode math in a single pass. Weaker on end-to-end failure-mode reasoning (cache/loop coupling reintroducing fake-quiet; UNAVAILABLE suppressing a real signal) and on adversarial test rigor (mocking away the layer where the dangerous bugs actually live). That failure-mode/testing gap is roughly where the four rounds of human + bot review on #463 added their value.

Closing — implemented via #463.

---

PR #476, opened 2026-07-22
Fix GRIB cache bloat: age by init time, add size cap, gate warming window

### LINKED ISSUE #475: ICON-D2 pre-deploy: cache retention by init time, size-cap backstop, and a 03Z-21Z warming window
## Why this exists

ICON-D2 is **not yet deployed to prod** (prod is on `f21b0152`, 28 commits behind `main`; no `precache_icon_d2_flights`, zero `icon_d2` log lines in 24h, no `.cache/grib/icon-d2` directory). Before it ships, three cache-management changes need to land, because a pre-deploy sizing exercise showed the D2 flight warming introduced by #470 (issue #469 phase 3) will **fill the production disk** as currently written.

This issue collects everything we believe is needed before the D2 deploy.

## Measured facts (not estimates)

All numbers below were measured against real prod data and a real DWD fetch on 2026-07-22.

**Unit cost.** One ICON-D2 `(variable, level)` model-level file, fetched from `opendata.dwd.de`:
- 948,158 bytes downloaded (bz2), 1,623,253 bytes on disk (decompressed) — ratio 0.58

**Per forecast-hour** (full column, since #469 phase 2 is gated off — see #474):
- 9 vars x 50 levels (16-65) = **450 files ≈ 427 MB download / 730 MB disk**

**Prod flight load.** 73 future flights (1469 total). Driving the real `_prepare_icon_eu` for every D2 run publishing over a ~4.75-day window; 20-26 flights are D2-eligible per run, with a per-run union of 21-29 forecast-hours:

| | forecast-hours | download | per day |
|---|---|---|---|
| all 8 passes/day | 817 | 340 GB | **71.8 GB/day** |
| 03Z-21Z window | 618 | 257 GB | **54.3 GB/day** |
| saving | 199 | 83 GB | **24%** |

**Prod disk today:** `/mnt/flyfun_data` 149 GB total, 91 GB used, **52 GB free (64%)**. Existing cache: `icon-eu` 22 GB, `gfs` 6.0 GB, `icon-d2` absent.

**Peak retained D2 cache** (whole run dirs, expiring by TTL) — assuming age is measured correctly:
- TTL 6h -> **41.6 GB** (2 runs). TTL 4h -> also 41.6 GB (identical: D2 runs every 3h, so 4h and 6h both retain exactly 2 runs). TTL 3h -> 21.2 GB (1 run, but loses the prior-run fallback).

### Sizing note for the record

`designs/weather-engine-specs.md` currently describes flight warming as "a near-100% hit rate at a few GB/day", and justifies it as the cheap alternative to a broad daylight precache "sized at 37-75 GB/day, ~20% hit rate". The measured figure is **71.8 GB/day** — at the top of the range it was chosen to avoid, i.e. the original sizing was off by roughly 20x. It wasn't caught because the dev DB has 2 future flights while prod has 73.

The root cause is structural and not fixable here: DWD serves whole-domain files (1215x746) with no byte-range/`.idx` subsetting like GFS, so we download all of Germany to sample a 10 NM corridor. The only levers are fewer hours, fewer levels, fewer runs.

---

## 1. Age runs by init time, not directory mtime (BLOCKER)

`purge_old_runs` (`src/weatherbrief/fetch/grib/cache.py:161`) ages a run by `run_dir.stat().st_mtime`. A directory's mtime tracks file **creation/deletion**, not modification — verified empirically. We never rewrite a cache file (`is_cached` skips it), but we *do* keep adding **new** `(fhour, var, level)` files to an existing run dir: a later briefing wanting a forecast-hour nobody requested yet, a different route, or a phase-1 top-up for a deeper column. Every such write resets the directory's apparent age to zero.

**This is happening in prod today.** Comparing each `icon-eu` run dir's init (from its name) to its actual mtime:

| run | last new file written |
|---|---|
| `20260721_12z` | **8.9 h** after init |
| `20260721_15z` | 5.4 h after init |
| `20260721_18z` | 4.8 h after init |
| `20260722_00z` | 5.7 h after init |

So the 12z run is not purged at init+12h; it survives to roughly **init+21h**. The TTL is silently ~75% longer than configured.

For D2 at a 3h cadence, every extra 3h of retention adds a whole run (~21 GB):

| drift | effective retention | runs kept | peak disk | fits in 52 GB free? |
|---|---|---|---|---|
| 0 h | 6 h | 2 | 42 GB | yes |
| 2 h | 8 h | 2 | 42 GB | yes |
| **4 h** | 10 h | 3 | **64 GB** | **NO** |
| 6 h | 12 h | 4 | 85 GB | NO |
| 9 h | 15 h | 5 | 106 GB | NO |

If D2 drifts like ICON-EU does (~5h observed), peak lands at **64-85 GB against 52 GB free**.

**Fix.** The init time is already encoded in the directory name (`cache_dir_for_run` -> `{YYYYMMDD}_{HH}z`). Parse it and compute age from that rather than `st_mtime`. TTL then means what it says, and the 41.6 GB figure becomes real instead of aspirational. This also fixes ICON-EU's unintended ~21h retention as a side effect (expect a one-off drop in `icon-eu` cache size after deploy — that is the fix working, not data loss).

Keep a fallback to `st_mtime` for any directory whose name doesn't parse, so a stray/legacy directory is still eventually collected rather than pinned forever.

## 2. Size-cap backstop with oldest-run-first eviction

TTL alone cannot bound disk: retained size scales with whatever the flight set demands, and flight count grows (73 today). Add a hard bound *after* the TTL rule:

```
while total_size(model) > CAP[model]:
    drop the run dir with the oldest init
    stop if only FLOOR runs remain (default 2 — preserves current + prior fallback)
```

**Evict whole run directories, not individual files.** Per-file eviction is technically safe — phase 1's per-level cache means `is_cached` answers per level and a briefing transparently tops up whatever is missing — but it would be actively harmful: deleting half a run's levels to reclaim disk means the next briefing re-downloads them at ~0.95 MB each, trading bandwidth for disk in a system where bandwidth is already the expensive resource. Whole-run eviction avoids the thrash and preserves "a retained run is complete" as an invariant.

Note this is deliberately **not** LRU. Classic LRU evicts by access recency, which is the wrong utility signal for weather data — a 12h-old run is stale regardless of how recently something read it. Age is the right proxy.

Log every eviction (model, run, bytes reclaimed, resulting total) so the cap firing is visible rather than silent.

### Open decision: what caps?

The disk budget is genuinely tight and should be settled before deploy. Today's cache totals 28 GB (`icon-eu` 22 + `gfs` 6) with 52 GB free. D2 at TTL 6h wants ~42 GB, which leaves almost nothing spare — and `/mnt/flyfun_data` also carries pack data, the DB and logs, which grow independently.

Options, roughly in order of preference:
- Re-land #469 phase 2 (ceiling-limited fetch, tracked in #474) — removes ~15-19% of every run, taking the D2 peak from ~42 GB to ~34 GB. Real headroom rather than a policy patch, but it is blocked on its own consumer-safety work.
- Set a conservative D2 cap (~35-40 GB) and let eviction bite, accepting some re-fetch.
- Drop D2 TTL to 3h (~21 GB) — cheapest on disk, but retains only the current run and loses the prior-run fallback the current comment calls deliberate.
- Grow the volume.

Suggested pre-deploy posture: ship with a conservative cap, measure actual drift and peak for a few days, then tune.

## 3. Restrict warming to a 03Z-21Z wall-clock window

No user refreshes a flight in the middle of the night, so the overnight passes are wasted bandwidth. Gate on the **wall-clock time the warming runs**, not on the run's init hour.

D2 warming currently fires at **02:00, 05:00, 08:00, 11:00, 14:00, 17:00, 20:00, 23:00 UTC** (8 cycles, `publish_delay_hours=2.0`). A 03Z-21Z window drops the 23:00 and 02:00 passes: **24% fewer forecast-hours, ~83 GB less downloaded** over the measured period.

Because the passes land on a 3h grid, any window from 03Z to 20Z inclusive yields the same six daytime passes — 03Z-21Z is chosen for margin, not because the boundaries are tight.

Apply to:
- **ICON-D2 flight warming** — yes.
- **ICON-EU airport-profile precache** (`MAIN_CYCLE_HOURS`, publishing 03:00/09:00/15:00/21:00 UTC, ~19-21 GB download per pass) — yes; users are European.
- **GFS** — **no.** It is the cheapest of the three (~1.5 GB/run, 6 GB resident), so gating saves least, and it is the one thing that would have to be un-gated for the US expansion (`project_us_expansion_models`). Leave it running.

Implement as **per-model** window config rather than one global gate, so US expansion is a config change rather than a code change.

**Skip, don't defer.** When a pass is skipped, `last_done` must NOT be updated, and we must not backfill the skipped run later — at 03:00 the loop should warm whatever run is freshest *then*, not replay the 21z/00z runs. Warming a stale run at dawn is pure waste.

## 4. Correct the sizing claims in the design doc

Update `designs/weather-engine-specs.md` (the #469 phase 3 bullet) to carry the measured figures rather than "a few GB/day", including the per-file / per-forecast-hour unit costs and the 03Z-21Z window. Whoever next reasons about precache economics should start from measurements, not the original estimate.

---

## Acceptance criteria

- [ ] `purge_old_runs` derives run age from the init time in the directory name; falls back to `st_mtime` only when the name doesn't parse. Unit test: a run dir whose mtime is fresh but whose init is older than the TTL **is** purged.
- [ ] Size cap with oldest-init-first, whole-directory eviction and a configurable floor (default 2 runs). Unit tests: cap triggers eviction; floor is never breached; evictions are logged.
- [ ] Per-model wall-clock warming window, default 03Z-21Z for `icon-d2` and `icon-eu`, GFS ungated. Unit tests: a pass inside the window runs; a pass outside is skipped **without** updating `last_done`; a skipped run is not backfilled later.
- [ ] Design doc updated with measured sizing.
- [ ] Full suite green.

## Verification after deploy

- `WB_GRIB_PRECACHE_ENABLED=false` remains the kill switch if disk moves the wrong way.
- Watch `df -h /mnt/flyfun_data` and `du -sh .cache/grib/*` per model for the first 48h; compare observed D2 peak against the predicted ~42 GB.
- Measure real D2 drift (dir mtime - init) once warming is live, and re-tune the cap.
- Expect a one-off shrink of the `icon-eu` cache as item 1 starts collecting runs that were previously pinned by mtime drift.

## Related

- #469 — the original fetch-cost work (phases 1/3 shipped, phase 2 gated off)
- #474 — re-land the ceiling-limited fetch; would cut ~15-19% off every run and is the durable disk fix

roznet:
## What & why

Fixes issue #475 (GRIB cache bloat) with three complementary changes:

1. **Age cache by init time, not mtime** — `purge_old_runs()` now parses the init datetime from the cache directory name (`YYYYMMDD_HHz`) instead of trusting mtime. This fixes the core bug: mtime resets to "now" every time a later briefing tops up the run with a new file, so TTL-based eviction was ineffective. Unparseable/legacy directory names fall back to mtime.

2. **Add per-model disk size cap** — After TTL eviction, if the retained set still exceeds a cap (45 GiB for ICON-D2, configurable via `WB_GRIB_CACHE_CAP_GB_<MODEL>` env), evict the oldest-init runs whole until it fits, never dropping below a floor of 2 runs (current + prior-run fallback). This backstops the TTL: retained size scales with flight count, which grows over time, so TTL alone can't bound disk.

3. **Gate precache warming to daytime hours** — New `is_within_warming_window()` and `should_warm()` functions gate ICON-D2 and ICON-EU precache to 03Z–21Z UTC (daytime in Europe), skipping overnight passes that waste DWD bandwidth. GFS remains ungated. A skipped pass is never backfilled; the next in-window tick simply warms whatever run is freshest then.

The scheduler's daily `_run_retention_once()` now includes ICON-D2 in its purge loop (previously only GFS and ICON-EU), and the precache loop uses the new gating logic.

## Issue linkage

Closes #475

## Testing / verification

- Added 14 new unit tests covering:
  - Init-time parsing from directory names (valid/invalid)
  - TTL eviction by init time vs. mtime
  - Size cap eviction (oldest-init-first, floor preservation, logging)
  - Env override and default cap lookup
  - Warming window gating (per-model, boundary cases, no backfill)
- All existing tests pass; no breaking changes to public APIs (new parameters are keyword-only with sensible defaults).

https://claude.ai/code/session_01Hud2NMVtCdoaHioXhsBzP4


roznet:
Thanks for the review — both Important findings were spot-on. Fixed in `92b8ce4`:

**1. Cap walk off the hot path.** `purge_old_runs` now gates the size-cap behind a new `enforce_cap` flag (default `False`). The cheap name-based **TTL purge still runs everywhere** (including the per-briefing `_prepare_icon_eu`/GFS path), but the expensive recursive `rglob`+`stat` cap walk runs **only from the scheduled contexts** — the warm loop after each warm, and the daily retention pass. So a user-facing D2 briefing no longer pays 10–25k stats to answer "over cap?". Init-time aging already bounds that path to ~2 runs, and the cap is a slow-moving backstop, so per-warm + daily enforcement is the right cadence.

**2. Concurrent-vanish race.** Hardened the purge helpers to degrade-to-skip instead of propagating `FileNotFoundError` into enrichment: the `_run_age_seconds` mtime fallback now returns `+inf` on a failed stat (sorts/purges as oldest; the eviction `rmtree` is `ignore_errors`), and `_dir_size_bytes`'s scandir is wrapped. This matches the surrounding `_prepare_icon_eu` degrade-to-skip pattern.

**Tests:** cap tests now pass `enforce_cap=True`; added coverage that the cap is *not* enforced without the flag (hot-path guarantee), that a dir vanishing mid-cap-walk doesn't raise, and that the helpers tolerate nonexistent paths.

**Deferred (Minor):** the per-eviction log total can understate if an `rmtree` is partially blocked. Leaving as-is — it matches the established `ignore_errors` pattern in the TTL loop above it and is debug-log-only, per your own note. Happy to revisit if cap enforcement ever gets debugged from logs alone.

---

PR #477, opened 2026-07-22
feat(icon-d2): re-land ceiling-limited fetch, make sounding consumers truncation-safe (#474)

### LINKED ISSUE #474: Re-land ICON-D2 ceiling-limited fetch: make sounding consumers safe for a truncated column (#469 phase 2)
Phase 2 of #469 (ceiling-limited ICON-D2 sounding fetch) shipped in PR #470 but is **gated off by default** (`WB_ICON_CEILING_LIMIT_ENABLED`, default false) — see commit 79f51bb6. The fetch-side cut is correct; the blocker is that the sounding *consumers* aren't safe for a truncated column, and every failure mode reads as **reassuring** rather than unavailable.

## Why it's off

The cut is asymmetric by necessity — `t/qv/p` stay full column so the MetPy CAPE buoyancy integral can reach its equilibrium level (live ESMX→EKRK: EL FL306 at 6,000 ft cruise). That leaves pressure levels above the cut carrying temperature but **no wind**, and:

1. **Wind is gated all-or-nothing.** `analysis/sounding/prepare.py:99` — `has_wind = all(lv.wind_speed_kt is not None ... for lv, _ in valid)`. One wind-less level drops wind for the **entire** profile. Downstream, ICON-only, at *all* altitudes: no Richardson number → `cat_risk_layers == []` → no EDR, no e-shear, no 0–6 km / 0–1 km bulk shear, no cross-section wind, no Skew-T barbs, and `generate_hodograph` raises.
2. **Turbulence then grades GREEN.** `analysis/advisories/turbulence.py:101` returns UNAVAILABLE only when the vertical-motion assessment is missing — omega survives the cut, so `vm` exists, `affected == 0`, and it reports "smooth". This is the #391/#393 "absence reads as clear" failure exactly.
3. **Truncated `clc` reads as clear sky.** `analysis/sounding/clouds.py:443` appends `None` for levels lacking cloud fraction, indistinguishable from sub-FEW. A deck above the cut disappears → understated tops in the cloud-top advisory, spurious "DD/NWP clouds disagree" in `dd_nwp_agreement.py`, and a wrong digest.
4. **Charts blank silently.** The cross-section renders to ceiling + 5,000 ft (`web/ts/visualization/data-extract.ts:94`, and the same `+ 5000` in `CrossSectionViewModel.swift:383`); the Skew-T renders to 250 hPa. With an 18,000 ft ceiling the cut sits at ~18,680 ft, so ~18.3k–23k ft goes blank for ICON only while ECMWF/GFS still draw it.

## Two ways to re-land

**A — fix the consumers, keep the asymmetric cut.**
- Make the wind gate per-level, mirroring the omega branch directly below it (`any(...)` + NaN fill), in `prepare.py` and `digest/skewt.py:244,407`.
- Give `clouds.py` an "above the fetched cut" sentinel distinct from "sub-FEW", so a truncated column can't publish deck tops that read as clear.
- Clip or label the cross-section and Skew-T at the fetched top.

**B — uniform cut across all variables.** Removes the mixed-column problem entirely (every level present has every variable, so the `all()` gate passes untouched) and is the simpler mental model, but must then solve CAPE — the thermodynamic column genuinely needs to reach the equilibrium level, which is a property of the atmosphere and not of the flight.

Either way, the guiding invariant: **a column truncated by a fetch optimisation must never be reportable as calm, clear, or smooth** — only as unavailable.

## Notes

- All the phase-2 machinery stays in the tree and unit-tested (`tests/test_icon_d2_ceiling.py`, incl. gate-on and gate-off cases), so re-landing is a default flip plus the above.
- Phases 1 (per-level cache + top-up) and 3 (flight warming) are **independent of the flag and active** — they don't read it.
- Worth re-measuring the prize first: phase 2 was sized at ~19% of a D2 run, but with phase 3 warming at a near-100% hit rate the marginal saving may be much smaller than it looked.

roznet:
## What & why

Re-lands the #469 phase-2 ceiling-limited ICON-D2 fetch (**approach A** from the issue: fix the consumers, keep the asymmetric cut) and flips `WB_ICON_CEILING_LIMIT_ENABLED` to **default true**.

The cut keeps `t/qv/p` full-column (MetPy CAPE integrates to an equilibrium level far above the flight) and drops `u/v/w/qc/qi/clc` above a ceiling-derived level, so upper pressure levels carry temperature but no wind/cloud. It was held back because every failure mode of that truncated column read as **reassuring** rather than unavailable. This PR teaches the consumers the guiding invariant: *a column truncated by a fetch optimisation must never read as calm, clear, or smooth — only as unavailable.*

**Server-side correctness (the actual gate on the flag):**
- `prepare.py` — wind gated **per level** (NaN fill, mirroring the omega branch) instead of all-or-nothing, so the flyable column below the cut keeps real wind. Richardson/CAT still grade at cruise and turbulence no longer reads a false GREEN "smooth" — the #391/#393 absence-reads-as-clear failure.
- `thermodynamics.py` — `_compute_bulk_shear` returns `None` (never NaN) when its top level is truncated.
- `clouds.py` + models — `build_nwp_cloud_layers_from_fraction` flags a deck `top_truncated`; `SoundingAnalysis.fetched_column_top_ft` records the fetched top. `cloud_top` never treats a truncated deck as toppable; `dd_nwp_agreement` clips the DD-vs-NWP overlap to the fetched top.
- `digest/skewt.py` — barbs drawn per level and hodograph built from the wind-carrying levels rather than failing all-or-nothing.

**Fetch cut covers the display band, not the bare ceiling (e5343d1).** The cross-section / Skew-T render wind/cloud to `ceiling + 5000 ft`, so the cut gates on `ceiling + CROSS_SECTION_DISPLAY_BUFFER_FT` — a shared, named constant mirrored on web (`visualization/types.ts`) and iOS (`VizRouteData.displayBufferFt`) — so ICON-D2 never shows a blank strip where the full-column models draw. Consequence: with the coarse anchors, typical high-ceiling flights (FL180) now fetch the full column; lower ceilings still truncate.

**Client (label, not refactor):** `fetched_column_top_ft` flows to the web + iOS DTOs; the cross-section draws a labelled "not modeled above FLxxx" reference line at the fetched top (dormant once the cut covers the display band, kept as a safety net).

**Scope note on the NWP-default consumers.** The default grading method is NWP (`ENGINE_METHOD_DEFAULTS` = `ogimet_nwp`/`nwp`), and the icing/`vfr_feasibility`/`vmc_cruise`/`convective_character` cloud scans don't read `top_truncated`. This is **safe** because the cut always sits at/above the ceiling (now +5000): the truncated strip is entirely above the ceiling, and every advisory is ceiling-bounded, so no advisory evaluates on truncated data at flight-relevant altitudes. Propagating truncation-awareness through those sibling consumers is deferred to a follow-up as belt-and-suspenders. (See the review thread for the full analysis.)

## Issue linkage

Closes #474

## Testing / verification

- New `tests/test_icon_d2_ceiling_consumers.py`: per-level wind gate, `fetched_column_top_ft`, CAT alive in the flyable column, bulk-shear None-not-NaN, the `top_truncated` sentinel, the `dd_nwp` clip.
- `tests/test_icon_d2_ceiling.py`: cut updated to `ceiling + buffer` semantics + an explicit coupling test; gate default flipped to on.
- `tests/test_skewt.py`: truncated-column Skew-T barbs + hodograph smoke tests.
- Sounding / advisory / digest / pipeline / grib / models suites green (1300+ tests). Ruff clean on touched files.
- Web `npm run build` (esbuild) passes; `tsc --noEmit` reports no new errors from touched files.
- Swift not compiled (no toolchain on Linux); additive optional DTO fields + a reference line and constant mirroring existing patterns.

https://claude.ai/code/session_01XtSypDxNfhtnxdWtiXd9ZJ


roznet:
Thanks for the review — worked through each finding with the maintainer.

**Pushed (e5343d1):** a separate, more fundamental gap the review surfaced indirectly — the cross-section renders wind/cloud to `ceiling + 5000 ft`, but the fetch cut only covered the bare ceiling, so ICON-D2 went blank across that whole rendered band. The cut now gates on `ceiling + CROSS_SECTION_DISPLAY_BUFFER_FT` (a shared, named constant mirrored on web + iOS), so the fetched column always reaches the top of what's drawn. Also cleaned up two now-stale comments from the flag flip (`__init__.py` "GATED OFF", `prepare.py` omega note).

**Critical (icing/cloud NWP-default consumers don't check `top_truncated`) — not acting, by design.** The mechanism is real (`ENGINE_METHOD_DEFAULTS` is indeed `ogimet_nwp`/`nwp`, and those `is_in_cloud_layer` scans don't read truncation), but it can't produce a reassuring error in the normal flow:
- The cut is chosen to sit at/above the ceiling **and** now covers `ceiling + 5000`, so within the flight envelope every level is present and evaluated on complete data.
- Every advisory is ceiling-bounded (`altitude_table` evaluates `range(2000, flight_ceiling_ft+1)`; `climb_above_icing` is feasible only if `≤ ceiling`; `cloud_top` only considers decks with `base ≤ ceiling`). The truncated strip is entirely above the ceiling, which no advisory acts on.
- The only trigger is evaluating *strictly above* the fetched ceiling on a stale pack — and even there the climb-escape target stays `≥ ceiling+margin` → infeasible, not falsely "clear to climb." As you noted for the `cloud_top` branch, it's the same non-normal "ceiling raised after fetch" case.

**Important / Minor:** `prompt_builder` truncated-top note and the `fetched_column_top_ft` fingerprint-vs-flag gating are deferred to the same follow-up if we ever want belt-and-suspenders truncation-awareness across the NWP-default consumers; they don't affect correctness at the evaluated altitudes. The `prepare.py` stale-comment Minor is fixed in this push.

(One correction to my own PR description: it said "the default DD track is unaffected" — the default grading is actually NWP; the point that stands is that the truncation is confined to above the ceiling.)

---

roznet:
Closing without merging — superseded by #478.

The review (Codex, plus an independent pass) surfaced four issues, all verified in code. Rather than fix them, we re-examined whether the optimisation still earns its complexity, and it doesn't:

- Widening the cut to `ceiling + 5000` (correctly, so ICON-D2 doesn't draw a blank strip where full-column models draw) leaves **45% of prod flights (33/73) truncating nothing at all**, and a **flight-weighted saving of ~9.8%** — roughly 6 GB/day of ~62 GB/day.
- The cost is a provenance layer threaded through fetch → decode → analysis → pack schema → DTOs → web → iOS, because the mixed column (t/qv/p full, u/v/w/qc/qi/clc cut) has to be *inferred* from decoded data shape. That inference isn't reversible, and every miss fails toward "clear/calm/smooth" — the direction the invariant forbids.

Dropping the cut makes findings 1, 3 and 4 evaporate and reduces finding 2 to a three-line assertion, since a constant request makes completeness a set-size check.

**Salvaged into #478** — the three consumer-hardening fixes, which are valuable independently of truncation because download failures happen: the per-level wind gate in `prepare.py`, its required companion (bulk shear returning `None` not NaN), and the per-level Skew-T barbs / hodograph. On main today a single failed `u`/`v` level drops wind from the whole profile and turbulence grades GREEN "smooth".

Thanks — the `ceiling + 5000` correction in e5343d1 is what made the economics legible.

PR #479, opened 2026-07-22
Simplify ICON-D2 fetch: drop the ceiling-limited cut, skip incomplete hours, harden against partial wind (#478)

### LINKED ISSUE #478: Simplify ICON-D2 fetch: drop the ceiling-limited cut, always fetch the full column, harden against partial data
## Decision

Drop the ICON-D2 ceiling-limited fetch (#469 phase 2) permanently: **always fetch the full model column**. Remove the machinery rather than leaving it gated off, and salvage the three consumer-hardening fixes from PR #477 that are valuable on their own.

Supersedes #474 (closed without implementing) and PR #477 (closed without merging).

### Why

**The optimisation stopped paying.** PR #477 correctly widened the cut from the bare ceiling to `ceiling + 5000 ft`, because the cross-section and Skew-T render to `ceiling + 5000` and ICON-D2 would otherwise draw a blank strip where full-column models draw. Measured against the 73 real prod flights, that leaves:

- **45% of flights (33/73) truncate nothing at all** — every flight at ceiling ≳ 14,000 ft fetches the full column anyway
- **flight-weighted volume saving: ~9.8%** — roughly 6 GB/day out of ~62 GB/day

**The cost was an architecture.** Because `t/qv/p` must stay full-column for the CAPE integral while `u/v/w/qc/qi/clc` are cut, the truncated column is *mixed*: upper pressure levels carry temperature but no wind or cloud. Nothing recorded that this was intentional, so PR #477 had to **infer** it from the shape of the decoded data (two separate fingerprints, in `sounding/__init__.py` and `clouds.py`). That inference is not reversible — fetch failure, a whole-variable failure, and a legacy whole-column blob all produce shapes it misreads or misses — and **every miss fails toward "clear/calm/smooth"**, the exact direction the invariant forbids. External review (Codex) found four issues on that basis; all four were verified in code.

Paying for a provenance layer threaded through fetch → decode → analysis → pack schema → DTOs → web → iOS, to unlock <10% of one model's fetch volume, is the wrong trade. Dropping the cut makes findings 1, 3 and 4 evaporate and reduces finding 2 to a three-line assertion, because **once the full column is always requested, the request is a constant** and completeness is just a set-size check.

### What does NOT change

The #475 disk and bandwidth sizing already assumed the full column (450 files/forecast-hour = 9 vars × 50 levels, ~41.6 GB peak, ~62 GB/day). So the 45 GiB `icon-d2` cap stays correct and there is **no disk regression** — we are forgoing a saving that never ran in production.

---

## Part 1 — Remove the ceiling machinery

Delete from `fetch/grib/icon_eu_fetch.py`:
- `icon_ceiling_limit_enabled` (and the `WB_ICON_CEILING_LIMIT_ENABLED` env var)
- `icon_limited_top_level`, `icon_levels_by_var`
- `ICON_D2_CEILING_LEVEL_CUTS`, `FULL_COLUMN_VARIABLES`, `LIMITED_LEVEL_VARIABLES`

Delete from `fetch/grib/__init__.py`:
- `_IconEuContext.levels_by_var` + `levels_for_var()` (and the `__slots__` entry) — callers use `ctx.levels`
- the `icon_levels_by_var(...)` call site in `_prepare_icon_eu`

**Signature change — grep every caller** (CLAUDE.md): remove the `flight_ceiling_ft` parameter from `enrich_forecasts`, `_enrich_forecasts_inner` and `_prepare_icon_eu`. Known call sites passing it: `tasks/fetch.py:463`, `tasks/time_scan.py:595`, `fetch/grib/precache.py:318`. Note `api/airport_profile.py:481` and `tests/test_icon_d2_explicit_conv.py:637` also call `enrich_forecasts` but do **not** pass it, so they need no change — verify before landing.

Delete `tests/test_icon_d2_ceiling.py`.

## Part 2 — Keep the per-level cache, but re-justify it

**Keep `per_level_cache=True` for ICON-D2.** It is volume-neutral (DWD already publishes one file per level; phase 1 only changed how they are stored), and reverting to whole-column blobs would make the flakiness *worse*: a blob cannot distinguish a complete column from a partial one, so a partial write is cached as complete and silently serves a truncated sounding — the original #469 phase-1 finding.

**But its documentation currently justifies it via phase 2 and must be rewritten to stand alone.** The new rationale is *partial-write safety*, not *prerequisite for truncation*. Sites to rewrite:
- `icon_eu_fetch.py:83-88` — the `per_level_cache` field docstring ("makes a partial-level fetch SAFE … the ceiling-limited top-up, phase 2")
- `icon_eu_fetch.py:278` — the ICON_D2 variant comment ("where a ceiling-limited fetch pays off")
- `icon_eu_fetch.py:1112` — `fetch_icon_eu_per_level` docstring ("the per-level cache layout … that makes a partial-level fetch safe")
- `__init__.py:3550` — the decode comment ("any ceiling-limited subset is satisfiable from it")

Leave the legacy whole-column blob paths (`icon_model_level_var_legacy_label`, `{prefix}_QC_QI_P`) alone: they are variant-generic and remain ICON-EU's normal operating path. Special-casing them out for D2 would add a branch, not remove one.

## Part 3 — Salvage the hardening fixes from PR #477

All three are "be robust to missing wind at *some* levels", valuable independently of truncation because download failures happen. **On main today, one missing `u`/`v` level drops wind from the entire profile → no Richardson → no CAT layers → turbulence grades GREEN "smooth"** — the #391/#393 absence-reads-as-clear failure, reachable via a single failed download.

1. `analysis/sounding/prepare.py` — wind gated **per level** with NaN fill (mirroring the omega branch) instead of `has_wind = all(...)`.
2. `analysis/sounding/thermodynamics.py` — `_compute_bulk_shear` returns `None` (never NaN) when a layer endpoint has no wind. **Required companion to (1)**, not optional: the NaN fill is what lets NaN reach the shear calc, and NaN would poison the JSON.
3. `digest/skewt.py` — barbs drawn per level; `generate_hodograph` built from the wind-carrying levels instead of failing all-or-nothing.

Do **not** salvage: the `top_truncated` sentinel, `fetched_column_top_ft` inference, the `dd_nwp_agreement` `max_alt_ft` clip (only existed for truncation), or any client work (web TS, iOS, i18n, DTOs, reference line, the tri-language `CROSS_SECTION_DISPLAY_BUFFER_FT`).

## Part 4 — Verify decode completeness (replaces the provenance architecture)

`_decode_and_merge_icon_eu` currently collects whatever `is_cached` happens to return and proceeds on `if level_paths:` — merely non-empty. It never checks that the requested set arrived, so a partial fetch failure silently produces a truncated column. This is Codex's finding #2, and it is **live on main today** independently of the ceiling cut, because phase 1 made partial columns possible.

Now that the request is a constant (`range(level_min, level_max + 1)`), require the complete set per variable:
- per-level variants: all expected levels present, else the hour is incomplete
- whole-column blob (ICON-EU's normal path, or a legacy hit): accept as-is — completeness is not verifiable there, which is precisely why D2 uses per-level

**Decided: (a) skip that forecast hour for ICON.** An incomplete hour is dropped from the ICON slot rather than decoded partially. A missing hour is an already-handled state, and the next briefing tops it up naturally via the per-level cache, whereas a partial column is a state nothing handles. (Rejected: falling back to ICON-EU for the whole briefing — too blunt for one bad hour; decoding-but-marking — reintroduces the "partial column that consumers must remember to check" problem this issue exists to remove.)

**Verify during implementation that a skipped hour genuinely reads as UNAVAILABLE and not as quiet/clear** — that is the same class of bug being fixed here and it must not be reintroduced at the hour level. Log the incompleteness (model, run, fhour, variable, missing count) — never drop data silently.

## Cleanup bar

This is a simplification, so the result must read as simple:
- No vestigial parameters, dead flags, or unused helpers left behind.
- No comment or docstring left explaining itself by reference to a feature that no longer exists (see the Part 2 list — the reviewer should grep for `ceiling-limited`, `phase 2`, `#474` and find only the design-doc history entry).
- `designs/weather-engine-specs.md`: replace the phase-2 bullet with a short **evaluated → rejected → removed** entry carrying the measured numbers (9.8% flight-weighted, 45% of flights unaffected, the mixed-column/inference hazard), so nobody re-proposes it blind. Rewrite the phase-1 bullet to justify the per-level cache on partial-write safety.

## Acceptance criteria

- [ ] No references remain to `WB_ICON_CEILING_LIMIT_ENABLED`, `icon_levels_by_var`, `icon_limited_top_level`, `ICON_D2_CEILING_LEVEL_CUTS`, `FULL_COLUMN_VARIABLES`, `LIMITED_LEVEL_VARIABLES`, `levels_by_var`, `levels_for_var` in `src/` or `tests/`.
- [ ] `flight_ceiling_ft` no longer threads through the GRIB fetch path; all callers updated; no caller passes a now-removed kwarg.
- [ ] ICON-D2 always fetches levels `level_min..level_max` for all 9 model-level variables.
- [ ] Per-level wind gate: a profile with wind missing at some levels still grades CAT/Richardson on the levels that have it; test asserts turbulence does **not** read GREEN "smooth" purely because some levels lack wind.
- [ ] Bulk shear returns `None`, never NaN, when a layer endpoint lacks wind.
- [ ] Skew-T renders barbs from a partial-wind column; hodograph builds from wind-carrying levels.
- [ ] Decode rejects an incomplete per-level set and handles it per the Part 4 decision, with a log line.
- [ ] Design doc updated per the cleanup bar.
- [ ] Full suite green.

## Related

- #469 — original fetch-cost work (phase 1 per-level cache and phase 3 flight warming stay; phase 2 removed here)
- #474 — closed without implementing, superseded by this
- PR #477 — closed without merging; parts 3's fixes are salvaged from it
- #475 — cache retention/cap/window; its sizing already assumed the full column, so unaffected

roznet:
## What & why

Implements #478: drop the ICON-D2 ceiling-limited fetch, always fetch the full model column, and harden the sounding against partial data. Two commits — a mechanical removal, then the behavioural changes.

**Why the cut goes.** PR #477 correctly widened it to `ceiling + 5000 ft` (the cross-section and Skew-T render that high, so a bare-ceiling cut drew a blank strip where full-column models draw). Measured against the 73 real prod flights that leaves **45% of flights (33/73) truncating nothing at all** and a **flight-weighted saving of ~9.8%** — roughly 6 GB/day of ~62 GB/day.

The cost was an architecture. Because `t/qv/p` must stay full-column for the CAPE integral while `u/v/w/qc/qi/clc` are cut, the truncated column is *mixed*: upper levels carry temperature but no wind or cloud. Nothing recorded that this was intentional, so consumers had to **infer** it from decoded data shape — an inference that can't distinguish intent from fetch failure, and whose every miss fails toward *clear / calm / smooth*, the direction the safety asymmetry forbids.

## Commit 1 — remove the machinery

Deletes `icon_ceiling_limit_enabled` (+ `WB_ICON_CEILING_LIMIT_ENABLED`), `icon_limited_top_level`, `icon_levels_by_var`, `ICON_D2_CEILING_LEVEL_CUTS`, `FULL_COLUMN_VARIABLES`, `LIMITED_LEVEL_VARIABLES`, `_IconEuContext.levels_by_var`/`levels_for_var`, and the `flight_ceiling_ft` parameter threaded through `enrich_forecasts` → `_enrich_forecasts_inner` → `_prepare_icon_eu` with its three call sites. `tests/test_icon_d2_ceiling.py` deleted.

**The per-level cache (phase 1) stays**, re-justified on its own terms: it is volume-neutral (DWD publishes one file per level either way) and a whole-column blob cannot tell a complete column from a partial one, so a failed partial write would be cached as complete.

## Commit 2 — completeness + partial-wind hardening

**Decode completeness.** `_decode_and_merge_icon_eu` collected whatever `is_cached` returned and proceeded on `if level_paths:` — merely non-empty. Now that every variable is requested over the same full column, "did we get what we asked for" is a count: an hour missing any level of any variable is **skipped for ICON entirely**, with a log line naming the shortfall. Per #478's decision (a): a missing hour is a state the pipeline handles honestly, a partial column is not, and the next briefing tops up only what's absent. Whole-column blobs are accepted without a count — their completeness is unverifiable, which is exactly why D2 doesn't use that layout.

**Partial-wind hardening**, salvaged from the closed PR #477 (written there for the ceiling cut, but a failed `u`/`v` download produces the same shape, so they stand alone):
- `prepare.py` gated wind all-or-nothing, so **one** missing level dropped wind from the whole profile → no Richardson → `cat_risk_layers == []` → turbulence graded GREEN "smooth". That's the #391/#393 failure, reachable today from a single failed download. Now per-level with NaN fill.
- `thermodynamics.py` bulk shear returns `None`, never NaN — **required companion**, since the NaN fill is what lets NaN reach it and NaN would poison the JSON.
- `skewt.py` draws barbs per level and builds the hodograph from wind-carrying levels rather than failing the whole plot.

## Cleanup

Per the issue's cleanup bar — no vestigial parameters or flags, and nothing left explaining itself by reference to a feature that no longer exists. `grep -rn "ceiling-limited\|levels_by_var\|WB_ICON_CEILING\|phase 2" src/` is empty. The phase-1 docstrings, the ICON_D2 variant comment, the warming comment in `precache.py` and the cap-warning in `cache.py` were all rewritten to stand on their own. The design doc's phase-2 bullet is now an **evaluated → rejected → removed** record carrying the measured numbers so nobody re-proposes it blind.

## Issue linkage

Closes #478

## Testing / verification

- New `tests/test_sounding_partial_wind.py` (5 tests): per-level wind gate keeps wind for the levels that have it; all-windless still yields `None` (NaN-fill doesn't over-trigger); turbulence axis stays alive when upper wind is missing; bulk shear `None`-not-NaN; a full column is unaffected.
- New completeness tests in `test_icon_d2_per_level.py` (3): incomplete hour skipped with the shortfall logged, complete hour still decodes (not over-eager), whole-column blob accepted without a level count.
- `test_skewt.py`: partial-wind Skew-T + hodograph, re-framed from #477.
- `test_grib_precache.py`: the ceiling-passing test narrowed to duration, now asserting `flight_ceiling_ft` is *not* passed.
- Full suite: **3829 passed, 9 skipped** (excluding `test_auth.py`, which has a pre-existing `GOOGLE_CLIENT_ID` env failure unrelated to this work).

## Not done here

No re-measurement of what phase-3 warming's near-100% hit rate does to the marginal value of *any* future fetch optimisation — worth doing before one is proposed again. The #475 disk/bandwidth sizing already assumed the full column, so the 45 GiB cap and ~62 GB/day figures are unchanged by this PR.


roznet:
## Review round 2 — external review (Codex) findings addressed

Both P1s were real and verified in code. Pushed `3cf77650`.

**P1 #1 — missing whole variables undetected on ICON-EU.** Confirmed by driving the decode planner directly: with `u` and `v` absent, ICON-EU decoded 7 of 9 variables. Since the GRIB sounding replaces the Open-Meteo levels wholesale, that means no wind anywhere → no Richardson → empty `cat_risk_layers` → turbulence grades GREEN "smooth" — the exact failure this PR exists to prevent, on the model actually in production (D2 isn't deployed). Variable presence is now required for **every** variant; the level count remains per-level-only, which is all it can be.

**P1 #2 — legacy blobs are not "complete by construction".** Confirmed at the writer: `fetch_icon_eu_per_variable` cached on `if buf:` — any single level arriving — and even logged `downloaded %d/%d levels` before caching. My previous commit asserted the opposite in both a code comment and the design doc; that claim is now removed. Fixed on both sides: the writer caches only when `downloaded == len(levels)`, and per-level variants refuse legacy blobs (per-variable and the older combined one).

**Knock-on the review didn't flag, caught by the failing tests:** refusing legacy blobs at decode without changing prefetch would have **deadlocked** the hour — prefetch treated a blob as covering the column, so nothing per-level would be fetched, decode would then decline the blob, and the hour would be skipped until the blob aged out. Prefetch now ignores blobs for per-level variants too, so the two paths agree. Two prefetch tests were updated accordingly (they had encoded the old trusting behaviour).

**On existing ICON-EU blobs:** letting the 12h TTL age them out rather than bumping the cache key. A key bump would force an immediate ~22 GB refetch in prod, for a risk that already exists today and self-heals.

**Also fixed (P2/P3):** `weather-engine-specs.md:176,179` stale passages, and the vestigial `flight_ceiling_ft` in `_fake_prepare`.

New tests: ICON-EU missing-variable skip, ICON-EU complete-blob-set still decodes, per-level refusal of both legacy layouts, writer discards/keeps partial vs complete columns. Full suite **3833 passed, 9 skipped**.

Issue #487, opened 2026-07-22
Freezing-level fallback chain mixes AGL (ECMWF deg0l) and MSL datums


roznet:
## Problem

`_explicit_freezing_level_ft` (`src/weatherbrief/analysis/sounding/convective.py:960-977`) builds a fallback chain across three freezing-level sources:

```python
candidates = [indices.freezing_level_ft, indices.nwp_freezing_level_ft]
if nwp_diagnostics is not None:
    candidates.append(nwp_diagnostics.freezing_level_ft)
```

For ECMWF, `nwp_diagnostics.freezing_level_ft` comes from `deg0l`, which the decoder documents as **metres above ground**:

```
# DATUM: ECMWF ceil/cbh/hcct/deg0l are all metres ABOVE GROUND (AGL),
"deg0l": "freezing_level_m",   # meters AGL (geometric height above ground, not MSL)
```
(`src/weatherbrief/fetch/grib/decode.py:153`, `:160`)

The first two candidates are sounding-derived and MSL-referenced. The chain falls back across datums without conversion.

The decode layer itself is correct and consistent - this is a consumer that mixes them.

## Impact

The value feeds the bright-band gate for explicit convection (`_EXPLICIT_BRIGHT_BAND_DELTA_FT`, echo-top-vs-melting-layer comparison). Over European terrain the AGL/MSL difference is typically small, but it is a real systematic offset equal to terrain elevation, and it only appears when the two MSL candidates are both absent - i.e. exactly when there is least other information to catch it.

The docstring currently asserts "Any is a usable datum for the echo-top-vs-melting-layer comparison", which is the assumption worth re-examining, since the echo top's own altitude conversion uses the sounding column.

## Questions to settle

1. Should the ECMWF AGL value be converted to MSL (needs the point's terrain elevation) before entering the chain?
2. Or should it be dropped from this particular gate, given two better-referenced candidates already precede it?
3. Does any other consumer of `NWPCloudDiagnostics.freezing_level_ft` make the same assumption?

Worth a line in `designs/meteorology-decisions.md` either way.

Found while reviewing an external audit of the GRIB interpolation logic (adjacent to the spatial-interpolation field-list issue).

roznet:
Deployed to https://weather.flyfun.aero — give it a try and let us know how it works.

Issue #486, opened 2026-07-22
ICON geopotential-height reconstruction anchors on ISA rather than model surface


roznet:
## Problem

`_fill_missing_geopotential_height` (`src/weatherbrief/fetch/grib/decode.py:2180`) anchors a reconstructed pressure column at the ISA altitude of its highest-pressure level when no real geopotential is present, then integrates upward hypsometrically. This is the active path for ICON-EU and ICON-D2, because DWD never ships geopotential on model levels (`icon_eu_fetch.py:354`).

The function's own docstring already flags this as known simplification (b), and names the fix: anchor using surface `z` + `sp`.

## Correcting the framing

An external review reported this as a *terrain* error - claiming the whole sounding shifts by the terrain-versus-ISA difference, "potentially thousands of feet near high terrain". **That mechanism is wrong and should not be used to justify the work.**

`pressure_hpa_to_altitude_m` maps *pressure* to ISA altitude, and the anchor level's pressure already encodes terrain. Verified:

```
795 hPa -> 1999.6 m
```

Over 2000 m terrain the near-surface pressure *is* about 795 hPa, so the anchor lands on terrain automatically. No terrain offset is imported.

## The real error

The anchor error is the deviation of that pressure surface's true geopotential height from ISA - driven by MSLP and column mean temperature. Order **100-300 m (roughly 300-1000 ft)** in a deep low or a cold column.

Critically it is a *uniform column offset*: layer thicknesses come from the hypsometric integration with real temperatures and are unaffected. Since icing, turbulence and echo-top geometry depend mostly on relative thickness rather than absolute datum, the practical impact is smaller again.

## Worth noting alongside

The *other* documented simplification - dry-air temperature instead of virtual temperature in the layer mean - is the one that actually distorts thicknesses (about 30-50 m surface to 500 hPa). Smaller in magnitude but not a pure offset, so arguably the more interesting of the two.

## Fix

Anchor using ICON surface pressure and orography (or half-level height) instead of ISA. Optionally switch the layer mean to virtual temperature at the same time.

Priority: low. Both effects are modest and one is a pure offset.

roznet:
Deployed to https://weather.flyfun.aero — give it a try and let us know how it works.

Issue #481, opened 2026-07-22
GFS averaged cloud cover is labelled at its window endpoint, giving a 0.5-3 h lag


roznet:
## Problem

GFS low/mid/high cloud cover is read from the **time-averaged** records (`_PREFER_AVERAGED_PAIRS`, `src/weatherbrief/fetch/grib/gfs_idx.py:167`) — correctly, because the paired `PRES`/`TMP` cloud bottom/top geometry is published averaged-only.

But the averaged value is then used *as if it were valid at the record's forecast hour*. Verified against the live index, the window at f006 is `0-6 hour ave` — its true temporal centre is f003. At f012 the window is `6-12`, centred on f009.

So every GFS anchor carries a lag equal to half its window, oscillating through each 6-hour cycle:

| fhour | window | lag |
|---|---|---|
| f001 | 0-1  | 0.5 h |
| f003 | 0-3  | 1.5 h |
| f006 | 0-6  | 3.0 h |
| f007 | 6-7  | 0.5 h |
| f012 | 6-12 | 3.0 h |

`_interp_gfs_diag_hourly` already computes window midpoints for *gap-hour* interpolation, but GFS anchors are hourly through f120 (`compute_flight_window_hours`), so the interpolation loop short-circuits at `fill.py:217` and no midpoint correction is ever applied to the native values.

## Impact

This is always-on for every GFS briefing inside f120 — i.e. essentially all of them — and it is a systematic bias, not an edge case. A sawtooth 0.5-3.0 h lag in low/mid/high cloud cover, worst at the 6-hourly boundaries.

## Why this needs a decision, not just a patch

There is no free fix. The options each cost something:

1. **Relocate averaged anchors to their window midpoint** and interpolate the hourly grid from those. Correct in time, but the reported value at hour N is then no longer any single published GFS number, and it needs the neighbouring anchor to reconstruct.
2. **Switch cover to the instantaneous records** and accept that cover and its base/top/top-temp geometry then come from different statistical processing (geometry is averaged-only — there is no instantaneous form to pair with).
3. **Accept the lag and document it**, and make sure nothing downstream (convective timing, cross-check against DD) treats GFS cloud-cover timing as precise.

Option 2 in particular interacts with `designs/meteorology-decisions.md` — cover and geometry currently agree by construction, and decoupling them is a meteorology decision, not a refactor.

## Suggested next step

[...]

PR #488, opened 2026-07-22
fix(grib): GFS averaging windows, ECMWF accumulation + step bracketing, Greenwich seam, diagnostics field coverage

### LINKED ISSUE #480: GFS averaged-cloud window length table does not match NCEP cadence
## Problem

`_gfs_window_length_hours()` (`src/weatherbrief/fetch/grib/fill.py:269`) models the GFS averaging-window cadence as a repeating 1/2/3-hour cycle through f120, then a constant 3 hours afterwards. That is not what NCEP publishes.

Verified against the live index (`gfs.20260722/00/atmos/gfs.t00z.pgrb2.0p25.fNNN.idx`), LCDC/MCDC/HCDC averaging windows **reset at multiples of 6 and grow**:

```
f001 0-1      f004 0-4      f006 0-6      f012 6-12
f007 6-7      f009 6-9      f120 114-120
f123 120-123  f126 120-126  f129 126-129  f132 126-132
```

So the current table is wrong whenever `f % 6` is 4, 5 or 0:

| fhour | code | actual |
|---|---|---|
| 4   | 1 | 4 |
| 5   | 2 | 5 |
| 6   | 3 | 6 |
| 12  | 3 | 6 |
| 126 | 3 | 6 |
| 132 | 3 | 6 |

A single formula covers the entire range including past f120:

```python
window = fhour - 6 * ((fhour - 1) // 6)
```

`tests/test_grib_fill.py:386` bakes the wrong model in explicitly (`(126, 3), (132, 3)`) and must be updated with it.

## Impact

Narrower than it first looks. The function only feeds gap-hour interpolation in `_interp_gfs_diag_hourly`, and `compute_flight_window_hours` requests one fhour per UTC hour, so GFS anchors are hourly through f120 and the loop short-circuits at `next_i - prev_i <= 1` (`fill.py:217`). The wrong length therefore only reaches interpolation past f120, where cadence is 3-hourly — and there it is wrong on every *other* anchor.

Max resulting error in the interpolation midpoint is 1.5 h (half the window error).

The separate, larger issue of native anchors being labelled at the window endpoint is tracked in its own issue.

## Confirmation the averaged records are the right ones to read

Worth stating so nobody "fixes" this by switching to the instantaneous form: the code deliberately prefers the averaged records via `_PREFER_AVERAGED_PAIRS` (`src/weatherbrief/fetch/grib/gfs_idx.py:167`), and that is correct — `PRES`/`TMP` cloud bottom/top exist **only** in averaged form (verified at f006), so cover and geometry must share the same statistical processing.

## Fix

- Replace the table in `_gfs_window_length_hours` with the modulo-6 formula.
- Update `TestGfsWindowLength` parametrisation to the real values.

Found by an external review of the GRIB interpolation logic; verified against the live NOAA index.

### LINKED ISSUE #482: ECMWF accumulated precip: stale predecessor after a missing step inflates the next rate
## Problem

In the ECMWF a1 merge loop (`src/weatherbrief/fetch/grib/__init__.py:2385-2445`), the shared previous-valid-time advances unconditionally at the end of every processed step:

```python
prev_a1_valid_utc = valid_time
```

but the per-point accumulators only update when the value is present:

```python
tp = raw.get("total_precip_m")
if tp is not None:
    prev_tp_per_point[i] = tp
```

If a point's `tp` (or `sf`, or `cp`) is missing at step *k* but present at *k-1* and *k+1*, then at *k+1* the delta `tp - ptp` spans two steps while `window_hours` is computed from one — inflating the rate by roughly the ratio of the spans.

Affects three fields:
- `tp`, `sf` via `_apply_ecmwf_surface_to_hourly` (`__init__.py:2850`, `2854`)
- `cp` inline in the merge loop (`__init__.py:2404`)

## The fix already exists in this file

The ICON loop does the right thing at `__init__.py:3773-3783`, and its comment explicitly documents this failure mode:

> Assign unconditionally: a point that lacks rain_con at this step ... resets its predecessor to None, so the NEXT step differences against nothing -> None (missing-data-safe) rather than dividing a multi-step accumulation delta by a single-step window and silently inflating the rate. prev_valid_utc advances every processed step, so a stale predecessor value would otherwise desync from the window (review on 6c5555a).

That comment also says the ICON loop is *"a direct port of the ECMWF a1 loop (#421)"* — so ECMWF is the origin that never received the fix applied to its own descendant.

## Reachability

Lower than it might appear. If an *entire* a1 step fails to decode, `step_hours in a1_results` is False, the whole block is skipped and `prev_a1_valid_utc` is **not** advanced — that path is already safe. The bug needs a step that decodes successfully but has a per-point or per-field gap (partial GRIB, point coverage flapping).

## Fix

Mirror ICON exactly: assign `prev_tp_per_point[i]` / `prev_sf_per_point[i]` / `prev_cp_per_point[i]` unconditionally from `raw.get(...)`, so a missing value resets the predecessor to `None` and the next step declines to compute a rate rather than computing an inflated one.

Found by an external review of the GRIB interpolation logic.

### LINKED ISSUE #483: ECMWF decode margin is not cadence-aware and cannot bracket 6-hourly steps
## Problem

`ECMWF_FLIGHT_WINDOW_MARGIN` is a fixed 3 hours (`src/weatherbrief/fetch/grib/__init__.py:809`), used by the step filter at `__init__.py:2314-2316`:

```python
if valid_time < flight_start - margin or valid_time > flight_end + margin:
    continue
```

ECMWF output is not uniformly hourly or 3-hourly. Per the delivery manifest (`delivery_config.json`), the 00z/12z cycles run:

```
0..90 hourly, 93..144 3-hourly, then 150, 156, 162, 168
```

i.e. **6-hourly past hour 144**. A flight departing at hour 154 therefore keeps only step 156 — step 150 fails the filter because `150 < 154 - 3`.

## Impact

Degradation rather than corruption:
- pressure-level and instantaneous surface fields have a single anchor, so no bracketing interpolation happens
- accumulated `tp`/`sf`/`cp` have no predecessor and fall to the documented "first a1 in the processed window, leave Open-Meteo's value" path

Reachable for flights 6-7 days out, which the app supports.

Also worth noting: the 93-144 h 3-hourly region has *exactly* zero slack. It works only because the comparison is a strict `<`, so a step landing precisely 3 h before departure survives. Any change to that comparison silently breaks the 3-hourly range too.

## Fix

Make the margin cadence-aware — at least 6 h once steps are 6-hourly, and enough to guarantee both bracketing steps plus the accumulated-field predecessor.

**Constraint:** the constant is deliberately module-level and shared with `src/weatherbrief/tasks/time_scan.py:65`, which reads it for coverage detection specifically so the two do not drift. Whatever replaces it must stay shared in the same way — a cadence-aware helper both sides call, not a second magic number.

Found by an external review of the GRIB interpolation logic; step cadence verified against the delivery manifest.

### LINKED ISSUE #484: GFS bilinear interpolation returns nothing for longitudes in (-0.25, 0) at the Greenwich seam
## Problem

GFS route points are normalised with `lon % 360` (`src/weatherbrief/fetch/grib/decode.py:614` and again at `:700`), matching the GFS 0-360 convention. But the regular 0.25-degree grid ends at **359.75**, and `xarray.DataArray.interp` neither extrapolates past the last coordinate nor knows the longitude axis is cyclic.

Longitudes in `(-0.25, 0.0)` map to `(359.75, 360.0)` and fall off the end of the axis, returning NaN instead of wrapping across the seam.

Reproduced against a synthetic 0.25-degree global grid in the project venv:

```
lon   -0.10 -> 359.900 -> nan
lon   -0.29 -> 359.710 -> 0.0885
lon  359.90 -> nan
lon  359.70 -> 0.0967
```

The broken band is exactly `(-0.25, 0.0)`.

## Impact

Not exotic. Any UK-to-continent route crossing the Greenwich meridian has interpolated route points landing in that band.

Interior points are usually rescued by `interpolate_diagnostics_spatially`, so it typically degrades rather than voids. But an **endpoint or waypoint** in the band gets nothing, because the spatial fill deliberately refuses edge extrapolation (both neighbours must exist).

Affects both decoders:
- `decode_grib_per_point` (`decode.py:614`) - pressure levels
- `decode_cloud_diag_per_point` (`decode.py:700`) - cloud diagnostics

## Fix

Pad the longitude axis with a wrapped column at 360.0 (copy of the 0.0 column) before interpolating, so the seam cell interpolates correctly. Alternatively wrap targets at or above the last coordinate back into the seam cell explicitly.

Found by an external review of the GRIB interpolation logic; reproduced locally.

### LINKED ISSUE #485: Cloud-diagnostics interpolators have divergent field lists (spatial vs time axis)
## Problem

`NWPCloudDiagnostics` is interpolated by two separate functions with **independently drifting field lists**:

- `_lerp_diagnostics` - spatial axis, `src/weatherbrief/analysis/spatial_interpolation.py:218`
- `_interp_diag_at` - time axis (GFS gap hours), `src/weatherbrief/fetch/grib/fill.py:288`

Each omits fields the other handles:

| field | spatial | time |
|---|---|---|
| `freezing_level_ft` | **missing** | present |
| `convective_precip_mm_h` | present | **missing** |
| `k_index` | present | **missing** |
| `total_totals` | present | **missing** |
| `ml_cape_jkg` | present | **missing** |
| `ml_cin_jkg` | present | **missing** |

Every time a field is added to the model, both call sites must be updated, and nothing enforces it. That is the actual defect - the individual omissions are symptoms.

## Concrete consequence today

`freezing_level_ft` is populated for ECMWF from `deg0l` (`decode.py:160`, `:2384`). An interior route point filled spatially from two valid neighbours interpolates every other diagnostic but loses its freezing level.

Impact is third-order in practice: `_explicit_freezing_level_ft` (`analysis/sounding/convective.py:971-973`) tries `indices.freezing_level_ft`, then `indices.nwp_freezing_level_ft`, and only then the NWP-diag value. But the omission is silent and the next field added could matter more.

## Fix

Reconcile both interpolators onto one shared field list so a new field cannot be handled by only one axis. Add a test that fails when a `NWPCloudDiagnostics` field is absent from either interpolator - e.g. introspect the model fields and assert every non-layer scalar is covered.

Note the two axes legitimately differ in *how* they interpolate (the GFS time axis distinguishes averaged-window fields on `mid_frac` from instantaneous fields on `step_frac`), so the shared thing is the field inventory and its averaged/instantaneous classification, not the interpolation call itself.

Found by an external review of the GRIB interpolation logic.

roznet:
Fixes from an external audit of the GRIB spatial / temporal / vertical interpolation logic. I reviewed each reported finding against the code and verified the external claims against live data before fixing anything — one of the six reported findings turned out to rest on incorrect physics and is filed with a corrected framing rather than fixed here.

Closes #480, Closes #482, Closes #483, Closes #484, Closes #485

### #480 — GFS averaging-window model was wrong

`_gfs_window_length_hours` assumed a repeating 1/2/3-hour cycle capped at 3 h past f120. NCEP actually resets the averaging window at every multiple of 6 and lets it **grow** to the next reset. Verified against a live index (`gfs.20260722/00/.../fNNN.idx`, `LCDC:low cloud layer`):

```
f001 0-1      f004 0-4      f006 0-6      f012 6-12
f007 6-7      f009 6-9      f120 114-120
f123 120-123  f126 120-126  f129 126-129  f132 126-132
```

So f004 (code 1, real 4), f006 (3 vs 6), f012 (3 vs 6), f126 (3 vs 6) were all wrong — up to 3 h of window error, 1.5 h of interpolation-midpoint error. One formula now covers the whole range, including past f120 where the 3-hourly output cadence makes widths alternate 3 / 6.

The old model was pinned in `TestGfsWindowLength` and in a worked example in `designs/time-alignment-audit.md`; both are corrected. `TestGfsMidpointInterp` also encoded it — its f132 anchor is `126-132` (midpoint 09:00z), not `09-12z` (10:30z), which I re-confirmed against the live index before changing the expected value.

**Scope note:** GFS anchors are hourly through f120, so the gap-hour interpolation this feeds short-circuits there. The wrong length only reached interpolation past f120. The larger, separate problem — that averaged values are *labelled at their window endpoint*, giving an always-on 0.5–3 h lag — needs a product decision and is filed as #481, not fixed here.

### #482 — ECMWF accumulated precip kept a stale predecessor

`prev_a1_valid_utc` advanced for every processed step, but `prev_tp/sf/cp_per_point` only advanced when the value was present. After a per-point gap the next step subtracted a two-step-old accumulator while dividing by one step's duration, inflating the rate.

Predecessors are now assigned unconditionally (including `None`), so a gap invalidates the chain and the next step declines to emit a rate. This mirrors the ICON `rain_con` loop — whose comment says it was *"a direct port of the ECMWF a1 loop"* and which already carried this exact fix. Extracted to `_advance_ecmwf_accumulators` so it is testable, and it now walks every route point rather than the decoded prefix (a short `sfc_data` previously left the tail stale).

### #483 — ECMWF step selection could not bracket 6-hourly output

Per the delivery manifest, 00z/12z run `…141, 144, 150, 156, 162, 168`. With a fixed ±3 h margin a flight at +154h kept only step 156 — 150 fell outside `start − 3h` — leaving instantaneous fields with a single anchor and accumulated fields with no predecessor.

Rather than tune the magic number, step selection now unions the margin window with the **immediately bracketing steps**. That is cadence-independent and self-maintaining if ECMWF changes its schedule again. The shared `ECMWF_FLIGHT_WINDOW_MARGIN` deliberately stays at 3 h so `time_scan`'s coverage rule window is unchanged — it can now only under-claim relative to what was decoded, never over-claim.

### #484 — GFS interpolation returned nothing at the Greenwich seam

Targets are normalised with `lon % 360`, but the grid's last coordinate is 359.75 and `xarray.interp` neither extrapolates nor knows the axis is cyclic. Longitudes in `(-0.25°, 0°)` returned NaN. Reproduced before fixing:

```
lon  -0.10 -> 359.900 -> nan
lon  -0.29 -> 359.710 -> 0.0885
```

Every UK↔continent route crossing the meridian has points in that band. Interior points were usually rescued by spatial fill, but an endpoint or waypoint there got nothing (spatial fill refuses edge extrapolation).

Fixed by closing the cyclic longitude axis inside `_interpolate_per_point`, so all **three** call sites that normalise this way benefit (the audit named only one). Regional grids are detected as non-cyclic and left untouched — tests cover both that a regional grid is unmodified and that it still correctly returns `None` outside its domain.

### #485 — the two diagnostics interpolators had drifted apart

`_lerp_diagnostics` (spatial) and `_interp_diag_at` (time) each silently dropped fields the other carried:

| field | spatial | time |
|---|---|---|
| `freezing_level_ft` | **missing** | present |
| `convective_precip_mm_h`, `k_index`, `total_totals`, `ml_cape_jkg`, `ml_cin_jkg` | present | **missing** |

The audit caught only the first direction. Both now derive coverage from one shared inventory in `models.analysis`, which also carries the averaged-vs-instantaneous classification the time axis needs. A test fails if a field is added to `NWPCloudDiagnostics` without being classified, so this cannot drift again.

## Filed but deliberately not fixed here

- **#481** — GFS averaged cover labelled at its window endpoint (0.5–3 h lag). Always-on and larger than the interpolation defect, but every fix costs something (relocating anchors, or decoupling cover from its averaged-only geometry) and it touches `meteorology-decisions`. Needs a decision and a measured before/after, not a patch.
- **#486** — ICON ISA height anchor. The audit reported this as a terrain error worth "thousands of feet". That mechanism is wrong: `pressure_hpa_to_altitude_m` maps *pressure* to altitude and the anchor level's pressure already encodes terrain (795 hPa → 1999.6 m). The real error is the MSLP/temperature-driven deviation from ISA, order 100–300 m, and it is a uniform column offset that leaves layer thicknesses intact. Filed with the corrected reasoning and low priority.
- **#487** — the freezing-level fallback chain mixes ECMWF `deg0l` (AGL) with sounding-derived MSL values. Found while checking #485; needs a meteorology decision.

## Testing

Full suite: **4113 passed, 9 skipped**.

New `tests/test_grib_interpolation_audit.py` covers all four fixed defects, including the exact reported failure cases (+154h bracketing, −0.10° seam) plus the edge cases the fixes introduce: no-lower-bracket, past-horizon, empty step list, short `sfc_data`, genuine-zero accumulation, regional-grid no-op, and averaged-vs-instant fraction routing.


roznet:
Thanks — all four findings verified and acted on in f41d1b12. Nothing deferred.

**1. Greenwich seam missed the vectorised path — confirmed, and the PR description was wrong.**

You're right, and this was the important one. I checked: `decode_grib_per_point` does not route through `_interpolate_per_point` — it goes `_decode_pressure_vars_from_datasets` → `_bilinear_grid_weights` → `_frac_grid_indices`, which drops out-of-range targets via `np.interp(..., left=nan, right=nan)`. Same defect, different mechanism. And it's live: `decode_gfs_pressure` is dispatched from `_fetch_clwmr_icmr_for_fhour` (`__init__.py:2713`), so GFS CLWMR/ICMR really was still returning `None` at the seam. My "all three call sites benefit" claim in the PR description was simply incorrect.

I did **not** route it back through `_interpolate_per_point` — that xarray loop was deliberately replaced with the numpy gather because it was GIL-bound (#441), and undoing that to fix a seam would be a bad trade. Instead the numpy path closes the same seam its own way: extend the longitude axis one step past the end for the fractional-index lookup, then fold the resulting column index back with `% W`, since column W is column 0. The gather indexes the real data array, so wrapping the index is the equivalent of the xarray path's appended column.

Cyclic detection is now a single shared `_is_cyclic_longitude` rather than duplicated in two places.

New tests cover the numpy path directly, including `j0 == last column, j1 == 0` at the wrap, and a cross-check asserting both paths agree to 1e-9 on the same seam point. Regional grids are still asserted both to interpolate normally *and* to keep reporting out-of-bounds outside their domain, so the wrap can't silently resurrect out-of-domain points.

**4. Far-lead-time widths weren't ground-truthed — fair hit, and now checked.**

Correct that `f168`/`f384`/the f199 invariants were asserting self-consistency of the new formula, not ground truth — exactly the failure mode #480 was. Checked against the live index:

```
f180 → 174-180    f240 → 234-240    f384 → 378-384
```

All width 6; the formula holds to the end of the run. Added to the test table with a comment recording that these are read off the index rather than extrapolated.

**2. `meteorology-decisions.md` §3 — fixed.**

Agreed this was the more damaging of the two doc issues, since CLAUDE.md points readers there first for the rationale. §3 now carries the corrected reset-at-every-6 model and the pt11 f132 window as 06–12 Z, plus an explicit dated note that the previous 1/2/3-hour description was an unverified assumption — so the next reader sees *why* it changed rather than just a different number. Also fixed the stale "publishes only the time-averaged form" phrasing, since an instantaneous form does exist for cover; it's the PRES/TMP geometry that's averaged-only, which is the actual reason `_PREFER_AVERAGED_PAIRS` picks the averaged cover.

**3. `fetch.md:256` — fixed.** Now describes margin-union-brackets and notes why the constant stays 3 h.

Full suite: **4129 passed, 9 skipped** (up from 4113 — 16 new tests).

roznet:
Second review round + a parallel external review, both addressed in 39b698eb. Two real bugs found, both mine.

**Exact-step departures had no accumulation predecessor** (external review) — the genuine one, and the bot's round-2 pass missed it too.

`_select_ecmwf_window_steps` picked the lower bracket with `valid_time <= flight_start`. A departure landing *exactly* on a forecast step therefore selected that step as its own predecessor:

```
dep = +156h  →  [156, 162]     # 150 missing
dep = +150h  →  [150, 156]     # 144 missing
```

So `tp`/`sf`/`cp` had nothing to difference against for the departure hour — the same class of defect as #483, reintroduced by my own fix for it. Now strict `<`, giving `[150, 156, 162]`; the exact step is still retained by the margin window, so nothing is lost.

Worth naming why this slipped past my tests: every departure hour in `test_every_departure_in_the_six_hourly_range_is_bracketed` sat *between* steps (146, 148, 151, 154, 158, 160, 165). The boundary case was never exercised. Added coverage for exact-step departures in both the 6-hourly and hourly/3-hourly ranges, plus an assertion that the exact step is still kept.

**Spatial fill set the diagnostics but not the field sounding reads** (external review) — also real.

#485 made `freezing_level_ft` interpolate spatially, but `interpolate_diagnostics_spatially` assigned `nwp_cloud_diagnostics` directly, while sounding analysis reads `HourlyForecast.freezing_level_m` to populate `indices.nwp_freezing_level_ft`. A spatially filled interior point ended up with a freezing level in the diagnostics and `None` in the field that actually matters — so #485 was only half a fix.

Rather than duplicate the mirroring, I lifted what `_apply_cloud_diagnostics` was already doing onto `HourlyForecast.attach_nwp_diagnostics`, and both paths now go through it. Same reasoning as the shared field inventory: make it structurally impossible to apply on one path and forget on the other. Tested end-to-end through `interpolate_diagnostics_spatially`, plus the "don't blank an existing Open-Meteo value when the diagnostics carry none" case.

**On the external review's first finding** (pressure-level seam still open): that was already fixed in f41d1b12 — that reviewer read the first commit. Its diagnosis was correct for `cfd8cca6` and matches what you flagged in round 1. I added the end-to-end test through `_decode_pressure_vars_from_datasets` it asked for, which confirms a `-0.10` target now returns `covered=True` with the correct value instead of an empty dict — previously only `_bilinear_grid_weights` was tested directly.

**Your round-2 items:**

- `cloud-layers-analysis.md:340` — fixed, `≤ 1.5 h` → `≤ 3 h`, with a note on where the old bound came from. Third doc carrying the same disproven claim; good catch that it describes the same mechanism from the desync angle.
- `decode_grib_to_points`/`_interpolate_to_points` — **deleted** rather than routed through `_close_cyclic_longitude`. Confirmed dead repo-wide (not just `src/` and `tests/`). Since the whole point of this audit is removing latent drift, leaving a dormant copy of the seam bug behind a resurrection risk seemed worse than a 137-line deletion.

Full suite: **4146 passed, 9 skipped**.

roznet:
Confirmed and fixed in 5a31e298 — this was a real active bug, and your diagnosis was right including the "runs on every request" part.

**You found three sites. There were four.**

`_gate_gfs_hourly` (`fill.py:621`) also assigned directly — and it was worse than the other three, because it rebuilt the diagnostics from a **hand-written field list**:

```python
h.nwp_cloud_diagnostics = NWPCloudDiagnostics(
    low=..., mid=..., high=...,
    convective_cover_pct=..., convective_base_ft=..., convective_top_ft=...,
    total_cover_pct=..., boundary_cover_pct=..., ceiling_ft=...,
    freezing_level_ft=...,
)
```

That list omits `convective_precip_mm_h`, `k_index`, `total_totals`, `ml_cape_jkg`, `ml_cin_jkg` — so the RH/condensate gate silently dropped all five #283 fields whenever it fired. Same field-list drift as #485, in a third location, and neither review round caught it. Replaced with `model_copy(update={low, mid, high})` so only the layer bands are touched and nothing can go stale by omission.

All four now route through `attach_nwp_diagnostics`.

**The docstring you quoted was the actual problem.** You're right that "the mirror cannot be applied on one and forgotten on the other" was incomplete — worse, it was aspirational prose asserting an invariant nothing enforced, which is exactly how the same omission then shipped three separate times in one PR (spatial fill → time-axis fill → RH gate).

So rather than fix the four sites and re-assert the claim, I made it checkable: `test_no_direct_diagnostics_assignment` scans the package source (271 files) and fails on any `.nwp_cloud_diagnostics =` outside `models/analysis.py`. I verified it isn't vacuously passing — the regex matches a direct assignment and correctly ignores `==` comparisons. The docstring now describes the enforced rule and drops the false guarantee.

Also added the two regression tests you suggested: `_fill_diag_hourly` mirroring on a forward-filled gap hour (with a stale Open-Meteo value pre-seeded, so it fails without the fix), and non-layer field preservation through the RH gate.

On the GFS sites being inert — agreed, and I fixed them anyway for the reason you gave. Worth noting the RH-gate case was *not* inert in the same sense: the dropped #283 fields are inert only because GFS doesn't decode them today, but ICON/ECMWF diagnostics do carry them, so that constructor was one code-path change away from live data loss.

Full suite: **4149 passed, 9 skipped**.

roznet:
All three cleared in 46cb65a1, plus two things I found auditing my own changes.

**Finding 2 was itself an instance of the pattern this audit is about** — I fixed `cloud-layers-analysis.md:340` and missed line 100 in the same file. So rather than patch the two sites you named, I grepped the repo for the disproven model. That found a **fifth** location neither of us had: `designs/future/known-issues.md:75-76`, which still described "1/2/3 h windows ... always 3 h past f120" in the write-up of the very incident (#148/pt11) that motivated this code.

Corrected: `fill.py` module docstring, `_interp_gfs_diag_hourly` docstring, `cloud-layers-analysis.md:100`, `known-issues.md:75`. Both source docstrings now defer to `_gfs_window_length_hours` as the single owner of the arithmetic and its live-index provenance, instead of restating the numbers — restating is what let five copies drift apart in the first place.

Left alone deliberately: the notes in `meteorology-decisions.md:287-290` and `time-alignment-audit.md:133` that describe the old model *as superseded*. Those are historical annotations, not stale claims.

**Finding 3** — both dead imports removed. Confirmed the third `NWPCloudDiagnostics` import in `fill.py` (in `_interp_diag_at`) is still live.

**Two things I found myself:**

*A memory regression I introduced.* `_close_cyclic_longitude` concatenated a wrapped column on **every** call, copying the whole field — 8.3 MB for a global 0.25° grid — once per variable per dataset, inside decode workers that run 3-up under this project's known memory pressure. Only Greenwich-crossing routes need it. It now short-circuits unless a target actually lands in the seam band: measured 54 ms → 0.5 ms per 20 calls, seam behaviour unchanged and still tested both ways. The numpy path's equivalent append is 1-D (~11 KB) so it stays unconditional.

*Two unguarded field lists.* `test_inventory_partitions_every_model_field` treats `low`/`mid`/`high` as opaque, so a field added to `NWPCloudLayerDiag` would be dropped by **both** layer interpolators with nothing to catch it — the #485 shape one level down. `_interp_levels_at` rebuilds `PressureLevelData` field-by-field with the same exposure. Both are correct today; the two new tests keep them that way.

Full suite: **4153 passed, 9 skipped**.

Issue #494, opened 2026-07-25
iOS/iPad: airport condition cards compress to per-character text wrapping in regular width


roznet:
## Summary

On iPad (regular width) the departure/arrival cards in `AirportConditionsView`
are squeezed so hard the text wraps one character per line. Found incidentally
while porting #492; **confirmed pre-existing** by screenshotting the Advisory tab
with the #492 section unwired.

## Symptoms (iPad Pro 11", mock mode, Advisory tab)

- The wind value `200@8kt` wraps vertically, one character per line
- The `VFR` category badge wraps to `V` / `FR`
- The section label `Departure` hyphenates to `Depar-` / `ture`
- Arrival card identical

## Cause

`AirportConditionsView` (`Views/Briefing/AirportConditionsView.swift:16`) puts the
two cards in an `HStack` when `sizeClass == .regular`. Inside each card,
`conditionRow` lays out a fixed 60pt model column, the category badge, a wind
`Label`, the runway text and a `Spacer` in one `HStack` — the intrinsic width of
that row exceeds half the iPad detail pane, so SwiftUI compresses the `Text`
views to the point of per-character wrapping instead of the row shrinking or
wrapping as a unit.

Note the detail pane is ~800pt wide, so there is nominally room for two ~380pt
cards — the failure is in the row's own compression behaviour, not the split.

## Suggested fix directions

- `.fixedSize(horizontal: true, vertical: false)` (or `.lineLimit(1)` +
  `.layoutPriority`) on the wind/badge cells so they keep their intrinsic width
  and the `Spacer` absorbs the slack
- and/or raise the regular-width threshold so two-across only applies when the
  pane is genuinely wide enough, falling back to the stacked `VStack`
- and/or drop the fixed `width: 60` model column in favour of intrinsic sizing

## Verification

Reproduce with the existing mock harness on an iPad destination and look at the
Conditions section of the Advisory tab; no backend or auth needed.

Issue #493, opened 2026-07-25
iOS: port the route SIGMET section (D-0 area hazards)


roznet:
## Summary

Route SIGMETs are rendered on web but **entirely absent from iOS** — no view, and
no DTO. Discovered while porting the sibling observations section (#492).

SIGMETs are the D-0 *area hazard* counterpart to METAR/TAF: same fetch cycle, same
realtime refresh seam, but area-polygon-keyed rather than airport-keyed, and with
no model-comparison analog (they are presented, not reconciled).

## Current state

- **Backend**: complete. `route_sigmets` is on `ForecastSnapshot`
  (`models/observations.py::RouteSigmets`), populated by `run_route_sigmets()`, and
  refreshed by `run_realtime_refresh()` alongside METAR/TAF.
- **Wire**: already served. `GET .../snapshot` returns raw `briefing.json`, so
  `route_sigmets` reaches the app today.
- **Web**: `briefing-ui.ts::renderRouteSigmets` — a table (Hazard, FIR, Levels,
  Enroute, Move, raw text) with a per-row info popup, `sigmets-section`.
- **iOS**: nothing. `SnapshotResponse` has no `routeSigmets` field, so the block is
  dropped at decode.

So this is a pure client gap — no backend or endpoint work.

## Scope

- [ ] `RouteSigmets` + `SigmetAlongRoute` Codable types on `SnapshotResponse`
- [ ] A `RouteSigmetsView` in the Advisory tab (natural slot: after Observations,
      before Alternates) + scroll-spy entry, gated on a non-empty SIGMET list
- [ ] Per-SIGMET detail sheet with the raw text
- [ ] Fixture coverage so mock mode renders it, plus a decode test

## Design notes

- `has_severe` and the `hazards` union make a good summary-bar signal, mirroring
  how #492 uses `worst_metar_category` + `phenomena_along_route`.
- Same **responsive** question as #492: the web row is wide (6 columns). Levels
  (`base_ft`/`top_ft`) and movement (`direction`/`speed_kt`) are the compressible
  pair on a phone.
- The model deliberately retains the polygon `coords`, the
  `enroute_distance_from/to_nm` span, and the vertical band so a later feature can
  draw SIGMETs on the route map and cross-section — see
  `designs/metar-taf-route-weather.md` ("Future cross-section / map overlay").
  Out of scope here; this issue is the table only.

## Verification note

D-0 only, like observations — so on-device verification needs a same-day flight
[...]

PR #495, opened 2026-07-25
iOS: Current Observations section (METAR/TAF/model comparison) with responsive axis toggle + phone row cap

### LINKED ISSUE #492: iOS: port the Current Observations section (METAR/TAF/model comparison) with a responsive category/wind axis toggle
## Summary

The web briefing has a **Current Observations** section (METAR / TAF / model
comparison per airport along the route) that the iOS app does not render at all.
It belongs between **Conditions** and **Alternates** — the same slot it occupies
on web.

This is an **undocumented** gap, not a deliberate one: it is absent from
`designs/future/ios-web-known-gaps.md`, which lists only the omissions we chose
(advisory recalculate, profile editing). The data is already on the wire and
partially decoded on iOS, then dropped on the floor.

## Current state

`SnapshotResponse.routeObservations` **is** decoded (`Models/API/SnapshotResponse.swift:11`)
but is **never referenced by any view** — a dead field. `grep -rn routeObservations Views/ ViewModels/`
returns nothing.

No backend work is required. `GET /api/flights/{id}/packs/{ts}/snapshot`
(`api/packs.py:2386`) returns raw `briefing.json` via `FileResponse`, so every field
below is already being served — iOS just doesn't model or draw it.

### DTO gap (`Models/API/SnapshotResponse.swift:65-89`)

`ObservationComparison` is **missing entirely** — which is the whole point of the
section (the obs-vs-model reconciliation).

Missing on `RouteObservations`:
`airports_found`, `airports_with_metar`, `airports_with_taf`, `comparisons`,
`has_conflicts`, `phenomena_along_route`

Missing on `AirportObservation`:
`enroute_distance_nm`, `nearest_waypoint_icao`, `metar_time`, `metar_visibility_m`,
`metar_wind_dir`, `metar_qnh`, `taf_trend_type`, `taf_wind_dir`, `taf_wind_speed_kt`,
`taf_wind_gust_kt`, `taf_applicable_text`, `taf_applicable_lines`,
`metar_wind_advisory`, `metar_best_runway_id`, `metar_crosswind_kt`, `metar_headwind_kt`,
`taf_wind_advisory`, `taf_best_runway_id`, `taf_crosswind_kt`, `taf_headwind_kt`,
`has_metar`, `has_taf`, `eta_hour_offset`

Source of truth: `src/weatherbrief/models/observations.py`.

## Design — responsive two-axis table

The web table (`briefing-ui.ts:renderRouteObservations`, `:1217-1238`) is **11
columns**: ICAO / Dist / ETA, then a 4-column **Condition** group
(METAR · TAF · Model · agreement) and a 4-column **Wind** group
(METAR · TAF · Model · agreement). Too wide for a phone.

**Adaptive rule** — matches the existing precedent in
`Views/Map/RouteMapView.swift:37` (`usesDualMetrics = horizontalSizeClass == .regular`):

- **Compact (iPhone)** — a segmented `Category | Wind` picker at the top of the
  section; render only the selected group (3 fixed + 4 = 7 columns). Default
  `Category`.
- **Regular (iPad)** — render **both** groups, no picker (the toggle would be
  pointless), reproducing the web's two-row grouped header.

## Scope

- [ ] Extend the iOS DTO to the full backend shape, incl. `ObservationComparison`
- [ ] New `Views/Briefing/RouteObservationsView.swift`: summary bar (METAR/TAF
      counts, corridor, worst category, phenomena, fetch time), conflict banner,
      adaptive table, per-airport detail sheet
- [ ] Per-airport detail sheet (the web's ⓘ popup): raw METAR, raw TAF with
      `taf_applicable_lines` highlighted, per-source wind summary with best
      runway + crosswind
- [ ] Wire into `AdvisoryTabView` between `conditions` and `alternates`, plus a
      scroll-spy section entry (`:143-150`)
- [ ] Extend `FixtureBriefingData.swift` `route_observations` with `comparisons`
      + the wind/advisory fields so mock mode renders the section
- [ ] Decode test covering the full shape

### Deliberately out of scope

- **Route SIGMETs** — a sibling D-0 surface (`route_sigmets`, web
  `renderRouteSigmets`) that iOS is *also* missing entirely. Separate gap,
  separate issue; do not fold in here.
- **Observations-only refresh button** — web has one (`POST .../observations/refresh`).
  On iOS the toolbar Refresh already routes through the tiered `decide_refresh`
  gate, which at D-0 performs exactly this cheap realtime refresh, so an
  in-section button is near-redundant. Revisit if it reads as missing.

## Verification note

This section is **D-0 only** (`days_out == 0`) — TAFs rarely cover the next day,
so the pipeline doesn't fetch observations otherwise. On-device verification
therefore requires a same-day flight; mock mode is the practical check for
layout, which is why the fixture extension is in scope.

## Incidental doc fix

`designs/metar-taf-route-weather.md` documents the wind advisory values as
`OK/CAUTION/WARNING`. The actual wire values from
`_wind_advisory_status()` (`tasks/route_weather.py`) are lowercase
`green`/`amber`/`red`. The web reads them correctly; the doc is stale.

roznet:
Ports the web briefing's **Current Observations** section (METAR/TAF/model
comparison) to iOS, where it was missing entirely.

Closes #492

## The gap

`SnapshotResponse.routeObservations` **was already decoded** but referenced by no
view — a dead field. Worse, the DTO silently dropped the part that carries the
point of the section: `comparisons`, every runway wind advisory, and the ETA.

Judging this an oversight rather than a deliberate omission: it isn't recorded
anywhere as a decision. (I originally cited `designs/future/ios-web-known-gaps.md`
here — the review bot correctly points out that file isn't in the repo. It exists
only as an uncommitted file in my working tree, from separate in-flight work, so
it's invisible to the repo and to this PR. Not committed here because it isn't
mine to commit.)

**No backend change.** `GET .../snapshot` returns raw `briefing.json`, so the
full shape was already on the wire.

## Design

Sits between Conditions and Alternates with a scroll-spy anchor, matching web.
Gated on an airport having actually reported, so the anchor never renders empty.

**Responsive axis.** The web table is 11 columns (ICAO/Dist/ETA + a 4-column
Condition group + a 4-column Wind group), which doesn't fit a phone:

| Width | Behaviour |
|---|---|
| Compact (iPhone) | Segmented Condition / Wind picker; one 7-column group at a time |
| Regular (iPad) | Both groups side-by-side with the web's grouped header; no picker |

Keyed off `horizontalSizeClass`, matching `RouteMapView`'s existing dual-metric split.

**Phone row cap.** Verified against real data: a 30 nm corridor on EDDC→EGTF
returned **29 airports with a METAR**. Web renders all of them — fine in a
browser window, a very long scroll in the Advisory tab. Compact width shows the
**10 nearest** with a "Show all 29 airports" toggle.

Two decisions there worth a reviewer's eye:
- Selection is by distance but **rendering stays in route order**, so the table
  still reads departure→destination instead of jumping around the route.
- Any `CONFLICTING` comparison (category **or** wind) is **pinned into the capped
  set** regardless of distance — otherwise the "observations conflict with the
  model" banner could point at a row the cap had hidden. Consequence: the
  collapsed view can exceed 10 rows, which is why the collapse label reads
  "Show fewer" rather than naming a count that would sometimes be wrong.

## Deliberate divergences from web

- **Wind cells show the crosswind in kt** coloured by advisory status, not web's
  bare `G`/`A`/`R` letter. Web puts the value in a hover tooltip; touch has no
  hover, so the value has to *be* the label. A legend states the unit, the colour
  meaning, and the ✓/⚠/✗ agreement key.
- **No in-section refresh button.** Web has one, but on iOS the toolbar Refresh
  already routes through the tiered `decide_refresh` gate, which at D-0 performs
  exactly this cheap realtime refresh — so it'd be near-redundant.
- **Row cap** (above).

## Two rendering bugs the screenshots caught

Both invisible to passing assertions:

1. `Grid` applies a row background **per cell**, so a CONFLICTING row's amber
   highlight tiled into a dashed line of blocks. Fixed with
   `Grid(horizontalSpacing: 0)` + the gutter inside each cell.
2. A missing flight category rendered as a **grey filled badge** containing "—",
   carrying the same visual weight as a real rating — it read as though "—" were
   itself a category. Now plain muted text, matching `WindAdvisoryBadge`.

## Verification

- **24 unit tests** — decode contract against the Python source of truth, the
  comparison join, the reporting filter, and 8 covering the cap (under/at/over
  limit, route-order preservation, conflict pinning incl. wind-only, nil distance
  sorting last, silent fields not consuming slots, non-positive limits).
- **2 XCUI journeys** asserting the compact-vs-regular contract and the detail sheet.
- **Visually confirmed on iPhone 17 Pro and iPad Pro 11"** — including against the
  real 29-airport dataset (temporarily spliced into the fixture, not committed),
  which exercised cases the fixture doesn't: airports with no runway wind data at
  all, and non-contiguous `taf_applicable_lines` like `[0, 2]`.
- Full suite green.

Fixture extended with comparisons + wind advisories + a CONFLICTING case, so mock
mode renders the section (real data today was benign — all VFR/green/CONFIRMING —
so the fixture is what covers the amber/red/conflict paths).

## Incidental doc fix

`designs/metar-taf-route-weather.md` documented the wind advisory values as
`OK/CAUTION/WARNING`. `_wind_advisory_status()` actually emits lowercase
`green`/`amber`/`red`. Web reads them correctly; the doc was stale.

## Follow-ups filed, not folded in

- **#493** — route SIGMETs are missing from iOS entirely (no DTO, no view). Same
  D-0 sibling surface; deliberately separate.
- **#494** — pre-existing iPad bug: airport condition cards compress to
  per-character text wrapping. Confirmed pre-existing by unwiring this section
  and re-screenshotting, so it isn't caused by this PR.

### DIFF
--- a/app/flyfun-weather/flyfun-weather/Models/API/SnapshotResponse.swift
+++ b/app/flyfun-weather/flyfun-weather/Models/API/SnapshotResponse.swift
@@ -86,8 +86,14 @@ struct RouteObservations: Codable, Sendable {
 
     /// Comparisons keyed by ICAO — the table joins each airport row to its
     /// model reconciliation.
+    ///
+    /// `uniquingKeysWith` rather than `uniqueKeysWithValues`: this is
+    /// server-derived JSON, and the trapping initializer would `fatalError` on a
+    /// duplicate ICAO — crashing the briefing screen over a malformed payload
+    /// instead of degrading. First entry wins, matching `HelpCatalogResponse`,
+    /// `SkewTVariableCatalog` and `TimingScenariosView`.
     var comparisonsByIcao: [String: ObservationComparison] {
-        Dictionary(uniqueKeysWithValues: (comparisons ?? []).map { ($0.icao, $0) })
+        Dictionary((comparisons ?? []).map { ($0.icao, $0) }, uniquingKeysWith: { first, _ in first })
     }
 
     /// Airports that actually reported something. The web filters the table the
--- a/app/flyfun-weather/flyfun-weather/Models/API/SnapshotResponse.swift
+++ b/app/flyfun-weather/flyfun-weather/Models/API/SnapshotResponse.swift
@@ -93,8 +93,39 @@ struct RouteObservations: Codable, Sendable {
     /// Airports that actually reported something. The web filters the table the
     /// same way (`apt.has_metar || apt.has_taf`) — a spatial query can return
     /// small GA fields with no data at all.
+    ///
+    /// Server order is along-route (by enroute distance), which is the useful
+    /// reading order; everything downstream preserves it.
     var reportingAirports: [AirportObservation] {
         (airports ?? []).filter { $0.hasMetar == true || $0.hasTaf == true }
+    }
+
+    /// The subset a phone-width table shows before "Show all": the `limit`
+    /// airports nearest the route, **plus** any airport whose model comparison
+    /// is CONFLICTING — the conflict banner must never point at a row the cap
+    /// hid. Returned in route order, not distance order, so the table still
+    /// reads departure→destination.
+    ///
+    /// A 30 nm corridor on a long route routinely reports 30+ fields, which is a
+    /// reasonable table in a browser window but a very long scroll inside the
+    /// iOS Advisory tab.
+    func nearestReportingAirports(limit: Int) -> [AirportObservation] {
+        let reporting = reportingAirports
+        guard limit > 0, reporting.count > limit else { return reporting }
+
+        // Missing distance sorts last rather than winning the "nearest" race.
+        let byDistance = reporting.sorted {
+            ($0.distanceFromRouteNm ?? .greatestFiniteMagnitude)
+                < ($1.distanceFromRouteNm ?? .greatestFiniteMagnitude)
+        }
+        var keep = Set(byDistance.prefix(limit).map(\.icao))
+
+        let conflicting = comparisonsByIcao
+            .filter { $0.value.categoryMatch == "CONFLICTING" || $0.value.windAdvisoryMatch == "CONFLICTING" }
+            .keys
+        keep.formUnion(conflicting)
+
+        return reporting.filter { keep.contains($0.icao) }
     }
 }
 
--- a/app/flyfun-weather/flyfun-weather/Models/API/SnapshotResponse.swift
+++ b/app/flyfun-weather/flyfun-weather/Models/API/SnapshotResponse.swift
@@ -62,28 +62,110 @@ struct ThermodynamicIndicesSummary: Codable, Sendable {
     let nwpCeilingFt: Double?
 }
 
+/// D-0 METAR/TAF observations along the route, plus the obs-vs-model
+/// reconciliation. Mirrors `models/observations.py::RouteObservations`
+/// (source of truth) and the web's `renderRouteObservations`.
+///
+/// Only populated when `days_out == 0` — European TAFs rarely cover the next
+/// day, so the pipeline skips the fetch entirely on D-1+.
+///
+/// SYNC: `web/ts/managers/briefing-ui.ts` (`renderRouteObservations`) renders
+/// the same shape; `src/weatherbrief/models/observations.py` defines it.
 struct RouteObservations: Codable, Sendable {
     let corridorNm: Double?
     let fetchTime: String?
+    let airportsFound: Int?
+    let airportsWithMetar: Int?
+    let airportsWithTaf: Int?
     let airports: [AirportObservation]?
+    let comparisons: [ObservationComparison]?
     let worstMetarCategory: String?
     let worstTafCategory: String?
+    let hasConflicts: Bool?
+    let phenomenaAlongRoute: [String]?
+
+    /// Comparisons keyed by ICAO — the table joins each airport row to its
+    /// model reconciliation.
+    var comparisonsByIcao: [String: ObservationComparison] {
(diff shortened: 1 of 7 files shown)


roznet:
Thanks — actioned the Important finding, plus one thing the review missed.

**Fixed (38d2fdf3):**

1. **`comparisonsByIcao` trapping initializer** — correct and a real defect. Switched to `uniquingKeysWith: { first, _ in first }`, matching the three sites you cited. Worth adding: it was the *only* `uniqueKeysWithValues` call in the app, so it was a genuine outlier rather than one convention among two. Added a regression test that decodes a duplicated-ICAO payload and asserts first-wins — note that before the fix that test would have taken down the test process rather than failing, which is a fair illustration of the severity.

2. **Not flagged, but adjacent and in the same lines:** `comparisonsByIcao` is a computed property and was being read *inside* the row `ForEach`, so the entire dictionary was rebuilt once per row — 29 rebuilds per render pass on the real 30 nm corridor I verified against. Hoisted to once per render. Batched in here since the push was already paid for.

**Minor findings:**

3. **`/sync-ios-web` pass** — done for this PR's scope: `metrics-catalog.json` is byte-identical between web and iOS, and `git diff --name-only main..HEAD` confirms this branch touches **no** parity surface other than `SnapshotResponse.swift` (which, as you note, is now strictly closer to the backend shape). Nothing to reconcile.

4. **`designs/future/ios-web-known-gaps.md` doesn't exist** — good catch, and you're right. It exists only as an *untracked* file in my working tree, from separate in-flight work, so it's invisible to the repo and to git history. Not committing it here since it isn't part of this change. I've corrected the PR description, which overstated it as an existing doc.

PR #497, opened 2026-07-25
feat(ios): delete a flight from the list, plus link/share polish

roznet:
Two independent iOS improvements that happened to grow in the same files, so they ship together.

## Delete a flight from the list

Parity with the web's Delete button — iOS previously had no way to remove a flight.

- **Trailing-swipe Delete next to Edit**, plus a matching context-menu item.
- Gated exactly like Edit: **owner-only** (the server 404s a subscriber's delete — a subscriber gets Unsubscribe instead) and **online-only**.
- **Never one-tap**: opens a destructive confirmation naming the route ("LFMD → LFML and all of its briefing history will be deleted. This can't be undone."), mirroring the web's `confirm()`. `allowsFullSwipe` is already false on that edge, so a long swipe can't fire it either.
- `deleteFlight(id:)` on `BriefingRepository` → `DELETE /api/flights/{id}` (the endpoint the web already uses).
- **Cache ordering matters**: the caching layer deletes server-side *first*, and only then evicts the flight's packs and sidecar directory. A failed delete leaves a flight that still exists with its offline data intact.
- `FlightListViewModel.deleteFlight` re-syncs the list, so the row, its offline badge, and its Spotlight entry drop together. It rethrows so the view surfaces the failure — a row that stays put must not read as a silent no-op. The iPad detail selection is cleared if it pointed at the deleted flight.

## Universal Link + share polish

- A `briefing.html?flight=<id>` link for a flight **outside the viewer's list** (a feedback email about another pilot's flight, a link from a different sign-in) now fetches it directly instead of silently landing on the list. A non-owner comes back `role == .subscriber` and gets the same preview-before-subscribe screen as a `/s/{code}` share link; a 404 surfaces an alert. Also tolerates the feedback email's unencoded `&t=<pack timestamp>`.
- Share sheet now uses `UIActivityViewController` with a custom **"Open in Safari"** activity — iOS offers Chrome (via its share extension) but never Safari on its own, so the briefing URL had no way back to Safari.
- Observations table: hoist `displayedAirports` out of the row builder so the nearest-N sort runs once per render instead of twice.
- `designs/future/ios-web-known-gaps.md` records the deliberate web-only gaps (advisory recalculate, profile editing) with reasoning, so a future "should we build this on iOS?" starts from context.

## Testing

`xcodebuild build` clean; full `flyfun-weatherTests` suite passes on iPhone 17 Pro (**TEST SUCCEEDED**).

New tests:
- `deleteFlightCallsRepositoryThenReloads` — repository called once, list re-synced.
- `deleteFlightFailureRethrowsAndKeepsRow` — rethrows, list untouched, no reload.
- `successfulDeleteEvictsEveryPackAndTheSidecarDirectory` — all packs + sidecars gone, other flights untouched.
- `failedDeleteKeepsCachedPacks` — server failure leaves the cache intact.
- `routesFeedbackEmailLinkWithPackTimestamp` — `&t=` tolerated.

**Not verified**: the swipe gesture itself on a device/simulator — there are no existing swipe-action UI tests to extend. The swipe and confirm buttons carry `deleteFlightSwipeButton` / `confirmDeleteFlightButton` identifiers if we want one later.

### DIFF
--- a/app/flyfun-weather/flyfun-weather/Services/BriefingRepository.swift
+++ b/app/flyfun-weather/flyfun-weather/Services/BriefingRepository.swift
@@ -12,8 +12,19 @@ enum RefreshSource: String, Sendable {
 /// Abstraction over briefing data access.
 protocol BriefingRepository: Sendable {
     func flights() async throws -> [FlightResponse]
+    /// Fetch one flight by id, independent of the viewer's list. Backs the
+    /// `briefing.html?flight=<id>` Universal Link fallback for flights that
+    /// aren't in the list (e.g. a feedback email about another pilot's flight):
+    /// the server allows the owner always and any authenticated viewer for a
+    /// non-private flight (role comes back `subscriber`), and 404s otherwise.
+    func flight(id: String) async throws -> FlightResponse
     func createFlight(_ request: CreateFlightRequest) async throws -> FlightResponse
     func updateFlight(flightId: String, request: UpdateFlightRequest) async throws -> UpdateFlightResponse
+    /// Permanently delete one of the viewer's own flights and all of its briefing
+    /// history (server 204, owner-only — a subscriber unsubscribes instead).
+    /// Online-only; the caching layer also drops the flight's local pack cache,
+    /// since a deleted flight can never be refreshed or re-downloaded.
+    func deleteFlight(id: String) async throws
     func aircraft() async throws -> [AircraftResponse]
     func profiles() async throws -> [ProfileResponse]
     /// Account usage summary — the iOS app reads only the durable timing-scan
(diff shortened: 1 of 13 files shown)


roznet:
Merged with `--rebase`. All five review findings were triaged as non-blocking (no correctness/security/data-loss issues), so rather than spending another review round they were folded into main directly as `8ef2e02a`:

- **Important #1** — the out-of-list Universal Link fallback now differentiates 404 (private/deleted/other account) from a transient failure, matching the `.share` case's error split.
- **Important #2** — `CachingBriefingRepository.flight(id:)` drops its unreachable cached-list fallback; the error now propagates so the caller can tell offline from 404.
- **Important #3** — the flight list's "Share Flight" context-menu item now uses `ShareActivitySheet` (with `OpenInSafariActivity`) instead of `ShareLink`, so both share entry points behave the same.
- **Minor #1** — `guard let repo` moved before `clearPendingNavigation()` so a link arriving pre-repository resumes after sign-in.
- **Minor #2** — corrected the `advisories-ui.ts` line citations in `ios-web-known-gaps.md`.

Full `flyfun-weatherTests` suite passes on iPhone 17 Pro.

PR #500, opened 2026-07-26
Add durable refresh tracking and resume after container restart

### LINKED ISSUE #499: Durable briefing-refresh tracking + resume interrupted refreshes after a container restart
## Problem

Briefing refresh state is **in-memory only**. `_RefreshRegistry` (`src/weatherbrief/api/packs.py:120-300`) holds a `dict[flight_id, RefreshEntry]` and the pipeline runs on a module-level `ThreadPoolExecutor(max_workers=2)`. If the container dies mid-refresh (OOM kill, deploy, crash), that state evaporates and **nothing in the DB or on disk records that a refresh was ever in flight**:

- The pack row is only written at the end (`_persist_pack_finalize`), or at the `briefing_ready` milestone for the SSE path only (`_persist_pack_provisional`, packs.py:1571).
- `_prepare_refresh` mkdirs the pack dir up front (packs.py:1122), so a killed refresh leaves an orphan directory with partial artifacts and no row pointing at it.
- Web SSE just errors out; iOS polling `/refresh/status` gets `{"active": false}` and no new pack — indistinguishable from "never asked".
- Recovery today is accidental: the scheduler picks the flight up at its *next* auto-refresh slot, if it happens to be due. Manual / Siri / MCP refreshes are simply lost.

The 2026-07-23 05:09Z OOM (#490) is the motivating case: a user briefing was killed and there was no record of it, no retry, and nothing to post-mortem.

## What makes this cheap

1. **The registry is already the single choke point.** All three refresh paths (sync `POST /refresh`, SSE `/refresh/stream`, scheduler `process_auto_refreshes`) go through `try_register` → `set_refreshing` → `update_progress` → `unregister`. Durability is a write-through mirror at four call sites, not a rewrite.
2. **Single uvicorn worker** (`Dockerfile:62`, no `--workers`). So no leases, no instance IDs, no distributed locking: *any* non-terminal row present at process boot is by definition an orphan.

### 1. Durable job row (migration 082, `briefing_refresh_jobs`)

`flight_id`, `user_id`, `triggered_by` / `source`, `as_of_date`, `status` (`queued|running|succeeded|failed|abandoned`), `attempt`, `created_at` / `started_at` / `finished_at` / `heartbeat_at`, `pack_path`, `last_error`.

The registry is keyed by `flight_id`, so at most one refresh per flight is ever live — two columns on `flights` would technically do. A separate table is preferred anyway: when the container OOMs, the record of what was in flight is exactly the post-mortem that wasn't possible for #490.

### 2. Write-through in `_RefreshRegistry`

- `try_register` → insert `queued`
- `set_refreshing` → `running` + `started_at`
- `update_progress` → bump `heartbeat_at` (already called from `progress_callback`)
- `unregister` → terminal status

Strictly best-effort inside `try/except`: a DB hiccup must never fail a refresh. The crash case is precisely the one where the `finally: unregister` does *not* run — that's the mechanism, not a bug.

### 3. Boot-time reconciliation + resume

Wired into `lifespan` like the other loops, after a startup delay (~90s, past the scheduler's own delay). For each orphan row (non-terminal at boot):

- Count the interruption as a consumed attempt. `attempt >= WB_REFRESH_MAX_ATTEMPTS` → mark `abandoned`, log loudly, stop.
- Flight deleted / already departed / pending-coverage → `abandoned`, no retry.
- Otherwise re-run `decide_refresh` — if the gate no longer says `full` (the scheduler already re-briefed it, or model data hasn't moved), close the row instead of burning compute. The gate is a free correctness check.
- Else re-queue via `_auto_refresh_one`-style reuse with `triggered_by="resume"`, `attempt + 1`.

Resumes register in the registry, so `idle_seconds()` goes to 0 and the discretionary GRIB warm loop yields to them automatically (existing #490 machinery) — that falls out for free.

### 4. Retry cap as a parameter

`WB_REFRESH_MAX_ATTEMPTS`, **default 2** = the original run plus one resume. Set it to 3 for two resumes, 1 to disable resume entirely. Sized so a briefing that genuinely does take down the container can only do it twice, while a single OOM or deploy never silently loses a user's refresh.

Deliberately **no crash attribution**: we do not try to distinguish "this briefing caused the OOM" from "something else did" (the GRIB precache and standalone cycle are the likelier culprits, and no briefing has been observed crashing the container). Any interruption burns an attempt regardless of cause. Likewise no SIGTERM-vs-SIGKILL distinction — a deploy mid-briefing burns an attempt and still gets its one resume, which is fine.

### 5. Client-visible terminal state

`/refresh/status` falls back to the job row when there's no in-memory entry, so a reload after a restart reports "interrupted — retrying (2/2)" or "gave up" instead of silently nothing. Reuse the same staleness bound as `STALE_ENTRY_SECONDS` so a leaked row can't pin a flight as permanently active. Web/iOS copy for that state can ship as a follow-on.

## Out of scope / follow-ons

- Web + iOS UI copy for the "retrying after restart" state (separate PR).
- Sweeping orphan pack directories left by killed refreshes (retention loop).
- Notifying the user on `abandoned` via the existing notify sink.

## Rough size

Migration + model ~80 lines, storage CRUD ~120, registry write-through ~60, resume task + lifespan wiring ~120, status endpoint ~40, tests (orphan detection, attempt cap, gate skip, past-flight skip, max-attempts parameter) ~200. One focused PR, no new dependencies.

roznet:
## What & why

Implements durable briefing-refresh job tracking and boot-time resume logic to handle refreshes interrupted by container restarts (OOM, deploy, crash).

**The problem:** Refresh state lived only in the in-memory `_RefreshRegistry`. When the container died mid-refresh, nothing recorded that a refresh had been in flight — the pack row is written at the end, and the pack directory created up front becomes an orphan with no row pointing at it. Recovery was accidental (scheduler's next auto-refresh slot) or nonexistent (manual/Siri/MCP refreshes were lost).

**The solution:** A write-through mirror of the registry into a new `briefing_refresh_jobs` table. Every state transition (`try_register` → `set_refreshing` → `update_progress` → terminal status) is recorded durably. At process boot, a one-shot reconciliation pass identifies non-terminal rows (orphans), decides whether to resume or abandon each one, and re-queues resumable refreshes.

**Key design choices:**
- Single uvicorn worker means any non-terminal row at boot is by definition an orphan — no leases, no instance IDs, no distributed locking needed.
- Write-through is best-effort (swallows DB errors) so durability is diagnostic, not a correctness invariant.
- `flight_id` is intentionally not a foreign key so the row outlives the flight, letting reconciliation distinguish "flight deleted, don't resume" from "no record".
- Separate table rather than columns on `flights` because the registry is keyed by `flight_id` (one refresh per flight max), and the post-mortem record is the whole point.
- Resume is a separate job row (not a mutating counter) so the post-mortem keeps one row per attempt.

## Changes

**New files:**
- `tests/test_refresh_durability.py` — comprehensive test suite covering the write-through mirror, `decide_resume` logic, and boot-time reconciliation
- `src/weatherbrief/tasks/refresh_resume.py` — orphan snapshot, resume decision logic, and reconciliation loop
- `src/weatherbrief/storage/refresh_jobs.py` — CRUD + best-effort write-through helpers
- `designs/refresh-durability.md` — architecture and design rationale
- `alembic/versions/082_briefing_refresh_jobs.py` — migration creating the `briefing_refresh_jobs` table

**Modified files:**
- `src/weatherbrief/db/models.py` — added `BriefingRefreshJobRow` model
- `src/weatherbrief/api/packs.py` — `_RefreshRegistry` now supports durable mode with write-through at every state transition; added `UNCAPPED_TRIGGERS` to let scheduler and resume bypass queue caps
- `src/weatherbrief/scheduler.py` — pass `user_id` and `source="scheduler"` to `try_register`
- `src/weatherbrief/api/app.py` — explicit sweep of `BriefingRefreshJobRow` on user deletion (bulk delete doesn't emit ORM cascades)
- `src/weatherbrief/api/agent.py` — pass `source` to `try_register` for Siri/MCP attribution
- `designs/INDEX.md` and `designs/multi-user-deployment.md` — documentation updates

## Issue linkage

Closes #499

## Testing / verification

Comprehensive test suite in `tests/test_refresh_durability.py` covers:
- Write-through mirror: queued → running → heartbeat → terminal transitions, crash case
- `decide_resume`: retry budget, pinned backtests, deleted/departed flights, forecast horizon, refresh gate checks
- Boot-time reconciliation: resume vs abandon decisions, error handling, already-terminal rows
- Orphan snapshot and max-attempts configuration

All tests pass; the registry defaults to non-durable mode so existing tests are unaffected.

https://claude.ai/code/session_01TcQxEKw62bAhfSbuX5SaTa

### DIFF
--- a/alembic/versions/082_briefing_refresh_jobs.py
+++ b/alembic/versions/082_briefing_refresh_jobs.py
@@ -1,0 +1,77 @@
+"""Durable briefing-refresh job tracking (issue #499).
+
+Refresh state used to live only in the in-memory ``_RefreshRegistry``. When the
+container died mid-refresh (OOM, deploy, crash) nothing recorded that a refresh
+had ever been in flight: the pack row is written at the *end*, and the pack
+directory created up front is an orphan with no row pointing at it. This table
+is the write-through mirror of the registry, so an interrupted refresh leaves a
+durable record that boot-time reconciliation can resume — and that a
+post-mortem can read.
+
+``flight_id`` is deliberately **not** a foreign key: the row must outlive the
+flight so reconciliation can distinguish "flight was deleted, don't resume"
+from "no record at all". ``user_id`` does cascade so account deletion takes the
+job history with it (the explicit sweep in ``app._on_delete_user`` covers the
+bulk-delete path, which doesn't emit ORM cascades).
+
+``create_table`` / ``create_index`` work on SQLite (dev) and MySQL (prod)
+without batch mode.
+
+Revision ID: 082
+Revises: 081
+Create Date: 2026-07-26
+"""
+from __future__ import annotations
+
+from typing import Sequence, Union
+
+import sqlalchemy as sa
+from alembic import op
+
+revision: str = "082"
+down_revision: Union[str, None] = "081"
+branch_labels: Union[str, Sequence[str], None] = None
+depends_on: Union[str, Sequence[str], None] = None
+
+
+def upgrade() -> None:
+    op.create_table(
+        "briefing_refresh_jobs",
+        sa.Column("id", sa.Integer, primary_key=True, autoincrement=True),
+        sa.Column("flight_id", sa.String(256), nullable=False),
+        sa.Column(
+            "user_id",
+            sa.String(64),
+            sa.ForeignKey("users.id", ondelete="CASCADE"),
+            nullable=True,
+        ),
+        # "user" | "scheduler" | "resume" — the registry's queue-cap class.
+        sa.Column("triggered_by", sa.String(16), nullable=False, server_default="user"),
+        # Client-declared attribution ("user" | "siri" | "mcp"), carried so a
+        # resumed refresh keeps the original surface in usage accounting.
+        sa.Column("source", sa.String(16), nullable=True),
+        sa.Column("as_of_date", sa.Date, nullable=True),
+        # queued | running | succeeded | failed | abandoned
+        sa.Column("status", sa.String(16), nullable=False, server_default="queued"),
+        sa.Column("attempt", sa.Integer, nullable=False, server_default="1"),
+        sa.Column("created_at", sa.DateTime(timezone=True), nullable=False),
+        sa.Column("started_at", sa.DateTime(timezone=True), nullable=True),
+        sa.Column("finished_at", sa.DateTime(timezone=True), nullable=True),
+        sa.Column("heartbeat_at", sa.DateTime(timezone=True), nullable=True),
+        # Last pipeline stage seen — the "where did it die" half of a post-mortem.
+        sa.Column("stage", sa.String(64), nullable=True),
+        sa.Column("pack_path", sa.String(512), nullable=True),
+        sa.Column("last_error", sa.Text, nullable=True),
+    )
+    # Boot-time reconciliation scans by status (non-terminal rows only).
+    op.create_index("ix_refresh_jobs_status", "briefing_refresh_jobs", ["status"])
+    # /refresh/status falls back to the newest row for a flight.
+    op.create_index(
+        "ix_refresh_jobs_flight", "briefing_refresh_jobs", ["flight_id", "created_at"]
+    )
+
+
+def downgrade() -> None:
+    op.drop_index("ix_refresh_jobs_flight", table_name="briefing_refresh_jobs")
+    op.drop_index("ix_refresh_jobs_status", table_name="briefing_refresh_jobs")
+    op.drop_table("briefing_refresh_jobs")
--- a/alembic/versions/082_briefing_refresh_jobs.py
+++ b/alembic/versions/082_briefing_refresh_jobs.py
@@ -51,7 +51,7 @@ def upgrade() -> None:
         # resumed refresh keeps the original surface in usage accounting.
         sa.Column("source", sa.String(16), nullable=True),
         sa.Column("as_of_date", sa.Date, nullable=True),
-        # queued | running | succeeded | failed | abandoned
+        # queued | running | succeeded | skipped | failed | abandoned
         sa.Column("status", sa.String(16), nullable=False, server_default="queued"),
         sa.Column("attempt", sa.Integer, nullable=False, server_default="1"),
         sa.Column("created_at", sa.DateTime(timezone=True), nullable=False),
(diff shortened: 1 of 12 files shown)


roznet:
Rebased onto current `main` (3 commits, no overlap — they touched `web/*`, `api/admin.py`, `api/donations.py`, `fetch/grib/precache.py`; 082 is still the only new revision) and pushed fixes for all three review findings in `778584c`.

**1. Gate-skip recorded as `succeeded` (Important) — fixed.** Confirmed: `_auto_refresh_one` returned `None` both when it ran the pipeline and when it short-circuited on its own refresh gate or a missing `AIRPORTS_DB`, and both callers marked `succeeded` after the await. It now returns a `bool`, and the scheduler cycle and resume pass close the row `succeeded` or `skipped` accordingly. `skipped` joins `TERMINAL`; the `/refresh/status` fallback already treats any non-`abandoned` terminal status as inactive, so it needed no change. Volume is bounded at ~1 row/flight/day (`last_auto_refresh_at` is bumped either way) — noted in the design doc, which sharpens the existing pruning follow-on.

**2. Missing `user_id` index (Important) — fixed.** Added as a named `Index("ix_refresh_jobs_user", ...)` in `__table_args__` rather than `index=True`, so dev's `create_all` and migration 082 agree on the index name; verified both produce the same three indexes. Migration 082 is unmerged, so this went into it directly rather than a follow-up 083.

**3. `reconcile_one` claiming a resume that may not happen (Minor) — fixed.** `try_register` now runs *before* the interrupted row is closed. The row closes `failed` (resumed as attempt N) or `abandoned` (*resume skipped, flight already active*). The trade-off is a narrow window where both rows are non-terminal — a crash inside it means the next boot resumes the old row and abandons the new one at the attempt cap, i.e. one extra run rather than a lost one, which is the right way round for this failure.

**Verification:** full suite green — 4220 passed, 12 skipped. Two new tests cover the fixes (`test_a_gate_skip_is_not_recorded_as_a_briefing`, `test_a_flight_already_refreshing_is_not_claimed_as_resumed`); the existing `ran` fixture was corrected to return `True`, since an implicit `None` now correctly reads as a skip. Migration re-applied cleanly on SQLite and diffed against the ORM. One unrelated flake, `test_grib_pool.py::test_two_concurrent_submits_run_in_parallel` (timing-sensitive, passes in isolation, untouched area), and the pre-existing `test_auth.py::test_google_login_redirects` failure that also fails on a clean checkout.

Nothing deferred from this review.

---

PR #505, opened 2026-07-26
fix(meteo): correct three time/datum errors from the meteorology audit (#481, #486, #487)

### LINKED ISSUE #481: GFS averaged cloud cover is labelled at its window endpoint, giving a 0.5-3 h lag

### LINKED ISSUE #486: ICON geopotential-height reconstruction anchors on ISA rather than model surface

### LINKED ISSUE #487: Freezing-level fallback chain mixes AGL (ECMWF deg0l) and MSL datums

roznet:
## What & why

Three independent findings from the meteorology audit. All three are systematic
biases rather than edge cases, and all three are fixed at the source rather than
at the consumer.

### #481 — GFS averaged cloud cover was labelled at its window endpoint

NCEP publishes LCDC/MCDC/HCDC (and their paired PRES/TMP geometry) as a
time-average over a window that resets every 6 h and grows to the next reset.
Decision §3 already established that such a value belongs at its window
**midpoint** — but the implementation only rewrote *gap* hours, and GFS anchors
are hourly through f120, so `_interp_gfs_diag_hourly` short-circuited and was a
no-op for essentially every briefing. Every native step carried its window mean
labelled at the window's END: a sawtooth lag of half the window width (0.5 h at
f001/f007, 3.0 h at f006/f012).

The averaged half — low/mid/high plus `NWP_CLOUD_DIAG_AVERAGED_SCALARS` — is now
resampled from midpoint knots at **every** hour. Two supporting changes:

- The averaged bracket is located in midpoint space independently of the step
  bracket. With uneven window widths they genuinely disagree (a gap hour at f122
  sits between the f123 and f126 midpoints, not f120's and f123's), which the
  old shared-bracket form could not express.
- At a native step the sub-5 % rule now drops **geometry only**, keeping the
  cover number. A published 0 % ("model says clear") must not degrade into None
  ("unknown") — `_has_native_cloud_content` and the DD-vs-NWP comparisons read
  that difference. A band whose cover is missing on either knot is left entirely
  alone, so a decode gap at one hour can't delete a decoded band at another.

Instantaneous and rate fields are untouched at native steps (they're published
*at* the step), and the last anchor has no upper knot so it keeps exactly what
NCEP filed.

The issue asked for a decision, not just a patch. This implements its **option 1**
(relocate to the midpoint) because §3 already recorded it as the correct
treatment for averaged data and explicitly rejected step-time anchoring; the
missing native-step case is an incomplete implementation of that decision rather
than a new one. The cost §3 named — a native step no longer reports any single
published NCEP number — is stated in the doc, along with why **de-averaging** the
nested windows was rejected despite being mathematically exact: the paired
geometry has no de-averaged form, so cover and geometry would stop sharing one
statistical process (the coupling `_PREFER_AVERAGED_PAIRS` exists to preserve).

### #486 — ICON geopotential reconstruction anchored on ISA

Agreed with the issue's corrected framing: this is **not** a terrain error. The
anchor level's pressure already encodes terrain, so the anchor lands on terrain
automatically. The real error is that pressure surface's deviation from ISA
(MSLP + column mean temperature), order 100–300 m, and it is a uniform column
offset.

The column now anchors on a real geopotential height taken from the hourly's
**outgoing** Open-Meteo levels — same model, point and valid hour — read before
the GRIB replacement overwrites them. Only the surface-most shared level is used:
the reference sets *where* the column sits, never *how thick* its layers are.
Seeding every match would let a different run of the same model dictate the
thicknesses and show up as kinks. Levels below the datum are now integrated
downward instead of staying stranded on the old anchor.

The layer mean also switches from dry-air to **virtual** temperature, fixing the
~30–50 m thickness distortion — the one the issue rightly calls the more
interesting of the two, since it isn't a pure offset.

Fetching ICON `ps` + orography (or `HHL`) is the cleaner end-state and is noted
as such, but it needs new DWD fetches, cache keys and decode paths for an error
that is mostly a uniform offset. The Open-Meteo reference gets the datum right
with data already in hand. ECMWF is unaffected (it delivers `gh` on every level,
so the all-present guard makes this a no-op).

### #487 — the freezing level mixed AGL and MSL

Agreed, and the leak is wider than the gate that surfaced it. The issue's
option 2 (drop the diagnostics candidate from `_explicit_freezing_level_ft`)
would have fixed nothing: `attach_nwp_diagnostics` mirrors
`diag.freezing_level_ft` onto `HourlyForecast.freezing_level_m`, so the AGL value
reaches candidate **two** as well — candidates 2 and 3 are the same number for
ECMWF. Downstream of that same mirror: `dd_nwp_agreement` (MSL sounding vs AGL
NWP), `comparison.py` model divergence (ECMWF-AGL vs GFS-MSL), and the digest.
Ironically the named gate bites least — the explicit track is ICON-D2, which
publishes no freezing level at all.

So this answers the issue's question 1: convert at the decode boundary, which
fixes every consumer at once and makes the invariant statable — `freezing_level_ft`
and `freezing_level_m` are MSL, always. Terrain comes from the **model's own**
orography, `z(surface_pressure)` read off the ECMWF geopotential column, not
SRTM: an AGL field references model terrain, which at ~9 km differs from real
terrain by hundreds of metres in mountains. Both inputs are already decoded and
the a2 merge runs before a1 at each step, so nothing new is fetched; terrain is
time-invariant so it's computed once per point per run.

Where terrain is unavailable the value is **dropped, not passed through** —
`attach_nwp_diagnostics` only mirrors a non-None value, so Open-Meteo's
correctly-referenced number stays in place. A wrong-datum ECMWF value is worse
than a right-datum Open-Meteo one.

`ceil`/`cbh`/`hcct` deliberately stay AGL — ceiling and cloud base are
conventionally AGL in aviation, so their datum is a real question with a
different answer (the #441 finding #3 follow-up), not the same bug.

## Issue linkage

Closes #481
Closes #486
Closes #487

## Testing / verification

Unit tests added for each finding:

- `tests/test_grib_fill.py::TestGfsNativeStepRelabelling` — native steps are
  realigned; steady state is unperturbed; instantaneous fields survive at native
  steps; a missing neighbour cover keeps the published band; no realignment
  without `gfs_init`. Two existing tests updated to the new intended behaviour
  (the pt11 12z native step now reports 33.3 % rather than asserting a 06–12z
  mean at 12z; the out-of-range clamp now takes cover and geometry from the same
  published step).
- `tests/test_ecmwf_sounding.py` — reference anchor produces an exactly uniform
  offset; the datum may sit above the surface-most level; a second reference
  level cannot bend the profile; a delivered height always wins; a moist column
  integrates thicker than a dry one.
- `tests/test_grib.py` — `deg0l` converted to MSL, dropped without terrain,
  sentinel not converted. `tests/test_ecmwf_sounding.py::TestModelSurfaceHeight`
  covers interpolation, below-lowest-level extension, and the None paths.

Full-suite verification is limited by this environment: `euro_aip` is not on
PyPI and `cfgrib`/`eccodes` are absent, so 87 of 190 test modules can't be
collected here. Across every module that does collect, the failure set is
**byte-identical with and without this change** (93 pre-existing failures, all
`ModuleNotFoundError` or an xarray-version artifact in the untouched
Greenwich-seam spatial code), and all new tests pass. The GRIB-decode paths
themselves have not been exercised against real GRIB files in this environment.

`designs/meteorology-decisions.md`: §3 extended with the #481 amendment and the
rejected de-averaging alternative; §20 and §21 added; INDEX.md updated.

---
_Generated by [Claude Code](https://claude.ai/code/session_01AgZfKogoemSWe7D3HDyS8r)_

### DIFF
--- a/designs/INDEX.md
+++ b/designs/INDEX.md
@@ -46,7 +46,7 @@ Key exports: `compute_wind_components`, `analyze_sounding`, `compute_altitude_ad
 → Full doc: analysis.md
 
 ### meteorology-decisions
-Dated log of explicit meteorological design decisions, each with context, reasoning, and rejected alternatives: ceiling DD-vs-NWP-adjusted, Ogimet icing-zone width / convective contribution at moderate CAPE, GFS cloud-diagnostics window-midpoint interp + RH/condensate gate, convective realizable-CAPE / regime discrimination / DD-stays-pure, NWP-cover vs CAPE-driven risk, ICON-D2 explicit-convection firing table (reflectivity × corroborators, §19). Read before changing any weather calibration or threshold — the choice was likely made deliberately.
+Dated log of explicit meteorological design decisions, each with context, reasoning, and rejected alternatives: ceiling DD-vs-NWP-adjusted, Ogimet icing-zone width / convective contribution at moderate CAPE, GFS cloud-diagnostics window-midpoint interp + RH/condensate gate, convective realizable-CAPE / regime discrimination / DD-stays-pure, NWP-cover vs CAPE-driven risk, ICON-D2 explicit-convection firing table (reflectivity × corroborators, §19), reconstructed-geopotential datum + virtual-temperature thicknesses (§20), freezing level is MSL everywhere with ECMWF `deg0l` normalized at decode (§21). Read before changing any weather calibration or threshold — the choice was likely made deliberately.
 → Full doc: meteorology-decisions.md
 
 ### analysis-metrics
--- a/designs/meteorology-decisions.md
+++ b/designs/meteorology-decisions.md
@@ -320,6 +320,40 @@ Each averaged native step's anchor sits at `step - window_length/2`. For
 the pt11 case f135's midpoint is 13:30 Z, so 14:00 Z sits past it and
 interpolates toward the next anchor's 0 % rather than holding f132's 100 %.
 
+> **Extended 2026-07-26 (#481).** As first written, A only rewrote *gap*
+> hours. Inside f120 GFS publishes hourly, so there are no gap hours and A
+> was a complete no-op there — every native step still carried its window
+> mean labelled at the window's END, a lag of half the window width that
+> sawtooths through each 6-hour cycle (0.5 h at f001/f007, 3.0 h at
+> f006/f012). Since #481 the averaged half — low/mid/high plus
+> `NWP_CLOUD_DIAG_AVERAGED_SCALARS` — is resampled from the midpoint knots
+> at **every** hour, native steps included. Two consequences worth stating:
+>
+> - The reported value at a native step is no longer any single published
+>   NCEP number. That is the unavoidable price of A's own premise: a
+>   time-mean cannot simultaneously be correct in value and correct in time.
+>   The **last** anchor is the exception — with no upper knot the resample
+>   clamps to it, so it keeps exactly what NCEP filed.
+> - The averaged bracket is now located in midpoint space independently of
+>   the step bracket. With uneven window widths the two disagree (a gap hour
+>   at f122 sits between the f123 and f126 midpoints, not between f120's and
+>   f123's), which the shared-bracket version could not express.
+>
+> Instantaneous and rate fields are untouched at native steps — they are
+> published *at* the step and remain exactly as decoded.
+>
+> **Rejected: de-averaging the nested windows.** Within a 6-hour cycle the
+> windows share a start (`0-1`, `0-2`, … `0-6`), so
+> `mean(n-1, n] = n·A_n − (n−1)·A_{n−1}` recovers exact disjoint hourly
+> means — strictly better timing than interpolating between midpoints. It
+> was rejected because the paired PRES/TMP geometry has no de-averaged form,
+> so cover and geometry would stop sharing one statistical process — the
+> coupling this whole section is built on (`_PREFER_AVERAGED_PAIRS`) — and
+> because differencing two nested means amplifies noise into out-of-range
+> covers needing clamping. Revisit only together with the "re-derive cover
+> from RH + condensate" end-state below, which removes the averaged form
+> entirely.
+
 **B. Geometry hold-over with sub-5% drop.** `base_ft`, `top_ft`, and
 `top_temp_c` are not numerically interpolatable — interpolating altitudes
 between two dissimilar layer geometries would create phantom intermediate
@@ -327,6 +361,16 @@ layers. The gap-hour layer reuses the geometry of whichever bracketing
 endpoint has the higher cover; when interpolated cover falls below
 `_GFS_LAYER_DROP_THRESHOLD_PCT` (5 %) the entire `NWPCloudLayerDiag` is
 emptied. Visually a dissipating deck thins toward zero and then drops out.
+
+> **#481 refinement at native steps.** At a *native* step the sub-5 % rule
+> drops the geometry but keeps the cover number, because there the
+> alternative is erasing a value the model actually published: 0 % ("clear")
+> must not degrade into `None` ("unknown"), which `_has_native_cloud_content`
+> and the DD-vs-NWP comparisons read as a real distinction. Gap hours keep
+> the original empty-the-band behaviour — they have no published value to
+> protect. A band whose cover is missing on either knot is left entirely
+> alone at a native step, so a decode gap at one hour cannot delete a
+> decoded band at another.
 
 **C. RH/condensate gate** (`apply_gfs_rh_condensate_gate` in `fill.py`).
 After interpolation, for each GFS hourly with both pressure_levels and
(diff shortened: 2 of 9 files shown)


roznet:
Thanks — the Important finding is correct and is now fixed in 35200e4. All three findings addressed.

**Important — negative `deg0l` dropped: agreed, fixed.**

Verified the evidence before acting: `tests/test_ecmwf_sample.py` does bound `freezing_level_m` at `(-500, 6000)` while `ceiling_m` / `cloud_base_m` / `convective_cloud_top_m` are all `(0, 25000)` in the same table. That asymmetry is this repo's own real-GRIB observation that negative `deg0l` occurs, and the physical reading is right: the 0 °C isotherm below the model surface means the near-surface column is sub-freezing — the cold-airmass case where the freezing level matters most for icing. With terrain in hand it converts to an ordinary MSL height (0 °C 300 m below an 800 m ridge = 500 m MSL).

Also agreed on the framing: not a regression (the pre-#487 code rejected negatives too via the shared `_m_to_ft` closure), but a gap, since #487 is precisely what put the terrain data in hand to handle it. §21 now records the case and the fact that the first cut missed it.

**Minor — duplicated range check: agreed, done in the same push.** Rather than leave two inline guards that would now legitimately differ, the sentinel check plus a per-field negative floor moved into a small `_agl_m` helper, with `_m_to_ft` built on top. The divergence — "a cloud below ground is meaningless" vs "a freezing level below ground is real" — is now stated once, in one place, with the reasoning attached. New test asserts negative `ceil`/`cbh`/`hcct` are *still* dropped, so the latitude stays `deg0l`'s alone.

**Minor — untested wiring: agreed, done in the same push.** Both invariants you named are load-bearing and were comment-only:

- `_geopotential_reference` must read the hourly's levels *before* the GRIB replacement overwrites them. There's now an end-to-end test that feeds a geopotential-free GRIB column (the ICON case) into `_replace_pressure_levels_from_grib` against an hourly whose Open-Meteo 1000 hPa height is deliberately far from ISA, and asserts the result anchors on the reference rather than the ISA fallback. That fails if the read ever moves after the assignment.
- `_fill_ecmwf_orography` — tests for the computed value, the cross-step memoization, and the unresolvable-point path (uncovered, and covered-but-no-`sp`) staying `None` so the freezing level is dropped rather than mislabelled.

On the a2-before-a1 ordering specifically: I left the behaviour as-is deliberately. If a1 ever arrives for a `step_hours` without a matching a2, the orography would be computed from the Open-Meteo geopotential column instead of ECMWF's own — but for ECMWF sections that Open-Meteo data is itself ECMWF-sourced, so both are real MSL heights for the same model and the difference is tens of metres feeding a terrain term. Worth knowing, not worth a guard that would have to detect "did these levels come from GRIB?" — I'd rather not add that inference. Flagging it here so it's on the record.

Verification: full-suite runs remain limited in this container (`euro_aip` absent, `cfgrib`/`eccodes` absent — 87 of 190 modules don't collect). Across everything that does collect, the failure set is byte-identical to the pre-change baseline, and all new tests pass.

One thing this round did **not** change, and worth a human eye before merge: #481 shifts cloud-cover values on every GFS briefing, and I have no way to quantify that against a real pack here.

---

roznet:
**#481 reworked before merge — the knots now come from disjoint windows, not the nested published ones.**

The review approved the first cut of #481 as "working as designed", and it was — the arithmetic followed the algorithm faithfully. The algorithm was wrong.

**Broken premise.** Anchoring each published window mean at its own midpoint is correct for *independent* samples. NCEP's windows inside a cycle are **nested** — they share the reset and grow (`0-4`, `0-5`, `0-6`) — so consecutive means are strongly correlated and their midpoints are not evenly spaced: f001..f006 sit at 0.5–3.0 h, then the reset jumps the next knot to 6.5 h. Hours 4, 5 and 6 fall in that 3.5 h hole and were interpolated across it, which discards the very windows covering them and drags them toward the *next* cycle's first value. Half of every 6-hour cycle.

**Consequence.** Not just blurring — inversion. Deck arriving at the reset, published `0 %` for f001–f006 then `100 %` from f007:

| step | model | pre-#481 | first cut | now |
|---|---|---|---|---|
| f004 | 0 % | 0 % | **28.6 %** + invented base | 0 % |
| f005 | 0 % | 0 % | **57.1 %** + invented base | 0 % |
| f006 | 0 % | 0 % | **85.7 %** + invented base | 50 % (the discontinuity) |

A published mean of 0 % over 0–6 h *proves* cover is 0 across that span — it cannot go negative to compensate. So this was three hours of phantom overcast, and strictly worse than the behaviour it replaced.

**Fix.** Difference the nested means first, then midpoint-anchor:

```
mean(0-5)·5 and mean(0-6)·6  →  mean over (5, 6]  at midpoint 5.5
```

Evenly spaced knots at 0.5, 1.5, 2.5 …, no hole. It also sharpens what #481 was after: `mean(0-6)` lags its label by 3 h, `mean(5,6]` by 0.5 h.

**Measured against fields whose truth is known exactly:**

| case | truth at f004–f006 | pre-#481 | first cut | now |
|---|---|---|---|---|
| smooth ramp (−4 %/h) | 84 / 80 / 76 | 92 / 90 / 88 (+2…+12 sawtooth) | exact | exact |
| mid-cycle burn-off | 0 / 0 / 0 | 87.5 / 70.0 / 58.4 | 41.7 / 25.0 / 8.3 | 25.0 / 0.1 / 0.1 |
| deck at reset | 0 / 0 / 0 | 0 / 0 / 0 | 28.6 / 57.1 / 85.7 | 0 / 0 / 50 |

**On the earlier rejection of de-averaging.** The stated reason was that paired PRES/TMP geometry has no de-averaged form, so cover and geometry would stop sharing one statistical process. Only cover is differenced here; geometry is carried from the published anchor unchanged — and per decision B it is never numerically interpolated anyway (`_interp_layer` holds it from the higher-cover endpoint, now the higher *disjoint* cover). So `_PREFER_AVERAGED_PAIRS` is preserved, not broken. The noise objection is real and is handled by clamping knots to [0, 100]; a knot with an incomplete input step is `None`, and the anchor falls back to what NCEP published.

**Tests.** New: reset-boundary regression guard, mid-cycle transition recovery, clamping, geometry preservation, missing-neighbour fallback. One existing test carried covers that cannot arise from real GRIB (`mean(0-1)=80` with `mean(0-3)=10` implies −25 % over hours 1–3); its data is corrected to a realizable pair, keeping what it actually tested.

Full suite: **4262 passed, 1 pre-existing failure** (`test_google_login_redirects`, unrelated and failing on `main` too). This environment collects everything, so the coverage gap noted earlier in the PR does not apply here.

§3 updated — the de-averaging rejection is replaced with the adopted decision and the reasoning above.

Issue #457, opened 2026-07-20
HRRR short-range upgrade: full-sounding replacement for the GFS slot when route fits CONUS + horizon (Lambert-grid decode)


roznet:
## Goal

When a route fits entirely inside the HRRR CONUS domain **and** the flight window is within the HRRR forecast horizon, source the GFS model slot's GRIB enrichment from **HRRR** (3 km, convection-allowing, radar-assimilating) instead of GFS — and, unlike today's GFS patch-onto-Open-Meteo style, do a **full sounding replacement** like ICON/ECMWF. Otherwise, GFS exactly as today.

Same product decision as the ICON-D2 upgrade (companion issue): **in-place upgrade of the existing `gfs` model slot**, all-or-nothing gate, no mixing of HRRR and GFS data within one briefing. One honest difference from the D2 case: ICON-D2 is the same model family nested in ICON-EU, while HRRR (WRF-ARW regional) is a genuinely different model from GFS (global FV3) — so the upgrade must be **visibly badged** (e.g. `GFS (HRRR)` in the freshness bar), and day-over-day assessment jumps at the horizon boundary are expected and acceptable.

Related context: the US expansion plan currently scopes US routes to GFS + ECMWF; this issue is the short-range quality upgrade for that region. Note also that Open-Meteo's `/v1/gfs` default (`gfs_seamless`) may already blend HRRR into the *surface* fields over CONUS at short range — verify; if true, HRRR GRIB replacement makes the sounding **consistent** with the surface data rather than introducing a new inconsistency.

## Gate

- All route points inside the HRRR CONUS domain AND `flight_window_end ≤ run_init + horizon` for the selected run.
- **Only 00/06/12/18z cycles extend to 48 h**; other hourly cycles stop at 18 h. Run selection: prefer the freshest cycle whose horizon covers the flight window (this usually means an extended cycle for anything beyond a few hours out). If no covering run → plain GFS.
- Domain test: HRRR's grid is Lambert-projected, so "inside CONUS" is not a simple lat/lon rectangle. Practical approach: transform route points into grid x/y with the same pyproj projection used for interpolation (below) and check they land within grid bounds — exact and free once the projection exists.

## What's verified reusable (checked against the live bucket, 2026-07-18 00z run)

[...]

PR #508, opened 2026-07-27
feat(fetch): HRRR full-sounding upgrade of the gfs slot on CONUS routes (#457)

roznet:
When a route fits entirely inside the HRRR CONUS domain and the flight
window is within the selected run's horizon, source the gfs model slot
from HRRR (3 km, convection-allowing) as a FULL sounding replacement —
the ICON/ECMWF pattern, not the GFS patch. Otherwise plain GFS, exactly
as today. Same in-place-upgrade product decision as the ICON-D2 icon
slot (#456): all-or-nothing gate, never a mixed HRRR/GFS briefing, and
visibly badged "GFS (HRRR)" since HRRR (WRF-ARW) is a genuinely
different model from GFS (FV3).

- hrrr_fetch.py: 2-digit-fhour S3 URLs (noaa-hrrr-bdp-pds), run
  selection across hourly cycles with per-cycle horizons (00/06/12/18z
  reach 48h, others 18h; probe the last-needed fhour — publication is
  progressive), Lambert-projection domain gate sharing the decode's
  pyproj math, idx planning for the sounding + diagnostics sets.
- decode.py: Lambert-grid branch in the shared interpolation helpers
  (no lat/lon dims but (y, x) dims + projection attrs → project targets,
  same vectorised bilinear on the projected axes). HRRR decoders with
  grid-relative → earth-relative wind rotation (uvRelativeToGrid=1),
  CIMIXR-decodes-as-"unknown" mapping, direct-DPT support in
  _convert_raw_sounding, and an ECMWF-shaped diagnostics builder (band
  covers + ceiling + overall base; ML CAPE/CIN, CIN already negative).
- Enrichment: _enrich_gfs returns (ts, source_key); the HRRR path
  rebuilds pressure_levels via the ECMWF replacement flow per hour and
  falls back to plain GFS whole if nothing merges. When HRRR sourced
  the slot, the GFS averaged-window machinery (window-midpoint interp,
  RH/condensate gate) is disabled — every HRRR field is instantaneous.
- Wiring: hrrr cache dir (6h TTL), hrrr:noaa freshness source tracking
  extended cycles only (hourly tracking would auto-refresh short-lead
  US packs every hour), pack model_sources via grib_sources, and the
  GFS (HRRR) badge in the freshness bar + popover.
- gfs_idx.py parse/plan helpers parametrized on variable sets (defaults
  keep GFS behaviour); pyproj promoted from transitive to declared dep.
- Tests: unit coverage for URLs/gate/run-selection/idx-planning/Lambert
  math/rotation/builders/gate-fallback/fill-semantics/registry, plus
  golden decode tests against a real byte-ranged wrfprs sample
  (tests/data/hrrr_samples/, gitignored, skip-if-absent) verifying the
  projected-grid interpolation sub-kelvin against a nearest-neighbour
  readback and the wind rotation against the cone-constant closed form.

Co-Authored-By: Claude Fable 5 <noreply@anthropic.com>
Claude-Session: https://claude.ai/code/session_014TfdLuNVvYUVKSGTc4sSBs


roznet:
## Code Review

Reviewed the full HRRR CONUS upgrade of the `gfs` slot (#457) — 19 files: `hrrr_fetch.py`, the Lambert-grid branch in `decode.py`, the `_enrich_gfs` → `(ts, source_key)` rework, freshness/cache/deps wiring, the web badge, and both test files. The design-doc updates (`fetch.md`, `weather-engine-specs.md`) are unusually thorough and match the code. The wind-rotation math checks out (`u_e = u_g·ny − v_g·nx`, `v_e = u_g·nx + v_g·ny` is the correct projection onto grid-space east/north), the `_enrich_gfs` signature change updated every caller, and the `xarray.interp` on coord-less `(y, x)` dims does interpolate on implicit integer indices as the comment claims (verified against xarray 2026.1.0). Findings below; one is a real data-coverage bug that the tests just miss.

### Critical

- `src/weatherbrief/fetch/grib/hrrr_fetch.py:375` — `fhours.add(round(delta))` uses Python's banker's rounding, so a departure at exactly `HH:30` collapses consecutive half-hour deltas onto the same *even* forecast hour and silently drops every other hour. Verified against the module on this branch:

  ```
  dep 15:00 +2.0h -> [3, 4, 5]
  dep 15:15 +2.0h -> [3, 4, 5, 6]
  dep 15:30 +2.0h -> [3, 4, 6]     <-- f05 missing
  dep 15:45 +2.0h -> [3, 4, 5, 6, 7]
  ```

  `round(0.5)=0, round(1.5)=2, round(2.5)=2, round(3.5)=4` — so for a `:30` departure the set degrades to 2-hourly, and the skipped hour keeps its Open-Meteo 28-level sounding instead of a native HRRR one. `:30` is one of the most common GA departure slots, and this defeats the headline benefit (hourly 3 km output) for it. `TestFlightWindow.test_non_round_departure_brackets_window_end` (`tests/test_hrrr.py:218`) uses `15:15`, which is the one non-round minute value that doesn't trip the bug. Use `math.floor(delta + 0.5)` (or floor/ceil bracketing) and add a `:30` case to the test.

### Important

- `src/weatherbrief/fetch/grib/__init__.py:2945` — the "never a mixed HRRR/GFS briefing" invariant isn't enforced on the diagnostics path. `enriched_hours` only increments when `replaced > 0` (pressure levels), but `_apply_cloud_diagnostics_to_sections` at line 2938 has already mutated the sections independently of `replaced`. If the sounding decode yields nothing for every hour while the diag decode succeeds — a corrupt/truncated cached sounding blob with a good diag blob is enough — the function returns `None`, `_try_enrich_gfs_from_hrrr` falls back to plain GFS, and the briefing ends up with HRRR's *instantaneous* cloud diagnostics already merged, badged `gfs:noaa`, with `gfs_init` set and `apply_gfs_rh_condensate_gate` running over them. That is precisely the averaged-window machinery the PR is careful to disable everywhere else. Track "did we mutate anything" (diagnostics included) rather than `replaced > 0`, and either commit to HRRR or don't touch the sections until the hour is known good.

- `src/weatherbrief/fetch/grib/__init__.py:2817` — the comment "Only reachable when NOTHING was merged (per-fhour errors are contained inside the loop)" asserts an invariant the code doesn't structurally guarantee: the per-fhour `finally: _grib_gc()` sits outside the per-fhour `except`, so a raise there escapes the loop *after* mutations, into the caller's blanket `except → return None → plain GFS`. Low probability, but it's the same mixed-state hazard as above and a one-line `merged_any` flag closes both.

- `src/weatherbrief/fetch/grib/hrrr_fetch.py:425-458` — `fetch_hrrr_byte_ranges` holds the whole forecast hour in memory three times over: `results` keeps every downloaded chunk alive, `out` accumulates a full copy, then `bytes(out)` makes a third. At the PR's own ~190 MB/fhour that's a ~570 MB transient before anything reaches disk, ~6× the GFS CLMR/ICMR payload the identical pattern was written for. Production runs a 4 GiB cgroup that has already OOM-killed a user briefing (which is why `WB_GRIB_PRECACHE_ENABLED` is currently off). Streaming each chunk to the cache file in offset order, or at least `del`-ing chunks from `results` as they're extended, would keep the peak near 1×. Same applies to `fetch_hrrr_cloud_diag_ranges:459-486`, though ~10 MB there is harmless.

- `src/weatherbrief/fetch/grib/hrrr_fetch.py:325` — `hrrr_window_out_of_range` is dead code: nothing in `src/` calls it, only `tests/test_hrrr.py` imports it. Its docstring says it exists to "distinguish the expected *flight beyond the 48 h horizon* skip from a genuine probe failure", which is exactly what `grib_skip_reasons` is for — but no HRRR entry is ever recorded there, so an out-of-horizon US flight is indistinguishable from an S3 outage in the pack metadata. Either wire it into the skip-reason path in `_try_enrich_gfs_from_hrrr` (where `run_info is None` currently just logs) or drop it.

### Minor

- `src/weatherbrief/fetch/grib/decode.py:2416` — `ml_cin = min(0.0, float(ml_cin_raw))` fails in the unsafe direction. The sign convention was verified live on a single date (2026-07-26); if HRRR ever emits a positive magnitude the way ECMWF/ICON do, a strong cap (say +400 J/kg) silently becomes `0.0` = "no inhibition" and the convective suppression gate stops suppressing. Given `_normalize_model_cin` already exists to disambiguate this exact thing for other models, consider dropping implausible positives to `None` (unknown) rather than flooring them to zero.

- `tests/test_hrrr_sample.py` — the golden decode tests are the only coverage of the Lambert projected-grid interpolation and the grid→earth wind rotation against real data, and `tests/data/hrrr_samples/` is gitignored, so they skip everywhere except a machine that fetched the sample by hand. The run on this branch is `117 passed, 5 skipped` — those 5 are these. Worth noting in the PR that the two genuinely novel numerical paths are unverified in CI; a small committed fixture (single message, few hundred KB) would cover the projection math without the byte-range download.

- `src/weatherbrief/fetch/grib/hrrr_fetch.py:145-160` — the domain gate uses hard-pinned grid constants while `decode.py` reads the grid from each message's own attributes. The docstring says a grid change "degrades to a decode miss, never to wrong positions", but the failure is quieter than that: constants that drift outward relative to the true grid let a route pass the all-or-nothing gate while `_lambert_grid_weights` drops the out-of-bounds points from `inb_idx`, producing HRRR at some route points and Open-Meteo at others — a spatially mixed slot, with `enriched_hours > 0` so no fallback fires. A cheap guard is to compare the pinned `Nx/Ny/Dx/Dy/LoV` against the first decoded message's attrs once per run and fall back to GFS on mismatch.

- `src/weatherbrief/fetch/grib/hrrr_fetch.py:420` — `plan_hrrr_cloud_diag_byte_ranges` passes `pairs=` but leaves `prefer_averaged` at the GFS default (`_PREFER_AVERAGED_PAIRS`, which contains the same `LCDC/MCDC/HCDC` keys HRRR uses). It's inert today because HRRR publishes no averaged forms, and both docstrings say so — but the safe expression of "this model has no averaged forms" is `prefer_averaged=set()`, which can't quietly start preferring an averaged message if NCEP ever adds one.

### Cross-cutting notes (no action needed)

- **Cross-model robustness**: the Lambert branch is correctly placed at the shared-helper level (`_decode_pressure_vars_from_datasets` / `_interpolate_per_point`) rather than in an HRRR-specific decoder, and it keys off GRIB attributes, so a future RRFS or any other NCEP Lambert product rides it unchanged. Both branches return the same `_GridWeights` contract, and `frac_i`/`frac_j` are consistently (row=y, col=x) across the two paths.
- **Sibling metrics**: the `unknown` → ice-mixing-ratio mapping is the only genuinely risky map entry, and the constraint that makes it safe (this decoder only ever sees our own byte-ranged downloads, where CIMIXR is the single planned variable eccodes can't name) is documented in three places. No sibling decoder shares `_HRRR_FULL_VAR_MAP`, so there's no bleed risk.
- **Web/CSP**: `sourcedModelLabel()` is a clean consolidation of the duplicated D2 tagging in both the summary line and the popover; no new external loads, fetches, or inline scripts, so nothing in `deploy/weather.flyfun.aero.caddy` is implicated.
- **Cross-platform parity**: this PR is web-only on the client side and the badge is derived from `model_sources`, which iOS reads through the same `/api/flights` payload — no hand-copied surface (preset tables, `metrics-catalog.json`, DTOs) is touched, so no `/sync-ios-web` pass is needed.

roznet:
## Detailed review — changes required before merge

I reviewed this implementation at `e33a807e9da4d0df30780ff114f19cd9e4c798b2`, focusing on HRRR GRIB semantics, mapping into the sounding/diagnostic/advisory pipeline, spatial and temporal interpolation, model-boundary behavior, meteorological interpretation, and efficiency.

The Lambert-grid interpolation and wind rotation look sound, but there are **four blocking correctness issues**. Please treat items 1–4 as required and add regression coverage for each.

---

### 1. Use the actual HRRR cloud-liquid-water identifier: `CLWMR`, not `CLMR`

**Location:** `src/weatherbrief/fetch/grib/hrrr_fetch.py`, sounding variable set near line 83  
**Test fixture:** `tests/test_hrrr.py`, index-planning fixture near line 247

The HRRR `wrfprsf` pressure-level inventory publishes cloud liquid water mixing ratio as:

```
CLWMR:<pressure level>:Cloud Mixing Ratio [kg/kg]
```

The implementation requests `CLMR`. The unit tests use the same incorrect spelling, so they validate the implementation assumption but do not match a real HRRR inventory.

Official inventory example:  
https://www.nco.ncep.noaa.gov/pmb/products/hrrr/hrrr.t00z.wrfprsf02.grib2.shtml

With a representative real index entry, the current planner selects `CIMIXR` and `TMP` but omits `CLWMR`. This means `cloud_liquid_water_kg_kg` is absent in production. That is materially significant:

- the default Ogimet-NWP icing calculation cannot apply its liquid/ice glaciation adjustment;
- SFIP can fall back to proxy behavior instead of using native condensate;
- the sounding may appear complete while its cloud microphysics are not.

**Required implementation:**

- Replace `CLMR` with `CLWMR` in the requested HRRR pressure-level variables.
- Update parser/planner tests to use realistic `CLWMR` inventory lines.
- Add a builder-level test proving a real-name `CLWMR` message reaches `PressureLevel.cloud_liquid_water_kg_kg`.
- Add a negative regression assertion that a realistic inventory does not silently omit the condensate field.

---

### 2. Select a contiguous set of native HRRR forecast hours

**Location:** `src/weatherbrief/fetch/grib/hrrr_fetch.py`, `select_hrrr_run_and_fhours`, near lines 368–382

The current implementation builds a small set of offsets and applies `round(delta)`. Python rounding is not appropriate for determining a forecast window: it uses ties-to-even and does not guarantee that every native hourly field intersecting the flight window is selected.

Reproduction:

- run initialization: 12:00Z
- departure: 17:30Z
- duration: 2.5 hours
- current selection: `[5, 6, 8, 10]`
- required covering hours: `[5, 6, 7, 8]`

This skips f07/19Z, then fetches f10/22Z outside the flight. The missing hour is particularly important because pressure levels and diagnostics do not share identical fill behavior: pressure levels may be linearly interpolated across anchors, whereas diagnostics are forward-filled/persisted. Native hourly HRRR cloud/CAPE/CIN information can therefore be replaced with synthetic or stale values even though the source data exists.

**Required implementation:**

Compute an inclusive contiguous range:

```python
first_fhour = floor((departure_time - init_time).total_seconds() / 3600)
last_fhour = ceil((arrival_time - init_time).total_seconds() / 3600)
fhours = range(first_fhour, last_fhour + 1)
```

Then clamp to `[0, horizon]` and retain the existing availability/probe rules. Avoid using midpoint samples or `round()` to define coverage.

**Required tests:**

- half-hour departure and arrival;
- exact-hour boundaries;
- a flight shorter than one hour;
- crossing multiple native hours;
- horizon boundary;
- assertion that the returned hours are always contiguous, ordered, unique, and no wider than the minimal covering range.

---

### 3. Do not send HRRR’s generic cloud cover through the “native convective scheme” path

**Location:** HRRR diagnostic construction in `src/weatherbrief/fetch/grib/decode.py`, near lines 2428–2435, and its interaction with `analysis/sounding/convective.py`

The HRRR diagnostic object currently supplies:

- total, low, middle, and high cloud cover;
- mixed-layer CAPE and CIN;
- ceiling/base information.

It does **not** supply a convective realization channel such as convective cloud cover, convective base/top, convective precipitation, reflectivity, or echo tops.

The default advisory method is `nwp`. Existing native-content detection treats generic total/band cloud cover as evidence that a native convective-cloud scheme is present. That suppresses the CAPE fallback. Because the HRRR diagnostic has no firing/realization channel, the native path can conclude that there is no convection.

A reproduced diagnostic containing:

```
total cloud cover = 40%
ML CAPE = 2000 J/kg
ML CIN = -20 J/kg
```

returns `ConvectiveRisk.NONE` under the default NWP method. That is not a meteorologically safe interpretation. HRRR is a convection-allowing, explicitly convective model; generic cloud fraction is not evidence that a strongly unstable column is convectively quiet.

NOAA HRRR description:  
https://rapidrefresh.noaa.gov/hrrr/

**Required implementation:**

Introduce explicit diagnostic capability/source semantics rather than inferring convective capability from generic cloud fields. For example, the diagnostics can declare whether they contain:

- a native convection-realization channel;
- only thermodynamic potential;
- generic non-convective cloud cover.

For HRRR, either:

1. ingest suitable realization fields such as composite reflectivity, echo tops, or convective/total precipitation and map them deliberately into the advisory; or
2. classify the current HRRR payload as “thermodynamic potential without native convective realization” so the advisory uses the CAPE/CIN/potential path.

Do not globally reinterpret all generic NWP cloud cover without checking existing GFS, ECMWF, and ICON behavior. This should be a source/capability-aware change.

If realization fields are added, spatial aggregation should consider **corridor extrema**, not only interpolation at the route centerline, because a narrow convective cell near the corridor must not be averaged away.

**Required tests:**

- HRRR CAPE 2000/CIN -20 with generic cloud cover must not grade `NONE` solely because no convective-cloud field was supplied.
- Stable/low-CAPE HRRR cases must remain quiet.
- Existing GFS/ECMWF/ICON convection tests must remain unchanged unless an intentional model-specific correction is documented.
- Missing-data behavior must be explicit; absence of a realization channel is not equivalent to a zero realization value.

---

### 4. Make the HRRR upgrade transactional; never expose a mixed HRRR/GFS slot

**Location:** `src/weatherbrief/fetch/grib/__init__.py`, HRRR enrichment and source selection near lines 2776–2780 and 2928–2952

The PR description promises an all-or-nothing replacement, but the implementation treats `enriched_hours > 0` as success. This permits several mixed states:

1. One HRRR forecast hour succeeds while other requested hours remain GFS/Open-Meteo.
2. One decoded HRRR hour contains only a subset of fields but replaces a complete GFS pressure column.
3. A failure at the beginning or end of the flight remains unfilled because linear fill only operates strictly between at least two anchors.
4. HRRR diagnostics are attached before sounding replacement is proven successful. If `replaced == 0` and the code falls back to plain GFS, the mutated HRRR diagnostics remain on the time slots.

The result can therefore be a pack badged/source-keyed as HRRR while containing GFS soundings, or a GFS fallback pack containing HRRR diagnostics. Besides provenance problems, this disables/enables model-specific processing based on an inaccurate source identity.

**Required implementation:**

Stage HRRR enrichment separately from the live `TimeSeriesData`:

1. Decode into a temporary candidate structure or deep copy.
2. Validate the complete requested time window and every required route point.
3. Validate mandatory sounding fields according to an explicit policy—at minimum temperature, humidity/dew point, and U/V wind; also validate native condensate fields when the icing path claims native microphysics.
4. Validate the diagnostic coverage expected by the selected advisory capability.
5. Only after all validation succeeds, atomically commit the pressure levels, diagnostics, and source metadata.
6. On any failure, discard the entire candidate and leave the original GFS data untouched.

Do not use the number of replaced hours as the success contract. Return an explicit complete-success result, ideally including validation details suitable for logging.

If partial HRRR operation is ever desired later, it needs per-field/per-hour provenance and model-aware advisory behavior; it should not be introduced implicitly by this PR.

**Required tests:**

- failure on the first, middle, and last requested hour;
- exactly one successful HRRR hour;
- successful download with a mandatory variable absent;
- diagnostic success with sounding failure;
- sounding success with diagnostic failure, according to the declared completeness policy;
- after every failure, assert deep equality with the original GFS pack and assert the source remains GFS;
- successful complete replacement asserts every requested time/route point is HRRR and source metadata changes atomically.

---

## Performance and operational improvements

These are not as serious as the correctness issues above, but should be addressed or explicitly tracked before enabling this at scale.

### Coalesce byte ranges

The current plan can perform roughly 9 selected messages × 35 pressure levels = **315 S3 range requests per forecast hour**, plus diagnostics. In the HRRR inventory, useful messages at each level occur in adjacent groups, for example:

- `HGT/TMP/RH/DPT`
- `VVEL/UGRD/VGRD`
- `CLWMR/CIMIXR`

Coalescing adjacent byte ranges would reduce this to roughly 105 requests per hour without downloading materially more data. A single bounded range per pressure level could reduce it to about 35 requests/hour at the cost of a few unused neighboring messages.

Please add request-count and byte-count instrumentation so the production cost and latency are visible.

### Bound cache growth

A six-hour TTL can retain several large hourly run directories, especially with concurrent flights. Add or plan a size quota/LRU cleanup policy, not only an age-based purge. Cleanup must avoid deleting files in active use.

---

## Items reviewed and considered sound

These portions appear correct and should be preserved while applying the fixes:

- projection from route latitude/longitude into HRRR Lambert x/y coordinates;
- bilinear interpolation on the projected grid;
- rotation of grid-relative U/V winds into earth-relative components;
- direct dew-point conversion from kelvin;
- geopotential height in gpm mapped to meters/feet;
- vertical velocity retained as Pa/s;
- cloud liquid/ice mixing ratios interpreted as kg/kg;
- cloud-cover diagnostics interpreted as percentages;
- CIN retained with the conventional negative sign;
- the standard/extended HRRR cycle horizon distinction.

The 90–0 mb CAPE/CIN product is a reasonable mixed-layer diagnostic, but documentation should call it the delivered HRRR 90-hPa-layer quantity rather than imply exact algorithmic equivalence with another model’s nominal 100-hPa mixed layer.

---

## Validation expectations

The existing relevant unit suites are green, but the real-HRRR golden tests are skipped when `tests/data/hrrr_samples` is absent. Because the incorrect `CLMR` fixture passed, at least one CI-capable test must use a sanitized real HRRR index excerpt containing `CLWMR`. It does not need to include a large binary GRIB payload.

Before marking this review addressed, please provide:

- regression tests for all four blocking issues;
- targeted HRRR, interpolation, convection, icing, and GRIB tests;
- full Python suite results, with environment-only failures identified separately;
- frontend tests/build if source-badge or diagnostic display code changes;
- a concise description of the atomic replacement completeness policy;
- measured HTTP request count and downloaded bytes for a representative 2–3 hour flight.

**Merge recommendation: request changes until items 1–4 are corrected.**

roznet:
Both reviews addressed in 9a918aa. I verified each claim against the code and the live bucket before acting; here's the disposition.

## Agreed and fixed

**`:30`-departure rounding bug** (opus critical; codex #2) — confirmed. `round()`'s ties-to-even collapsed consecutive half-hour offsets onto even hours (`15:30 → [3, 4, 6]`). Window hours are now the contiguous inclusive `floor(departure)..ceil(end)` range clamped to the horizon — a contiguous range cannot skip an in-window native hour by construction. Regression tests cover `:30`, sub-hour flights, horizon clamping, plus a parametrized property test asserting the result is always contiguous, minimal, and bracketing.

**Mixed-state hazards** (opus important ×2; the concrete parts of codex #4) — confirmed and closed at the hour level:
- Diagnostics now apply **only after that hour's sounding merged** — a corrupt sounding blob with a healthy diag blob can no longer leave HRRR's instantaneous diagnostics under a plain-GFS fallback where the averaged-window machinery would run over them. Once the sounding is in, diag failures are contained so the hour still counts.
- An hour whose decode covers only **some** route points is skipped whole (this also implements the domain-gate drift guard from the opus minor: pinned constants drifting outward now degrade to a skipped hour + warning, never a spatially mixed slot).
- Success accounting increments at sounding-merge time, and the loop is exception-contained — once anything merged, the slot returns the HRRR timestamp; the caller's GFS fallback is only reachable from the mutation-free setup. Tests cover partial-point coverage, diag-with-failed-sounding, and diag-after-successful-sounding.

**Convective misinterpretation** (codex #3) — confirmed, and the most important catch of the round: `_has_native_cloud_content` read HRRR's generic band covers as "native scheme present but quiet", and with no cover/base/top/precip channel the native path graded a CAPE-2000 column `NONE` — a regression versus plain GFS, which always emits convective cover. Fixed per your option 2, source/capability-aware: new `NWPCloudDiagnostics.convective_scheme_absent` meta field, set only by `build_hrrr_cloud_diagnostics`, routes `assess_convective_nwp` to its existing `nwp_cape_fallback` (which also makes `dd_nwp_agreement` skip the now-circular comparison). The field is classified in a new `NWP_CLOUD_DIAG_META_FIELDS` inventory — both interpolation axes carry it through unchanged (never lerped; the interpolation-audit partition test now enforces this). GFS/ECMWF/ICON leave it unset, so their behaviour is bit-for-bit unchanged — verified by the full convective suite plus a control test that a GFS-style quiet scheme (cover 0%) still reads an honest `NONE`. Ingesting REFC/echo tops as a real realization channel — with corridor-extremum reduction like ICON-D2's #462 track, as you note — is the designed follow-up and needs its own calibration work.

**Download memory + request count** (opus important; codex perf) — both fixed together. Adjacent planned messages coalesce into single HTTP requests (zero extra bytes — HGT/TMP/RH/DPT and CLMR/CIMIXR are adjacent in the real file), and chunks stream to the cache file in offset-ordered batches via a new `cache.put_cached_from_chunks` (same atomic tmp+rename+fsync+fadvise contract), so the ~190 MB forecast hour never sits in memory, let alone 3×. Measured on the live 12z f06 `.idx`: **sounding 315 messages → 106 requests, 176.6 MB; diagnostics 8 → 2, 8.4 MB; a 2–3 h flight (~4 fhours) ≈ 432 requests / ~740 MB**. The per-fhour log line now records messages, requests, and MB, so production cost stays visible.

**Smaller items** — all taken: `hrrr_window_out_of_range` is now wired into the run-finder-miss path (expected beyond-horizon skip logs info; a probe failure inside the horizon logs a warning, so an S3 outage is distinguishable); `ml_cin` values > +5 J/kg map to None (unknown) instead of being floored to "no cap"; `prefer_averaged=set()` states explicitly that HRRR has no averaged forms; `hrrr` joined the daily retention pass with an 8 GiB size-cap backstop (TTL stays the primary bound — HRRR isn't precached, so the daily pass is its only enforcement site); the 90-0 mb CAPE/CIN doc wording no longer implies algorithmic equivalence with ECMWF's 100-hPa mixed layer.

## Partially adopted

**CLWMR vs CLMR** (codex #1, "blocking") — the factual claim doesn't hold against the live product: the bucket's `.idx` for `hrrr.20260726/conus/hrrr.t12z.wrfprsf06.grib2` contains **40 `CLMR` entries and zero `CLWMR`** (the NCO inventory page's spelling differs from what the wgrib2-generated idx actually publishes — same quirk as GFS, whose idx also says `CLMR`). So condensate was never being dropped. That said, the underlying worry — a fixture that only validates the implementation's own assumptions — was fair, so: a **verbatim excerpt of the real `.idx` is now committed as a text fixture** (`tests/fixtures/hrrr_idx_excerpt.txt`) with CI tests asserting CLMR+CIMIXR reach the plan at every level, distractors (SPFH/ABSV/RWMR/REFC) don't, and the real adjacency coalesces; and `CLWMR` is planned alongside `CLMR` defensively (the decode map already accepted both), so an upstream rename can't silently drop the field.

**Transactional full-window staging** (codex #4's prescribed implementation) — declined as specified, with the hazards fixed differently (above). Partial-hours-keeps-HRRR is the deliberate, documented product rule this PR inherits from the ICON-D2 slot (#456: "a partial D2 success keeps D2 and lets the normal time/spatial fill cover gaps"), and issue #457 specifies the same. A full deep-copy staging area for 40-level×80-point×N-hour cross-sections would roughly double peak memory in exactly the code path we just slimmed down. The invariant that actually matters — no hour and no route point ever carries a mix of the two sources, and provenance (`hrrr:noaa`) is only recorded when HRRR data is actually in the pack — is now enforced per-hour, which is the granularity at which the data is written. If per-field provenance is ever wanted, agreed that's its own issue.

**Committed binary GRIB fixture** (opus minor) — kept to repo convention: GRIB samples stay gitignored (`tests/data/hrrr_samples/`, like `ecmwf_samples/`), and a single real message is ~1 MB of complex-packed binary. The committed idx excerpt now covers the planner in CI against real data; the projection/rotation math is covered in CI by synthetic-grid tests validated against pyproj round-trips, and the golden decode tests (sub-kelvin vs nearest-neighbour readback, rotation vs cone-constant closed form) run wherever the sample exists — noted as skip-if-absent by design.

## Validation

- `test_hrrr` (79), `test_hrrr_sample` golden (5, with sample present), full GRIB/freshness/ICON/convective suites: **all green** (949 + 1038 across the two runs).
- Full suite: 4318 passed / 63 failed — the 63 reproduce identically on `main` in this environment (the private `euro-aip` package isn't installable here; those tests run against a stub) and are unrelated to this PR. `test_grib_pool::test_two_concurrent_submits_run_in_parallel` is a timing flake that passes in isolation.
- No client-side changes this round; the new pack field is additive-optional.

---

roznet:
## Second detailed review — changes still required

Reviewed at `9a918aacf03b06d10f5e567712625fb61389e2d6`.

I re-reviewed the complete PR after the first fix commit, including the HRRR GRIB interpretation, spatial/time interpolation, sounding replacement, cloud/icing/convective advisory integration, provenance, cache behavior, and performance.

Several earlier issues are now fixed correctly: contiguous hourly forecast selection, Lambert-grid interpolation, grid-relative wind rotation, defensive `CLMR`/`CLWMR` handling, CIN sign protection, diagnostics-after-sounding ordering, range coalescing, and streaming writes.

However, there are **four remaining correctness blockers** and one cache/reliability issue. Please treat items 1–4 as required before merge.

---

### 1. HRRR currently has no usable native cloud envelope, so the default NWP icing method cannot run

**Relevant code:**

- `src/weatherbrief/fetch/grib/decode.py::build_hrrr_cloud_diagnostics`
- `src/weatherbrief/analysis/sounding/clouds.py::build_nwp_cloud_layers`
- `src/weatherbrief/analysis/sounding/advisories.py::_nwp_cloud_cover_at`
- default method resolution in `src/weatherbrief/tasks/advise.py`

The HRRR diagnostic shape contains:

- low/mid/high band cover;
- one overall cloud base, currently attached to `low.base_ft`;
- ceiling;
- no per-band tops;
- no pressure-level cloud fraction.

`build_nwp_cloud_layers()` only recognizes a native cloud envelope when it has either:

1. pressure-level cloud fraction; or
2. diagnostic bands with both `base_ft` and `top_ft`.

HRRR satisfies neither condition, so `build_nwp_cloud_layers()` returns `None` even when the sounding contains substantial CLMR/CIMIXR and the diagnostic says 80% cloud cover.

I reproduced this with an HRRR-shaped diagnostic and an 850-hPa level containing `0.0003 kg/kg` liquid water plus `0.00005 kg/kg` ice: the resulting NWP cloud envelope is still `None`.

Downstream consequences:

- the default `ogimet_nwp` method is marked unavailable;
- full/native SFIP cannot run, because levels with native CLW are gated by `nwp_cloud_layers`;
- NWP cloud grading falls back to DD rather than using the advertised HRRR NWP source;
- `_nwp_cloud_cover_at()` sees the band percentages, enters its “full geometry” branch, finds no complete base/top layer, and returns `0.0` instead of falling back to the ICAO-band percentages. This can put false 0% cover into altitude regimes.

This means the headline HRRR microphysics upgrade is not reaching the default icing advisory.

**Required implementation:**

Create an honest native HRRR cloud envelope. The preferred approach is to derive contiguous cloudy layers from the pressure-level condensate profile (`CLMR + CIMIXR`), using geopotential height for base/top. This remains model-native evidence and avoids fabricating geometry solely from bulk cover.

The implementation needs an explicit, documented condensate threshold and handling for:

- zero condensate versus missing condensate;
- small numerical background values;
- adjacent cloudy levels and vertical gaps;
- liquid-only, ice-only, and mixed-phase layers;
- the surface/bottom and top pressure-level edges.

If a different HRRR field is selected instead, it must provide a native vertical cloud envelope with suitable pressure/height geometry.

Also change `_nwp_cloud_cover_at()` so it only enters the diagnostic-geometry path when at least one layer has complete base and top geometry. Band cover without geometry must use the fixed ICAO-band fallback, matching the more careful logic already present in `icing_common.nwp_cloud_cover_at_altitude`.

**Required regression tests:**

- HRRR condensate profile produces one or more native NWP cloud layers with correct height/pressure bounds.
- Clear HRRR profile produces an empty native envelope, not `None`.
- Missing condensate remains unavailable, not “clear.”
- Default `ogimet_nwp` can grade an HRRR supercooled-liquid case.
- Full SFIP can grade native HRRR CLW inside the native envelope.
- HRRR band cover without geometry falls back to the appropriate ICAO band rather than returning 0%.
- Run an end-to-end `analyze_sounding` → `_resolve_analyses` test using the default engine methods.

---

### 2. The HRRR ML CAPE/CIN fields are downloaded but ignored by the fallback introduced for HRRR

**Location:** `src/weatherbrief/analysis/sounding/convective.py::assess_convective_nwp`, around lines 773–790

The function calculates:

```python
cape = _effective_cape(indices)
cin = indices.cin_surface_jkg
```

before entering the `convective_scheme_absent` branch. The new native fields:

```python
nwp_diagnostics.ml_cape_jkg
nwp_diagnostics.ml_cin_jkg
```

are not used by that path.

Reproduction:

```text
HRRR diagnostic ML CAPE = 2500 J/kg
HRRR diagnostic ML CIN  = -10 J/kg
MetPy indices unavailable
```

Current result:

```text
ConvectiveRisk.NONE
cape=None
cin=None
method=nwp_cape_fallback
```

Therefore the previous “CAPE 2000 must not become NONE” fix only works when the independently computed MetPy CAPE is available. It does not actually map HRRR’s delivered ML CAPE/CIN into the NWP advisory.

It also makes the HRRR NWP track unnecessarily circular with the DD/MetPy track, despite downloading a model-native parcel diagnostic.

**Required implementation:**

On the explicit `convective_scheme_absent=True` capability path, use HRRR’s native ML CAPE/CIN as the primary potential/inhibition pair. Define and document the fallback when either native field is missing—likely native value first, then the corresponding sounding-derived value.

Do not change the native GFS/ECMWF/ICON realization paths. Their current cover/top/precip semantics must remain unchanged.

Recompute severity modifiers consistently with whichever CAPE value actually grades the HRRR assessment.

**Required regression tests:**

- HRRR ML CAPE 2500 / ML CIN -10 does not return `NONE` when MetPy CAPE is absent.
- Native strong ML CIN applies the intended suppression.
- Missing native CAPE falls back explicitly to sounding-derived CAPE.
- Missing native CIN falls back explicitly to sounding-derived CIN.
- Existing GFS/ECMWF/ICON convective tests remain unchanged.
- DD-vs-NWP agreement behavior remains deliberately documented for this non-realization fallback.

---

### 3. Failed byte ranges produce a partial cached sounding that can replace a complete GFS column

**Relevant code:**

- `src/weatherbrief/fetch/grib/hrrr_fetch.py::iter_hrrr_range_chunks`
- `src/weatherbrief/fetch/grib/hrrr_fetch.py::fetch_hrrr_ranges_to_cache`
- `src/weatherbrief/fetch/grib/__init__.py::_enrich_one_hrrr_hour`
- `src/weatherbrief/fetch/grib/decode.py::build_pressure_levels_from_grib`

`iter_hrrr_range_chunks()` silently skips any span whose download returns `None`. `put_cached_from_chunks()` then commits the file whenever at least one chunk was written.

That partial file becomes a valid cache hit for the full sounding key.

The post-decode validation only checks:

```python
all(decoded_points)
```

That proves each point dictionary is non-empty, not that it contains a complete sounding. `build_pressure_levels_from_grib()` accepts any converted field. I reproduced a height-only decoded column producing replacement `PressureLevelData` entries with:

```text
temperature_c=None
wind_speed_kt=None
geopotential_height_m=<present>
```

Those incomplete levels can replace a complete GFS/Open-Meteo sounding and remain cached for six hours.

**Required implementation:**

For the mandatory sounding artifact:

1. Validate the index plan before downloading:
   - expected pressure levels;
   - one of `CLMR`/`CLWMR` per level;
   - all mandatory core variables at every required level.
2. Treat any planned-span download failure as failure of the whole sounding artifact.
3. Make the streaming iterator raise or otherwise signal incompleteness so the temporary file is deleted instead of committed.
4. Validate the decoded result for every route point before replacement.
5. Delete/retry an existing cached artifact that fails validation.

At minimum, each expected level must contain:

- temperature;
- humidity or dew point according to the declared policy;
- both U and V wind components;
- geopotential height;
- the chosen mandatory microphysics fields if the source is advertised as native-microphysics capable.

VVEL may be optional only if that is an explicit policy; it must not become optional accidentally because a request failed.

Diagnostics may use a different completeness policy, but a partial diagnostic cache must not masquerade as a fully downloaded artifact. Record which fields are present or retry the missing ranges.

**Required regression tests:**

- one failed span after successful spans leaves no cache entry;
- a failed middle span cannot commit a truncated artifact;
- height-only, temperature-only, or wind-incomplete point data is rejected;
- one incomplete route point rejects the whole hour;
- invalid pre-existing cache file is removed and re-fetched;
- a complete artifact still streams without holding the full forecast-hour payload in memory.

---

### 4. `enriched_hours > 0` still permits a mixed HRRR/GFS slot

**Location:** `src/weatherbrief/fetch/grib/__init__.py::_enrich_gfs_from_hrrr`, around lines 2896–2923

The latest implementation provides per-hour ordering, but it still establishes pack-level HRRR provenance when any one requested hour succeeds:

```python
if enriched_hours > 0:
    return hrrr_timestamp
```

The normal pressure-level fill does not make this all-HRRR:

- hours before the first HRRR anchor remain GFS;
- hours after the last HRRR anchor remain GFS;
- a single HRRR anchor fills neither side;
- only gaps strictly between at least two HRRR anchors are rebuilt by interpolation.

The existing `test_grib_fill` tests explicitly enforce those edge semantics.

There is also a per-field hybrid after the last HRRR anchor: `_fill_clw_hourly()` can forward-fill HRRR condensate into pressure levels that otherwise remain GFS.

The resulting pack is badged and freshness-tracked as HRRR despite containing:

- HRRR hours;
- GFS sounding hours;
- potentially GFS sounding levels carrying forward-filled HRRR condensate.

That contradicts the PR’s stated “all-or-nothing” and “never a mixed HRRR/GFS briefing” contract.

The ICON-D2 precedent does not make this safe for HRRR: it has the same provenance weakness, while this PR explicitly advertises a full sounding replacement by a genuinely different model.

**Required implementation:**

Use a slot-level completeness contract. The safest rule is that every requested native HRRR hour must pass the mandatory sounding validation before the source changes to HRRR.

This does not require holding the full GRIB grids in memory. Options include:

- stage only the small decoded/converted per-point pressure columns and diagnostics;
- retain snapshots of the outgoing hourly pressure-level lists and roll back on failure;
- download/validate all hour artifacts first, then sequentially decode and commit only after completeness is established.

If internal missing hours are intentionally allowed, the minimum honest contract is:

- first and last covering HRRR anchors are mandatory;
- every remaining GFS hour is overwritten by HRRR interpolation;
- no GFS field survives under HRRR provenance;
- the policy and interpolation/persistence semantics are explicit and tested.

Per-hour mixed-model provenance would be a larger data-model change and should not be introduced implicitly.

**Required regression tests:**

- failure of first requested hour;
- failure of last requested hour;
- one successful hour only;
- failure of a middle hour;
- diagnostic success with sounding failure;
- complete success over every requested hour;
- after failure/fallback, assert the original GFS pack is unchanged;
- after HRRR success, assert no requested hour retains the original GFS pressure levels or GFS-only enrichment behavior;
- assert source metadata changes only after the completeness contract passes.

---

### 5. The six-hour cache TTL deletes extended HRRR runs that run selection still needs

**Location:** `src/weatherbrief/fetch/grib/cache.py::MODEL_TTL_SECONDS`

HRRR run directories are purged by initialization age. With:

```python
"hrrr": 6 * 3600
```

a 12Z extended run is deleted at 20Z even when run selection still chooses it for a flight beyond the short-cycle 18-hour horizon.

I reproduced the purge directly: a cached `20260727_12z` HRRR run evaluated at 20Z is removed.

This happens naturally while the next 18Z extended run is still inside its publication delay or while its last-needed forecast hour is incomplete.

Consequences:

- expensive approximately 740 MB representative-flight caches are repeatedly downloaded;
- the cache provides no reuse during part of every extended-cycle transition;
- concurrent briefings can race, with one caller deleting a run directory another caller is actively writing or decoding.

**Required implementation:**

Make HRRR retention aware of the extended-cycle cadence and publication delay. Options include:

- a TTL comfortably longer than six hours plus publication margin;
- explicitly preserving the currently selected run;
- last-use/file-age retention for HRRR combined with the existing size cap;
- synchronization preventing deletion of active run directories.

The 8 GiB cap can remain the disk backstop, but the minimum retained-run floor does not protect a selected directory during the TTL pass because TTL eviction runs first and ignores that floor.

**Required tests:**

- the most recent usable extended run survives until its successor is available;
- an obsolete extended run is eventually removed;
- active/pinned run directories cannot be purged;
- size-cap behavior remains bounded after the retention change.

---

## Verified correct in the current revision

Please preserve these parts while applying the fixes:

- contiguous `floor(departure)..ceil(arrival)` forecast-hour selection;
- Lambert-projected bilinear interpolation;
- grid-relative to earth-relative U/V rotation;
- direct DPT K→°C;
- HGT gpm treated numerically as geopotential-height metres;
- VVEL retained as Pa/s;
- CLMR/CLWMR and CIMIXR interpreted as kg/kg;
- defensive support for both operational `CLMR` and documented `CLWMR`;
- negative HRRR CIN convention with sign-flip protection;
- diagnostic mutation only after sounding replacement;
- strict all-route-point spatial coverage check;
- adjacent-range coalescing and streamed cache writes;
- UI source badge and extended-cycle freshness policy.

---

## Validation performed

- Focused HRRR/GRIB/interpolation/convection/icing suites: **821 passed, 5 skipped**
- Full Python suite: **4,371 passed, 17 skipped, 22 environment-dependent failures**
- Frontend: **556 tests passed**
- Briefing frontend build: passed
- `git diff --check`: passed
- Current GitHub checks: GitGuardian and Claude review passed

The real binary HRRR golden tests remain skip-if-absent because their GRIB sample is gitignored. The committed real `.idx` excerpt is useful for planner coverage, but it does not exercise the end-to-end advisory integration failures above.

## Merge recommendation

**Request changes until items 1–4 are corrected and covered end to end.** Item 5 should also be fixed before production use because of the download cost and active-directory race.

roznet:
## Code Review — latest head `46dcfb1c`

Full fourth-round review of PR #508, including the fixes for the previous five blocking findings. I rechecked GRIB variable interpretation, HRRR-vs-GFS behavior, spatial and temporal interpolation, advisory integration, slot/cache atomicity, efficiency, and regression coverage.

**Recommendation: request changes.** The staged HRRR replacement, native ML CAPE/CIN path, range-download atomicity, and forecast-hour selection are now sound. However, two meteorological/data-integrity problems remain, and the new tests expose a live-download regression.

### Important

1. **`src/weatherbrief/analysis/sounding/clouds.py:611-621` — HRRR cloud amount still comes from Open-Meteo, and known clear/sub-FEW cover becomes BKN.**

`build_nwp_cloud_layers_from_condensate()` receives `nwp_cloud_low/mid/high_pct` from `HourlyForecast.cloud_cover_*_pct`. Those fields are deliberately preserved Open-Meteo bulk values; `attach_nwp_diagnostics()` explicitly does not overwrite them. HRRR’s own 3 km band covers are present at `nwp_cloud_diagnostics.low/mid/high.cover_pct`, but the new builder never reads them.

Direct reproduction:

```text
HRRR low cover:                  90%
preserved Open-Meteo low cover:  10%
result mean_cloud_cover_pct:      10%
result coverage category:         BKN
```

The BKN result reveals a second bug in the same expression: `_nwp_pct_to_coverage()` intentionally returns `None` for a known value below 12.5%, including 0%, and the trailing `or CloudCoverage.BKN` converts that clear/sub-FEW result into BKN. This can activate the default Ogimet-NWP/SFIP cloud gate using both the wrong model’s cover and the wrong categorical interpretation.

**Required fix:** thread `nwp_cloud_diagnostics` into the condensate builder; prefer HRRR `diag.{low,mid,high}.cover_pct`; use Open-Meteo only when the corresponding native diagnostic is absent; and distinguish a known sub-FEW value from unknown cover rather than defaulting both to BKN. Add disagreement tests where HRRR and Open-Meteo cover differ, plus explicit 0%, 10%, `None`, and band-boundary cases.

2. **`src/weatherbrief/fetch/grib/hrrr_fetch.py:458-521` — incomplete microphysics is accepted as a complete HRRR sounding.**

The implementation and meteorological envelope require total condensate from liquid plus ice, but plan validation currently accepts any one of `CLMR`, `CLWMR`, or `CIMIXR`, and decoded-column validation accepts either `cloud_liquid_water_kg_kg` or `ice_mixing_ratio_kg_kg`.

I verified:

```text
all CLMR messages removed:   plan ACCEPTED
all CIMIXR messages removed: plan ACCEPTED
decoded liquid-only level:   complete
decoded ice-only level:      complete
```

A missing field is not equivalent to a valid zero. Losing liquid water can suppress SFIP icing and distort the new envelope; losing ice also changes its geometry.

**Required fix:** require `(CLMR or CLWMR) AND CIMIXR` independently at every planned pressure level, and require both decoded fields to be present at every retained level. A decoded numeric zero must remain valid. Add planner and decoded-column tests for missing-liquid, missing-ice, and zero-valued fields.

3. **`tests/test_hrrr.py:879-924` — three new unit tests fall through to live NOAA S3.**

The tests mock the first cached decode as incomplete. Production code then correctly deletes the rejected cache file and retries, but the tests do not mock `fetch_hrrr_idx`/`fetch_hrrr_ranges_to_cache` or provide a controlled second decode. In a restricted environment all three fail on DNS; in a networked runner each test can download a full HRRR sounding artifact, potentially hundreds of megabytes.

**Required fix:** mock the complete retry path and return a deterministic second decode, or test `_hrrr_decode_reason()` directly for the structural cases. Unit tests must not depend on NOAA availability or download full operational data.

4. **`src/weatherbrief/fetch/grib/__init__.py:2896-2901` — per-call `protect=` does not close the cross-caller purge race claimed by the comments.**

`protect=run_dir` protects only the run selected by that invocation. During an extended-cycle transition, briefing A can still be decoding the previous run while briefing B selects the newly available successor. B protects its own run and can purge A’s now-expired directory. The 9-hour TTL narrows the window but does not provide an active-reader guarantee.

**Required fix:** use shared/reference-counted active-run pins, a filesystem lock, or move TTL deletion off the decode hot path. Add a regression test with two active, different run directories. Also add direct coverage for `protect=` under TTL and size-cap eviction.

### Minor

- `src/weatherbrief/tasks/advise.py:117-126`: `nwp_condensate` and `nwp_3d` are model-native sources but are labelled `nwp_synthesized` in pilot-facing method provenance.
- `src/weatherbrief/analysis/sounding/clouds.py:515-520` and `designs/meteorology-decisions.md`: the threshold rationale is incorrect. SFIP `membership_clw(0.01 g/kg)` is 0.3, not zero. The threshold may be defensible, but it needs an accurate meteorological justification.
- `_enforce_size_cap()` excludes the protected directory from both the candidate set and the calculated total. If `protect` and `enforce_cap` are ever combined, the reported/enforced total omits the protected run’s bytes.

### Verified correct in this revision

- Contiguous `floor(departure)..ceil(arrival)` HRRR forecast-hour selection.
- Lambert-projected bilinear interpolation and all-route-point coverage validation.
- Grid-relative to earth-relative wind rotation.
- Correct unit handling: TMP K→°C, DPT K→°C, HGT gpm as geopotential-height metres, VVEL Pa/s, CLMR/CLWMR and CIMIXR kg/kg.
- Defensive support for operational `CLMR` and documented `CLWMR` aliases.
- Native HRRR ML CAPE/CIN now grades the NWP convective track with independent per-field fallback.
- Streamed, offset-ordered range download; any failed span prevents cache commit.
- Decoded sounding validation and one-time corrupt-cache retry.
- Slot-level staging: no HRRR mutation occurs unless every requested hour fetches, decodes, and validates.
- HRRR instantaneous fields remain outside GFS averaged-window handling.
- Nine-hour retention materially improves extended-cycle reuse, notwithstanding the active-reader race above.

### Validation performed

- Focused backend review suite: **902 passed, 5 skipped, 3 PR-introduced failures** (the live-NOAA tests above).
- Full backend suite: **4,389 passed, 17 skipped, 17 deselected, 25 failures** — 22 pre-existing environment-dependent failures plus the 3 new HRRR test failures.
- Frontend: **556 tests passed**.
- All frontend bundles built successfully.
- `git diff --check`: passed.
- Current GitHub checks: GitGuardian and Claude review passed.

The five real binary HRRR golden tests remain skip-if-absent because the GRIB sample is gitignored.

### Merge recommendation

**Request changes until items 1–3 are fixed.** Item 4 should also be corrected before production use because the current code explicitly claims a concurrency guarantee it does not provide.

roznet:
Rounds 3–4 addressed in d12b6aa (on top of `46dcfb1c`). I verified each finding against the code and, where it mattered, against the live product data before acting. All four Important items and the three minors are fixed; nothing was declined this round.

## Important items

**1. Condensate envelope's amount half wasn't model-native + known sub-FEW became BKN** — both confirmed. `build_nwp_cloud_layers_from_condensate` now receives `nwp_cloud_diagnostics` and prefers the diagnostic band covers (HRRR's own 3 km values) over the bulk `cloud_cover_*_pct` Open-Meteo fields, exactly the preference `_build_grib_layers` already applies for GFS — the round-4 repro (HRRR 90% / Open-Meteo 10%) now yields 90%/OVC. And the `x or BKN` expression is split: a **known** cover maps honestly (sub-FEW → FEW with the real percentage recorded — the condensate at the interpolated point proves the deck exists; the published percentage states its extent; BKN would contradict the model's own number and dropping the layer would contradict its microphysics), while only genuinely **unknown** cover takes the BKN default with `mean_cloud_cover_pct=None`. Tests cover the disagreement case, bulk-only fallback, explicit 0%, 10%, 12.5% boundary, 30/50/90%, and the unknown-cover default.

**2. One-sided microphysics accepted as complete** — confirmed and fixed at both validation stages: the plan now requires liquid (`CLMR|CLWMR`) **and** ice (`CIMIXR`) independently at every offered level, and the decoded column requires both fields present on every level. Decoded zeros remain valid — and that's safe by measurement, not assumption: the live sample shows zero NaN-masking on both fields (~98% real 0.0 values on clear cells), so a clear column can never be rejected as "missing". Tests: liquid-only rejected, ice-only rejected, zero-valued both-present passes.

**3. Tests falling through to live NOAA S3** — confirmed (the corrupt-cache retry path called the real `fetch_hrrr_idx`). The `_stage` helper now mocks the full retry path: `fetch_hrrr_idx` returns the committed real-idx fixture (so plan validation still runs against genuine inventory lines) and `fetch_hrrr_ranges_to_cache` is stubbed; a new retry-success test drives bad-cached-decode → unlink → re-download → good decode and asserts exactly one re-fetch. The whole `test_hrrr.py` file was verified green with networking disabled (`HTTPS_PROXY` pointed at a dead port) — 98 passed, no DNS, no downloads.

**4. `protect=` doesn't close the cross-caller race** — agreed; the per-call argument is gone, replaced by a reference-counted in-process pinned-run registry (`cache.pin_run_dir`). Enrichment pins its run for the whole stage+commit span, and **both** eviction rules (TTL and size cap) skip every pinned directory *whoever pinned it* — briefing B's purge can no longer delete briefing A's older run mid-decode during an extended-cycle transition. The size-cap accounting minor is folded in: pinned directories are never victims but their bytes **do** count toward the enforced total, so the cap logs "NOT met" instead of declaring victory over an omitted directory. In-process scope is deliberate and documented: every purge/enrichment/scheduler caller lives in the single uvicorn worker (the same standing assumption refresh-durability is built on), so no filesystem lock is needed. Regression tests: cross-caller TTL survival with two expired runs, refcounted double-pin, and cap-skips-pinned-but-counts-bytes.

## Minor items

- **`nwp_condensate`/`nwp_3d` badged "synthesized"** — fixed: all three native envelope sources (`grib`, `nwp_3d`, `nwp_condensate`) now badge `cloud_method_effective="nwp"`; `nwp_synthesized` is reserved for genuinely non-native layers. This also corrects the pre-existing `nwp_3d` mislabeling round 3 noted. The value vocabulary is unchanged, so no client-side impact.
- **SFIP threshold rationale factually wrong** — confirmed (`membership_clw(0.01 g/kg)` is 0.3, not 0; the ramp was misread). The code comment and §22 now give the honest justification: 1e-5 kg/kg is the conventional WRF-family cloud-mask threshold (the model family HRRR belongs to), sitting at the *bottom* of SFIP's first liquid-water ramp — the threshold value is unchanged, only the false claim about it.
- **`replaced == 0` commit-loop coverage** — added the `side_effect=[1, 0, 1, 1]` mixed-result test: the unmatched hour is skipped without its diagnostics landing alone (3 diag applies for 3 merged hours), while the slot still commits.

## Validation

- `test_hrrr.py` (98) green **with networking disabled**; golden real-GRIB tests (5) green with the sample present.
- Cloud/ceiling + condensate suite, GRIB/cache/fill/interpolation-audit, convective, clouds, method-provenance, and the full `tests/analysis` tree: 1,193 + 41 passed.
- Full backend suite: **4,358 passed / 63 failed** — the 63 are the same pre-existing euro_aip-stub environment failures as the `main` baseline (this session can't install the private package); zero new failures.
- Docs brought current: the weather-engine-specs HRRR section now describes the staged slot commit, 9h TTL, pins, and dual-microphysics validation as they exist at this head; §22 corrected accordingly.

---

roznet:
## Code Review — latest head `d12b6aa6`

Fifth-round review of the full cumulative PR, with a focused incremental review from `46dcfb1c` to `d12b6aa6`. I rechecked the previous findings from the actual implementation, including the implementation agent's evidence around operational `CLMR`, memory-efficient staging, and test-fixture policy.

### What this revision fixed correctly

All findings from my previous posted review are addressed:

- Operational HRRR `CLMR` is handled correctly; `CLWMR` remains a defensive alias rather than an assumed operational name.
- The planner and decoded-column validation now independently require liquid (`CLMR` or `CLWMR`) **and** ice (`CIMIXR`), while accepting real numeric zeroes.
- The condensate envelope now prefers HRRR's native diagnostic band-cover values over the deliberately preserved Open-Meteo bulk values.
- Known sub-FEW coverage no longer falls through to the unknown-cover BKN default.
- The corrupt-cache tests are hermetic and exercise the retry path without contacting NOAA.
- Active-run protection is now reference-counted across callers, and pinned bytes count toward the size cap.
- `nwp_3d` and `nwp_condensate` are correctly labelled as native NWP sources.
- The false statement that `0.01 g/kg` has zero SFIP membership was removed.

The decoded staging design is also appropriately memory-conscious: it retains small per-point columns, not a deep copy of the full GRIB data or briefing pack. A large committed binary GRIB fixture is not required for merge; the `.idx` fixture provides deterministic planner coverage, while the absent binary sample remains a disclosed residual limitation.

### Blocking findings

#### 1. The COMMIT phase is not exception-atomic

`src/weatherbrief/fetch/grib/__init__.py:2967-2986`

The STAGE phase is correctly guarded, but the subsequent mutation loop has no exception boundary or rollback. If a later `_replace_pressure_levels_from_grib()` or `_apply_cloud_diagnostics_to_sections()` call raises, earlier hours have already been changed to HRRR.

The exception then escapes `_enrich_gfs_from_hrrr_pinned()` and `_enrich_gfs_from_hrrr()`, reaches the blanket handler in `_try_enrich_gfs_from_hrrr()`, and returns `None`. The caller proceeds through plain-GFS enrichment and records `gfs:noaa`, while the earlier HRRR mutations remain in the pack.

I reproduced this directly: the second commit operation raised while the first HRRR mutation remained present.

This does **not** require a full-pack deep copy. Suitable fixes include:

- build the replacement `pressure_levels`/diagnostic assignments as mutation-free objects during staging, then swap references only after every construction succeeds; or
- retain rollback state only for the hourly entries touched by the commit.

Add a test where the second replacement and a later diagnostic application raise, and assert that the pack is either wholly unchanged/GFS or wholly HRRR with matching provenance.

#### 2. HRRR LCDC/MCDC/HCDC are selected using the wrong vertical definitions

`src/weatherbrief/analysis/sounding/clouds.py:637-658`

The condensate builder currently selects the native low/middle/high diagnostic using fixed 6,500/20,000-foot ICAO boundaries. Those are not the definitions of the HRRR products being ingested.

NOAA's RAP/HRRR diagnostic documentation defines:

- low cloud fraction: surface to 642 hPa;
- middle cloud fraction: 642 to 350 hPa;
- high cloud fraction: 350 hPa to model top.

Source: https://rapidrefresh.noaa.gov/RAP_var_diagnosis.html

Direct reproduction with a condensate deck from 3,609–11,811 ft:

```text
HRRR LCDC: 90 %
HRRR MCDC:  0 %
deck midpoint: 7,710 ft
result: FEW / 0 % for the entire deck
```

At roughly 7,700 ft that column remains well below 642 hPa and belongs to HRRR's **low** layer, but the altitude-based selector uses MCDC. This can turn a native overcast low deck into FEW for cloud/VFR/ceiling consumers.

Use source-specific pressure boundaries for HRRR. A condensate run crossing 642 or 350 hPa should ideally be split so each segment receives the corresponding native cover, rather than assigning one midpoint band to the whole vertical deck. The Open-Meteo fallback can retain its own source-specific band semantics.

#### 3. The `1e-5 kg/kg` cloud threshold is not supported by HRRR's documented diagnostic thresholds

`src/weatherbrief/analysis/sounding/clouds.py:513-525` and `designs/meteorology-decisions.md` §22

The revised rationale calls `1e-5 kg/kg` the standard WRF-family cloud-mask threshold. HRRR's own NOAA diagnostic documentation instead describes:

- combined cloud water plus ice above `1e-7 g/g` for explicit cloud-base/ceiling detection;
- `1e-6 g/g` for cloud-top detection.

Source: https://rapidrefresh.noaa.gov/Diag-vars-NOAA-TechMemo.pdf

Those are respectively 100× and 10× below the application threshold. That does not prove the application must copy either diagnostic threshold verbatim—the pressure-level product and this aviation envelope are not identical algorithms—but it does mean `1e-5` cannot be justified as the documented HRRR/WRF convention without evidence. At the current value, thin but genuine condensate layers can produce `[]`, which downstream consumers interpret as a genuinely clear model column.

Either align the threshold with the HRRR diagnostic definition, or document and test an independently calibrated threshold against real HRRR columns/ceilings. Include thin liquid, thin ice, mixed-phase, and clear-noise cases around the chosen cutoff.

### Important findings

#### 4. The pin registry has a snapshot/delete race

`src/weatherbrief/fetch/grib/cache.py:519-530` and `:431-449`

`purge_old_runs()` snapshots `_pinned_run_dirs()`, releases `_PIN_LOCK`, and only then walks and deletes directories. A second thread can pin a directory after that snapshot but before `shutil.rmtree()`.

I reproduced the race deterministically: a directory was deleted while its `pin_run_dir()` context was active. The size-cap path has a wider version of the same race because directory sizing occurs between the pin snapshot and deletion.

Re-check each victim under the same lock immediately before deletion, preventing pin acquisition until the deletion decision/action is complete, or use an equivalent reader/evictor synchronization scheme. Add a barrier-based test where a pin is acquired after purge begins but before the candidate is removed.

#### 5. `nwp_condensate` is omitted from DD-vs-NWP agreement

`src/weatherbrief/analysis/advisories/dd_nwp_agreement.py:192-205`

The advisory considers only `nwp_3d` and `grib` native:

```python
if cl.source in ("nwp_3d", "grib")
```

The new HRRR `nwp_condensate` source is therefore discarded. When HRRR produces a genuine condensate deck, `has_native_nwp` becomes false and the cloud-overlap comparison is silently skipped. The native-source classification was fixed in `tasks/advise.py` but not propagated to this sibling advisory.

Add `nwp_condensate` and cover both a non-empty HRRR deck and the native clear-column `[]` case.

### Minor

- `designs/fetch.md` still says HRRR has a 6-hour cache TTL; runtime is now 9 hours.
- `EnhancedCloudLayer.source` documentation omits `nwp_3d` and `nwp_condensate`, which encourages incomplete source enumerations.
- HRRR wind rotation silently does nothing if `GRIB_uvRelativeToGrid` or the Lambert attributes disappear. Since unrotated HRRR winds can carry a material directional bias, log loudly or fail the HRRR hour rather than silently accepting them.
- The domain tests do not pin exact inclusive Lambert-grid edges or just-outside values.

### Validation

- Focused HRRR/GRIB/cloud/icing/advisory suite: **1,277 passed, 5 skipped**
- Full backend suite: **4,409 passed, 17 skipped, 17 deselected**
- Remaining **22 failures** are the same environment-dependent auth/process-pool/precache failures; there are no new PR-specific failures
- `git diff --check`: passed
- GitGuardian and Claude review checks: passed
- No frontend files changed in this latest commit

### Merge recommendation

**Request changes.** Items 1–3 are blocking correctness/meteorology issues. Items 4–5 should also be fixed before production because they undermine the new cache-safety guarantee and omit HRRR from a model-agreement advisory.

roznet:
Round 5 addressed in 1f30050. My judgement first, then the disposition: these were **not** fix-on-main material — three of the findings sat squarely inside the guarantees this PR advertises (slot atomicity, model-native amounts, the domain gate's honesty), so they warranted one more round before merge. All five findings plus the minors are fixed; nothing declined.

## Blocking items

**1. COMMIT phase not exception-atomic** — confirmed; the "nothing can escape after mutation" comment was asserted, not enforced. The commit loop is now bracketed by an exact rollback journal: before any mutation, every gfs-slot hourly's writable state is snapshotted via new model methods `nwp_state_snapshot`/`restore_nwp_state` (they live on `HourlyForecast` so the diagnostics/`freezing_level_m` mirror pair stays consistent and the existing no-direct-assignment source guard keeps holding). The replacement paths swap whole references, so the journal holds references, not copies — no deep copy, per your suggested fix shape. On any commit failure the pack is restored reference-for-reference and the caller's plain-GFS fallback proceeds clean. Tests raise from the second `_replace` call and from a post-replace diagnostics application, both asserting the original `pressure_levels` list *identity* is restored on a real cross-section.

**2. LCDC/MCDC/HCDC vertical definitions** — confirmed, and the best catch of the round. Band membership for the diagnostic covers now uses NCEP's pressure bounds (surface–642 hPa / 642–350 / 350–top), and a condensate run crossing a boundary is **split** so each segment takes its own band's cover, as you suggested — one midpoint band no longer grades a deep deck with the wrong product. Your repro is a committed test: deck ~3,600–11,800 ft with LCDC 90 / MCDC 0 now yields a single OVC/90 layer (was FEW/0), plus a boundary-crossing test asserting OVC-90-below/FEW-20-above abutting at the level midpoint. The Open-Meteo bulk fallback keeps ICAO altitude bands — its own product's semantics.

**3. The 1e-5 threshold** — you were right that the justification didn't hold, and the live data settles it. Measured on the real f06 product (full-CONUS 500+700 hPa): **liquid (CLMR) is GRIB-quantized in 1e-5 steps** — the smallest representable nonzero liquid value *is* 1e-5, so no liquid threshold ≤1e-5 can differ; but **ice (CIMIXR) packs to a 1e-9 quantum with 98.9% of nonzero values below 1e-5** (median exactly your 1e-7) — the old threshold was silently erasing genuine thin cirrus, while the true numerical-residue cluster sits at 1e-9–5e-9. New threshold: **1e-7 kg/kg**, i.e. HRRR's own documented cloud-base detection threshold from the Tech Memo you cited, now with the packing measurement as the calibration evidence in §22 (and the twice-wrong rationale replaced, with both misreadings recorded). Tests: residue-cluster values (1e-9, 5e-9) stay clear, 1e-7 ice-only cirrus is a deck, 9e-8 is not, liquid behaviour unchanged by construction.

## Important items

**4. Pin snapshot/delete race** — confirmed. Victims are now re-checked against the registry **under the pin lock immediately before `rmtree`**, in both the TTL and size-cap paths (deletion itself also runs inside the lock, so a pin can't be acquired mid-delete). The residual window — a pin landing after that final check — is safe by construction and documented: pinning precedes the pinner's *first* cache read, so a directory lost in that sliver presents as an ordinary cold cache and the existing corrupt-cache/re-fetch path recovers; what the re-check protects is the dangerous case, a reader pinned before the deletion decision, possibly mid-decode. Deterministic regression test simulates a pin arriving mid-purge (inside the age check, i.e. after the snapshot) and asserts the directory survives.

**5. `nwp_condensate` missing from `dd_nwp_agreement`** — confirmed; and since this is the second consumer to drift on a hand-written native-source list, the fix removes the pattern rather than patching the instance: a shared `NATIVE_NWP_CLOUD_SOURCES` constant in `models/analysis.py` now classifies nativeness for both the provenance badge and the agreement advisory, with a drift-guard test asserting both consumers reference the shared set and that all three native sources are members.

## Minors — all taken

`designs/fetch.md` cache row corrected to the 9h HRRR TTL (+pins); wind rotation no longer skips silently — a missing `uvRelativeToGrid` logs "cannot verify wind orientation" and a grid-relative flag with unusable Lambert attrs logs the explicit ±15–20° bias warning; `EnhancedCloudLayer.source` docstring enumerates all five sources and points at the shared native set; domain-gate tests pin both Lambert-grid corners at 1e-6 cells inside (the exact corner is untestable through lat/lon — independently computed longitudes differ from the grid constants by float-wrap epsilons that straddle the boundary either way, which the test comment records) plus just-outside points on all four sides.

## Validation

- Focused suite (HRRR + sample goldens + cloud/ceiling + GRIB/cache/fill/audit + convective + clouds + ICON-D2 + full `tests/analysis`): **1,362 passed**, 1 pre-existing euro_aip-stub failure.
- Full backend suite: **4,372 passed / 63 failed** — the same 63-failure environment baseline as `main`, zero new.
- No frontend changes this round.

From my side this now covers every finding from all five rounds; barring new findings I'd consider it mergeable.

---

roznet:
**Round 6 verdict: technically not a blocker, but fixed anyway — it was cheaper to fix than to argue about (044144d).**

The surviving Important finding was real but narrow: in pressure-band mode, a segment whose diagnostic band cover was `None` fell back per-band to the Open-Meteo bulk percentage — a number measured over the ICAO altitude slabs, pasted onto a segment carved on NCEP's 642/350 hPa boundaries. It only bites when HRRR publishes *some* band covers but not all (uncommon), and the failure is a slightly-wrong extent grade, not a wrong deck. So: not merge-blocking on impact.

But it's the same bug class as rounds 3–5 (mixing vertical semantics between products), and the specs doc already claimed the correct behaviour — so the fix (~10 lines) landed now rather than as main-branch debt:

- `use_pressure_bands` is now true only when at least one diagnostic band cover exists; in that mode amounts come **exclusively** from the diagnostic triple. A missing band degrades to the existing unknown-extent BKN default (`mean_cloud_cover_pct=None`) instead of borrowing the bulk number.
- The bulk triple + ICAO bands are used only when no diagnostic band cover exists at all (diag absent, or diag carrying only ceiling/base) — amounts and membership always describe the same vertical slice.
- Regression test added (diag low=90 / mid absent / high=0, bulk mid=10 → mid-band deck reads BKN/None, low deck still OVC/90); docs (§22 + specs) updated to match.

Focused suites green (147), full suite at the environment baseline. From my side the PR is mergeable.

---

Issue #510, opened 2026-07-28
User-declared unpublished approaches (cloud break procedures) to damp Approach Feasibility false REDs


roznet:
Follow-up to #509 (Approach Feasibility advisory). Can ship independently, after it.

## Problem

#509 grades a destination with no published instrument approach as **RED** at IFR/LIFR
and **AMBER** at MVFR. That is correct as a default, but it over-fires in the UK, where a
field can have no AIP-published procedure while pilots routinely fly a self-briefed
approach — a **cloud break procedure** or a private/unpublished GNSS approach.

Concretely: `EGTF` (Fairoaks) has **zero** rows in `procedures`. For a pilot based there,
*every* IFR arrival at their home field would grade RED, when in practice it is an amber
— a real consideration, not a stopper. Same shape at EGSX North Weald, EGLK Blackbushe,
EGHR Goodwood and many other UK fields. Left alone this is enough noise to make the
advisory worth ignoring, which is worse than not having it.

Note this is *not* a data gap — #509 establishes that these absences are genuine (the
airports really have no published procedure). This is a case where the published data is
right and the operational reality differs.

## Proposal

A user-level list of airports where the pilot has an unpublished approach available to
them. A declared field is treated as **"an approach exists, alignment unknown"**.

### Storage

- **User preferences (`app_prefs_json`)**, not a profile advisory param. This is a fact
  about the pilot and the airport ("I'm current on the Fairoaks cloud break"), not about
  a flight profile — entered once, applies everywhere.
- Follows the existing `api/preferences.py` pattern: a `_parse_*` helper alongside
  `_parse_advisory_prefs` / `_parse_digest_config_from_prefs`, and the merge-preserving
  write at `api/preferences.py:296-372` so sibling keys survive.
- Deliberately **not** an `AdvisoryParameterDef`: `default` is typed `float` and all 83
  params in use are numeric (percent/altitude/number/speed/distance/boolean). Adding a
  text param type would touch the model, validation, web settings, iOS parity and the MCP
  catalog — for one advisory.

### Input format

- Free-form text field accepting **multiple ICAOs, space- or comma-separated**
  (`EGTF, EGSX EGLK`).
- Normalized on save: uppercase, trimmed, deduped, stored as a canonical JSON array.
- Validated against `nav.db`; **unknown codes are surfaced back to the user**, never
  silently dropped — a typo'd `EGFT` must not quietly become "no declaration".

### How it feeds the evaluator

[...]

PR #511, opened 2026-07-28
feat(advisories): approach feasibility — is the IAP aligned with the wind runway?

### LINKED ISSUE #509: Approach Feasibility advisory: is the instrument approach aligned with the runway the wind favours?
## Problem

When the destination is MVFR/IFR/LIFR, the briefing tells the pilot the ceiling, the
visibility, and the best runway by wind — but never whether those three facts are
*compatible*. Nothing in the system answers the actual arrival question:

> **Can I get in, on a runway I can also land on?**

Two concrete gaps:

1. **`enrich_wind()` picks the best runway with zero approach awareness.**
   `analysis/airport_consensus.py:122` selects `min(crosswind_kt, -headwind_kt)` across
   all runway ends. So the briefing can currently say *"destination IFR, ceiling 600 ft,
   best runway 24, crosswind 3 kt"* at a field where 24 has no approach at all and the
   ILS serves 06.

2. **Approach handling elsewhere is binary.** `tasks/alternates.py` asks only
   "does this field have *any* IAP" (`has_instrument_approach`, `best_approach_type`).
   Runway **alignment** appears nowhere in the codebase.

This is a joint function of ceiling **and** wind **and** infrastructure. No existing
evaluator owns that intersection: `flight_category` grades ceiling/vis, `airport_wind`
grades wind, `ifr_feasibility` composites airport+icing+convective.

### The premise checks out against real data

Queried `data/nav.db` (`procedures` + `runways`):

| Check | Result |
|---|---|
| Approach procedures total | 5,516 |
| …carrying a `runway_ident` | ~5,250 (95%) |
| …that join a live runway end | 5,055 (96% of those) |
| …where that end has a true heading | 4,988 (99% of those) |
| **End-to-end usable for alignment** | **4,988 (90%)** |

And the misalignment case is the *common* one, not an edge case:

- Of **677** airports with at least one IAP-matched runway end, **254 (38%)** have
  approaches on only *some* of their ends.
- Examples: `EGKA` — IAPs on 02/20, nothing on 06/24 or 13/31. `EGBJ` — 09/27 only.
  Same shape at EGNH, EGNJ, EGPK, EGPE, EGNS, EGPA.

Circling-only approaches are detectable via the ICAO single-letter suffix
(`RNP A`, `NDB C`, `LOC A`) — but only **39** of the 262 ident-less rows are real
circling procedures; the other **223** are parser artifacts
(`MIPS: RNP (LNAV) ARINC CODING`). A naive "no runway ident ⇒ circling" rule would be
wrong ~85% of the time.

## Proposal: a new `approach_feasibility` advisory

Considered folding into `airport_wind`/`flight_category` (rejected: a RED "Airport Wind"
that is actually red because there is no ILS is misleading, and it silently changes
grades for users who tuned those params) and into `ifr_feasibility` (rejected: it is
already a 3-axis composite; a 4th axis flattens the actionable content into one word,
and it can't carry its own params).

A dedicated advisory gets one clear semantic, its own tunable params, and its own
drillable detail payload for web / iOS / `get_advisory_detail` / digest. `ifr_feasibility`
can later **consume** its status as its airport axis rather than re-deriving it.

**Name:** "Approach Feasibility" (not "IFR Approach" — the misalignment penalty, a
tailwind on the only served end, is real even when it isn't strictly an IFR problem).

## Grade map

The destination's flight category selects which logic applies. The key principle: the
alignment/circling/tailwind reasoning only earns its keep once the pilot can no longer
be expected to complete the arrival visually.

**VFR → GREEN, always.** The field is visually reachable; approach infrastructure is
irrelevant. (Not UNAVAILABLE — see the note below.)

**MVFR → IAP presence only, capped at AMBER.**
- Any approach exists (aligned or not) → **GREEN**
- No approach at all → **AMBER**
- Alignment, circling and tailwind logic is **not** applied at MVFR: with a
  1,000–3,000 ft ceiling the pilot is expected to be able to complete the landing
  visually, so a misaligned or circling-only approach is not a penalty.
- **Never RED at MVFR.**

**IFR / LIFR → full logic.**
- **GREEN** — straight-in approach aligned with a runway whose wind components are
  within limits, ceiling/vis clear of the minima band.
- **AMBER** — an approach is available but with a compromise: circling required (and
  ceiling supports it); or aligned but ceiling in the marginal band; or aligned with a
  tailwind that is within limits; or models disagree on which end the wind favours.
- **RED** — no IAP at all; **or** ceiling below even the *best-case* DH for the
  available approach; **or** misaligned + ceiling won't support circling + the tailwind
  on the approach-served end exceeds limits.

**UNAVAILABLE** — only when the evaluator could not run at all, i.e. `arrival_approaches`
is absent from the `RouteContext` (old pack, or the collection step didn't run). This is
the `route_fronts` / `sun` precedent. It is **not** used for "conditions are fine"
(`AdvisoryStatus.UNAVAILABLE` is documented as *"we could not assess this... absent data
is not a clear sky"*, so a benign assessment is GREEN), and it is **not** used for an
airport with no approaches — see below.

### Absence of procedure data means "no approach"

**Zero `procedures` rows for an airport is treated as "no instrument approach available",
full stop.** The evaluator makes no attempt to distinguish "genuinely has none" from
"never ingested", and there is deliberately no third state for it.

This is safe because the data is trustworthy where it matters. Spot-checking UK fields:
every airport that should have an approach does (EGLL, EGKK, EGHI with 10 procedures,
EGLF, EGMD, EGKA, EGBJ, EGKB, EGTK), and every absence is genuine — **EGTF Fairoaks,
EGSX North Weald, EGLK Blackbushe, EGHR Goodwood** have no instrument approach in
reality. GB's 56% IAP rate among medium/large airports is the true ratio of licensed
aerodromes with procedures, not a parser gap.

Per-country coverage is also cleanly **bimodal** — countries sit at 40–100% or at exactly
0.0%, with nothing in between:

| Well covered | 0.0% coverage |
|---|---|
| BE/EE/HR 100%, NO 98%, PL 96%, ES 94%, NL 93%, DE 81%, FR 80%, GB 56%, IT 59%, PT 48% | US, CA, RU, UA, MX, and all other non-European |

The zero-coverage countries are all outside the current European scope, so the rule
cannot misfire today. **Caveat for the US-expansion work:** in a 0%-coverage country this
rule would grade *every* IFR destination as "no IAP" → RED. Whoever extends coverage
beyond Europe must revisit this before enabling the advisory there.
(`airac_country_coverage` and `aip_entries` are both empty in `nav.db`, so there is no
per-airport provenance signal to lean on instead.)

## Design constraints

These are the things that make this correct rather than merely plausible.

1. **Reuse `APPROACH_CLASS_PROXY`.** `analysis/alternate_requirement.py:73` already
   holds DH bands per approach class, with `proxy_for_approach()` at `:123` handling the
   unknown-type → most-demanding degradation. Do **not** introduce a second minima table
   — two different numbers answering the same question on the same page is a drift bug
   waiting to happen.

2. **Asymmetric uncertainty.** Those DH values are *estimates* — which is exactly why
   `alternate_requirement` deliberately reports Likely/Marginal/Unlikely instead of a
   traffic light. Emitting GREEN/AMBER/RED from the same estimated data makes a soft
   claim look hard. **Rule: estimate uncertainty may push toward AMBER, never toward
   RED.** RED requires either a hard fact (no IAP) or a ceiling below the *best-case* DH.

3. **Flag circling, don't grade it.** `nav.db` has no circling minima, and
   "circling not authorised" / night restrictions / category limits aren't in the data at
   all. Surface *"straight-in not available — circling required"* as a flag; never compute
   a circling verdict.

4. **Alignment is a brittle function of a noisy input.** Wind direction at D-3 can spread
   180° across models, and alignment is discrete. Evaluate per model and let the existing
   majority aggregation do its job; near the alignment boundary degrade to AMBER rather
   than flipping GREEN↔RED. At D-0 prefer TAF over NWP — otherwise this card will
   contradict the alternates card sitting next to it on the same page.

## Resolved decisions

- **Applicability: follow the existing precedent — no self-gating.** `default_enabled=True`,
  evaluated always, turned off per profile via the normal advisory toggle. This matches
  `vfr_feasibility` / `ifr_feasibility`, neither of which self-gates on pilot or aircraft
  capability. `UserAircraftRow.is_ifr` exists and was considered as a gate, but using it
  here would break that pattern and would silently no-op when a flight has no aircraft
  attached. The duplicate-RED risk for a VFR-only pilot is largely absorbed by the grade
  map itself: VFR is GREEN and MVFR caps at AMBER, so this only reaches RED at genuinely
  IFR destinations — which `vfr_feasibility` is already grading RED anyway.

- **Tailwind limit is an advisory param, not an aircraft field.** `UserAircraftRow` has
  `is_ifr` / `is_fiki` / `cruise_speed_kt` / `ceiling_ft` but **no** wind limits; the
  established pattern is per-profile advisory params (`airport_wind` uses
  `crosswind_green_kt`=15 / `crosswind_red_kt`=25). Default 10 kt. No new aircraft column.

- **Category: `flight_rules`**, joining `vfr_feasibility` / `ifr_feasibility`.

- **Card density: no new work required.** `advisories-ui.ts:515` already routes
  unanimous-green and UNAVAILABLE advisories into the compact "all clear" pill grid,
  reserving full cards for red / amber / green-mixed. The desired "quiet on normal days"
  behaviour is the existing default.

## Implementation sketch

**Data plumbing** (new, mirrors `get_runway_ends`)
- `airports.py`: `get_runway_approaches(icao_codes, db_path) -> dict[str, list[RunwayApproach]]`
  — per runway end, the approaches serving it (type + precision rank), plus the
  circling-only set identified by letter suffix, plus an explicit
  "no procedure data for this airport" signal distinct from "no approaches".
- Model: `RunwayApproach` alongside `RunwayEnd` in `models/airport_conditions.py`.
- `RouteContext`: new optional field (`arrival_approaches: ... | None = None`), following
  the `route_fronts` / `sun` precedent — `None` ⇒ evaluator returns UNAVAILABLE, so old
  packs degrade cleanly.
- **Two wiring points, both required** (the recalc path must not silently lose it):
  `pipeline.py:555-579` and `tasks/advise.py:282-302` — the same pair that already
  collect `runway_data`.

**Evaluator**
- `analysis/advisories/approach_feasibility.py`, `@register`, `category="flight_rules"`,
  `timing_class="cheap"`, `default_enabled=True`.
- Add to `ENTRY_ORDER` in `registry.py:53`.
- Params: `tailwind_limit_kt` (default 10), crosswind limit, circling-ceiling assumption,
  whether to require straight-in.
- Reads `airport_conditions.arrival` — `all_runways` already carries per-end
  crosswind/headwind per model, so the wind half needs no new computation.

**Consider (separate follow-up, not this issue):** an approach-aware "best usable runway"
*alongside* `best_runway` in `enrich_wind` — not replacing it, since the pure-wind answer
is the correct one for VFR arrivals and for the forecast map.

**Surfacing**
- Web advisory card + detail popup; iOS parity (run `sync-ios-web`); digest wiring;
  MCP `get_advisory_detail`.
- Digest phrasing stays in the "directs attention" register:
  *"wind favours 24; the ILS serves 06 — plan for circling or an 8 kt tailwind"* —
  never a usability verdict.

## Follow-up

- **#510 — user-declared unpublished approaches (cloud break procedures).** A user-level
  list of fields where the pilot has a self-briefed approach, damping the "no IAP → RED"
  case that would otherwise fire on every IFR arrival at UK GA bases like EGTF. Ships
  after this issue; the grade-map interaction is specified there.

## Out of scope

- Real plate minima from a procedures DB (would replace the proxy bands entirely —
  already tracked as a follow-up on the alternate-requirement side).
- Per-candidate approach feasibility for **alternates** — natural next step, deliberately
  deferred so the destination case lands first.
- Circling minima, night/category circling restrictions, "circling not authorised".
- Changing `enrich_wind`'s `best_runway` semantics.
- Gating on `aircraft.is_ifr` (considered and rejected — see Resolved decisions).
- Distinguishing "genuinely no approach" from "procedure data never ingested" — decided
  against; absence is treated as absence. Must be revisited if coverage extends to a
  0%-coverage country (see the grade map).

roznet:
## What & why

Adds a dedicated `approach_feasibility` evaluator answering the arrival question no existing evaluator owns: **can I get in, on a runway I can also land on?**

`enrich_wind` (`analysis/airport_consensus.py:122`) picks the best runway with **zero approach awareness** — `min(crosswind_kt, -headwind_kt)` across all ends — so the briefing could say *"destination IFR, ceiling 600 ft, best runway 24"* at a field where 24 has no approach and the ILS serves 06. `flight_category` grades ceiling/vis, `airport_wind` grades wind, `ifr_feasibility` composites airport+icing+convective; nothing joined ceiling **and** wind **and** infrastructure. This does, with its own semantic, its own tunable params and its own drillable detail payload. `enrich_wind`'s `best_runway` semantics are deliberately unchanged — the pure-wind answer is still correct for VFR arrivals and for the forecast map.

**Data plumbing.** `airports.py::get_runway_approaches` mirrors `get_runway_ends` (the wind side of the same question) and returns `AirportApproaches` per ICAO. Each `RunwayApproach` resolves into exactly one of three states, and keeping the third apart is the point:

| State | Means |
|-------|-------|
| `runway_id` set | straight-in to that runway END — only when the procedure's ident joins a **live** (non-closed) end with a true heading, so it can always be paired with that end's wind |
| `circling=True` | ICAO letter-suffixed circling-only procedure (`RNP A`, `NDB C`) |
| neither | alignment unresolved — treated as *uncertainty*, never hard misalignment |

Circling detection keys on the single trailing letter, excluding X/Y/Z (straight-in *variant* designators) and any name carrying a digit. That digit guard is what matters: only **39** of the 262 ident-less rows in `nav.db` are real circling procedures, so a naive "no runway ident ⇒ circling" rule would be wrong ~85% of the time (`MIPS: RNP (LNAV) ARINC CODING` and friends).

The container carries a fourth, airport-level state — `lookup_failed` — see UNAVAILABLE below.

**Grade map**, gated on the destination's flight category — the alignment/circling/tailwind reasoning only earns its keep once the pilot can no longer be expected to complete the arrival visually:

| Category | Logic |
|----------|-------|
| **VFR** | GREEN, always. Visually reachable; infrastructure irrelevant (GREEN not UNAVAILABLE) |
| **MVFR** | IAP presence only, capped at AMBER. Any approach → GREEN, none → AMBER. **Never RED** |
| **IFR / LIFR** | Full logic. GREEN = straight-in to an end whose wind is within limits with ceiling/vis clear of the estimated minima band; AMBER = a compromise; RED = a hard fact only |

**UNAVAILABLE** is reserved for "the evaluator could not run": `arrival_approaches` absent (old pack / no nav.db), no per-model airport condition, or `lookup_failed` — the ICAO was unknown to `nav.db`, or the procedure query raised. Zero procedure rows is graded as *no instrument approach*, full stop — no third state there, per the issue's decision. Documented caveat for the US-expansion work: in a 0%-coverage country that rule would grade every IFR destination RED.

`lookup_failed` exists because "this field has no instrument approach" is the **most severe input the grade map takes** (RED at IFR/LIFR), and it arrives as the same empty approach list a parse failure would produce. Only the flag tells them apart, so every path in `get_runway_approaches` that fails to *determine* the picture sets it rather than returning an empty list — a data problem must not be able to manufacture a RED.

**Two rules keep it honest:**
1. **One minima table** — decision heights come from `analysis/alternate_requirement.py::APPROACH_CLASS_PROXY` via `proxy_for_approach`, the same estimates the alternate-requirement card shows. No second table to drift. Circling-only procedures are **excluded** from the airport-wide best-case gate: that table holds *straight-in* minima, and real circling minima are categorically higher for the same class.
2. **Asymmetric uncertainty** — estimate uncertainty may push toward AMBER, never toward RED. RED needs no IAP at all, a ceiling/visibility below even the *best case*, an approach-served end with an out-of-limits tailwind **and** conditions that will not support circling, or that same best-case test applied per approach to an end the wind is happy with. A forecast *inside* the band is AMBER — the AMBER middle is exactly the width of our uncertainty about the plate. Neither an approach of unresolved alignment, nor a served end the wind model did not cover, nor a failed lookup can drive RED.

**The no-usable-straight-in verdict is a cross product**, and is built as one: `_Blocker` (`MINIMA` / `MISALIGNED` / `NO_STRAIGHT_IN`) × `_Softening` (`CIRCLING` / `UNRESOLVED` / `NO_WIND_DATA` / `NONE`), composed by `_fallback_detail` as `"{cause} — {clause}"`. Every cell has correct text by construction, and a new blocker or reason costs one string rather than a row or column. A matrix test asserts every cell resolves in all four locales and that no non-circling reason recommends circling.

**Circling is flagged, never graded** — `nav.db` has no circling minima. `circling_ceiling_ft` / `circling_visibility_m` only decide whether a blocked arrival softens to AMBER; circling needs *both*, since a generous ceiling in fog is not a way in. **Alignment is graded per model** so the registry's majority aggregation absorbs cross-model wind spread, and crossing the tailwind limit only removes the straight-in option, so the verdict degrades to AMBER near the boundary rather than flipping GREEN↔RED.

**Wiring — both context builders plus the alt path.** `RouteContext.arrival_approaches` follows the `route_fronts` / `sun` precedent (`None` ⇒ UNAVAILABLE, old packs degrade cleanly). The briefing run and the recalculate/preview re-run both build it, and the recalc/preview/alt API endpoints now hand `airports_db_path` down — the recalc path has no resolved `RouteConfig`, so `_arrival_icao` falls back to the recomputed airport conditions' arrival ICAO. Miss either point and the advisory silently degrades to a grey card on recalculate.

Params (`flight_rules` category, `default_enabled=True`, no self-gating on `aircraft.is_ifr` — matching `vfr_feasibility`/`ifr_feasibility`): `tailwind_limit_kt` (10 kt, pilot tier), `crosswind_limit_kt` (20 kt, pilot tier), `circling_ceiling_ft` (1000 ft, advanced), `circling_visibility_m` (1500 m, advanced).

**Surfacing.** The advisory itself needs no per-advisory registration — web/iOS render cards from the served catalog, and MCP `summarize_advisories` / `get_advisory_detail` plus the digest prompt builder are generic. But **four hand-maintained id→category tables do need an entry**, and all four already list both feasibility composites: `ADVISORY_PRIORITY` (`web/ts/helpers/advisory-order.ts`), `ADVISORY_TAG_MAP` (`debriefs/taxonomy.py` + its TS mirror; iOS reads it from the served catalog) → `IMC`, `ADVISORY_SITUATION` (`eval_workbench/situations.py`) → the existing `ifr_marginal` cell, and `_AIRPORT_CONDITION_ADVISORIES` (`eval_workbench/rerun.py`) so corpus re-runs flag it rather than reporting its UNAVAILABLE as an unexplained regression. Detail strings are localized en/fr/de/es. No cross-section highlights or chip — like every airport advisory this is point-in-space, not route geometry. No new API-serialized field, so no iOS DTO change.

### Deferred, and why

- **TAF preference at D-0.** The issue asks this to prefer TAF over NWP so the card cannot contradict the alternates card beside it. Advisories run at pipeline step 3 and METAR/TAF are fetched at step 3.5, so `RouteContext` has no observation to read — closing this means reordering the pipeline, well beyond this change. The NWP consensus in `airport_conditions.arrival` is what is graded today; recorded as a known gap in the evaluator docstring and in `designs/advisories.md`.
- Per-candidate approach feasibility for **alternates**, and an approach-aware "best usable runway" alongside `best_runway` — both listed as out-of-scope / follow-up in the issue.
- #510 (user-declared unpublished approaches) ships after this, as specified.

**Noticed, not fixed here (pre-existing):** `_AIRPORT_CONDITION_ADVISORIES` contains `"airport_conditions"`, which matches no registered advisory id — the airport category is `flight_category` / `airport_wind` / `density_altitude` / `llws` — so those first two are currently unflagged in every workbench diff. Flagged in a code comment; fixing it would silently change diff output for advisories this PR does not touch.

Design docs updated: a dedicated section in `designs/advisories.md` (grade map, resolution states, the honesty rules, the blocker × softening cross product, the wiring points, the known gap), evaluator counts in `designs/INDEX.md`, and the now-shipped follow-up struck from `designs/alternate-requirement.md`.

## Issue linkage

Closes #509

## Testing / verification

83 new tests across three files.

- **`tests/analysis/advisories/test_approach_feasibility.py`** (46) — UNAVAILABLE vs no-approach vs failed lookup; the VFR/MVFR/IFR gating incl. "MVFR ignores a misaligned approach"; ceiling and visibility band edges; the tailwind ladder; circling-only, unresolved-alignment and missing-wind-data never driving RED; the minima-vs-misalignment split (two served ends in *different* minima bands); the full blocker × softening matrix resolving in all four locales; circling needing visibility as well as ceiling; circling minima not lending straight-in minima to the gate; per-model split; and a real param-wiring test that moves each of the four params across the threshold it controls and asserts the verdict changes.
- **`tests/test_runway_approaches.py`** (29) — runs against the **real** euro_aip `Airport`/`Runway`/`Procedure` classes with only the DB layer mocked, so a euro_aip contract change fails here rather than silently degrading every destination to "no approach". Covers ident normalization (`RWY 09R` → `9R`; this caught a real bug in the first implementation), circling-vs-parser-artifact detection, closed runways, ends without a true heading, departures not counting as approaches, the EGKA shape from the issue, and a `TestFailureNeverImpersonatesAbsence` group pinning that an unknown ICAO and a raising procedure query both set `lookup_failed` while a genuine absence (EGTF Fairoaks) still grades.
- **`tests/test_advise_arrival_approaches.py`** (8) — both wiring points, incl. the recalc path's fallback to the airport-conditions arrival ICAO and lookup failure degrading to `None` rather than breaking a briefing.
- **Existing:** full backend suite — 4320 passed, with the failure set matching the pre-change baseline exactly (16 failures + 6 collection errors, all environmental in this sandbox: subprocess launching, thread-pool timing, OAuth redirect, ECMWF file paths). Web: `npm test` 556 passed across 35 files; `tsc --noEmit` shows only two pre-existing errors in `ts/eval/label-panel.ts`, untouched here.

Two limits worth a spot-check after deploy: the sandbox has no `nav.db`, so the euro_aip join is verified against the real model classes with a mocked DB layer rather than against production procedure rows (EGKA / EGBJ are the fields to eyeball); and `euro-aip` requires Python ≥3.12 while this container has 3.11, so it was installed with `--ignore-requires-python` (it is pure Python) — that affects only the local test environment, not the change.

### DIFF
--- a/designs/INDEX.md
+++ b/designs/INDEX.md
@@ -54,7 +54,7 @@ Comprehensive catalog of all ~85 weather metrics across 7 models: Open-Meteo API
 → Full doc: analysis-metrics.md
 
 ### advisories
-Route advisory system: 21 deterministic evaluators across 11 categories (icing incl. freezing precipitation, cloud, precipitation/en-route visibility, turbulence incl. wave-corroborated mountain wind, convective incl. convective-character, winds-aloft trip impact, airport conditions incl. density altitude, LLWS and terminal convective, feasibility, model-quality, fronts, sun) with per-model severity grading, user-tunable params, registry auto-discovery, worst/majority aggregation, and recalculation without re-fetching.
+Route advisory system: 22 deterministic evaluators across 11 categories (icing incl. freezing precipitation, cloud, precipitation/en-route visibility, turbulence incl. wave-corroborated mountain wind, convective incl. convective-character, winds-aloft trip impact, airport conditions incl. density altitude, LLWS and terminal convective, feasibility incl. approach feasibility, model-quality, fronts, sun) with per-model severity grading, user-tunable params, registry auto-discovery, worst/majority aggregation, and recalculation without re-fetching.
 Key exports: `evaluate_all`, `get_catalog`, `RouteContext`, `RouteAdvisoriesManifest`
 → Full doc: advisories.md
 
--- a/designs/advisories.md
+++ b/designs/advisories.md
@@ -263,6 +263,18 @@ non-precision minimum. Merging the two axes made the evaluator announce "wind
 favours RWY 20; approaches serve 02, **20**" — naming the served, wind-favoured
 runway as if it were unserved (caught in review on #511).
 
+**One discriminator drives both the grade and the copy** (`_Softening`, whose
+value doubles as the message-key suffix — `minima_{circling,uncertain,blocked}`
+× `misaligned_{…}`). Choosing them from different tests is a bug generator, and
+it produced two: an AMBER earned by an *unrelated* unresolved approach kept the
+hard "the ceiling will not support circling" claim, and — worse, in the branch
+review didn't flag — the misalignment case advised "plan for circling" at a
+ceiling the same call had just decided was below the circling minimum. Softening
+for uncertainty is correct, but it has to name that reason in its own words
+rather than borrow the circling advice. Note the fix is *not* to grade off
+`circling_supported` alone: that would harden genuine uncertainty into RED,
+which is precisely what rule 2 forbids.
+
 **Circling is flagged, never graded.** `nav.db` carries no circling minima, and
 "circling not authorised" / night / category restrictions are not in the data at
 all. `circling_ceiling_ft` (default 1000) only decides whether a misaligned
--- a/designs/advisories.md
+++ b/designs/advisories.md
@@ -176,6 +176,9 @@ states, and keeping them apart is the point:
 | `circling=True` | ICAO letter-suffixed circling-only procedure (`RNP A`, `NDB C`) |
 | neither | alignment unresolved — treated as *uncertainty*, never as hard misalignment |
 
+The container adds a fourth, airport-level state: `lookup_failed` (see
+UNAVAILABLE below).
+
 Circling detection keys on the single trailing letter, excluding X/Y/Z (those are
 straight-in *variant* designators) and any name carrying a digit. That digit
 guard is what matters: of the 262 ident-less rows in `nav.db`, only **39** are
@@ -201,15 +204,26 @@ no longer be expected to complete the arrival visually.
 | **MVFR** | IAP presence only, capped at AMBER. Any approach → GREEN; none → AMBER. **Never RED.** Alignment/circling/tailwind are not applied: with a 1000–3000 ft ceiling the landing can be completed visually, so a misaligned approach is not a penalty |
 | **IFR / LIFR** | Full logic. GREEN = straight-in to an end whose wind is within limits and a ceiling clear of the estimated minima band. AMBER = a compromise (circling required, ceiling *inside* the band, or a tailwind still within limits). RED = a hard fact only |
 
-**UNAVAILABLE** is reserved for "the evaluator could not run" — `arrival_approaches`
-absent (old pack / no nav.db) or no per-model airport condition. It is *not* used
-for an airport with no approaches: **zero procedure rows is treated as "no
-instrument approach available", full stop.** There is deliberately no third state
-distinguishing "genuinely has none" from "never ingested". Per-country coverage
-in `nav.db` is cleanly bimodal (40–100% or exactly 0.0%), and every 0% country is
-outside the current European scope, so the rule cannot misfire today.
-**Caveat for the US-expansion work:** in a 0%-coverage country this would grade
-every IFR destination RED. Revisit before enabling coverage there.
+**UNAVAILABLE** is reserved for "the evaluator could not run": `arrival_approaches`
+absent (old pack / no nav.db), no per-model airport condition, or
+`AirportApproaches.lookup_failed` — the ICAO was unknown to `nav.db`, or the
+procedure query raised. It is *not* used for an airport with no approaches:
+**zero procedure rows is treated as "no instrument approach available", full
+stop.** There is deliberately no third state distinguishing "genuinely has none"
+from "never ingested". Per-country coverage in `nav.db` is cleanly bimodal
+(40–100% or exactly 0.0%), and every 0% country is outside the current European
+scope, so the rule cannot misfire today. **Caveat for the US-expansion work:** in
+a 0%-coverage country this would grade every IFR destination RED. Revisit before
+enabling coverage there.
+
+`lookup_failed` exists because "this field has no instrument approach" is the
+**most severe input the grade map takes** (RED at IFR/LIFR), and it arrives as
+the same empty approach list a parse failure would produce. Only the flag tells
+them apart, so every path in `get_runway_approaches` that fails to *determine*
+the picture sets it rather than returning an empty list — a data problem must not
+be able to manufacture a RED. (The outer failure — the whole lookup raising —
+already degrades to `arrival_approaches=None` in
+`tasks/advise.py::_collect_arrival_approaches`; this closes the inner one.)
 
 **Two rules keep it honest.**
 
--- a/designs/advisories.md
+++ b/designs/advisories.md
@@ -237,8 +237,10 @@ already degrades to `arrival_approaches=None` in
    light. Emitting GREEN/AMBER/RED from the same data makes a soft claim look
    hard, so: **estimate uncertainty may push toward AMBER, never toward RED.**
    RED needs a hard fact — no IAP at all, a ceiling/visibility below even the
-   *best case* (`min(dh_lo)` / `min(vis_lo)`), or an approach-served end with an
-   out-of-limits tailwind *and* a ceiling that will not support circling. A
+   *best case* (`min(dh_lo)` / `min(vis_lo)`), an approach-served end with an
+   out-of-limits tailwind *and* a ceiling that will not support circling, or the
+   same best-case test applied **per approach** to an end the wind is happy with
+   (below). A
    forecast *inside* the band (`[dh_lo, dh_hi]`, `[vis_lo, vis_hi]`) is AMBER —
    the AMBER middle is exactly the width of our uncertainty about the published
    minima. Neither an approach of unresolved alignment nor a served end the wind
@@ -247,6 +249,19 @@ already degrades to `arrival_approaches=None` in
    absent visibility means the model does not publish one (ECMWF visibility is
    GRIB-only) and the axis is skipped, matching `flight_category`, which also
    does not read an absent visibility as a poor one.
+
+**Two failures reach the "no usable straight-in" fallback, and they are not the
+same story.** Either every approach-served end is out on wind (genuine
+misalignment — "wind favours 24; approaches serve 02"), or an end the wind is
+*happy* with is blocked only by its own approach's minima. The second is not
+misalignment at all, so it gets its own copy naming the runway and its approach
+class. This is why `_GradedPlan` keeps the wind and minima verdicts apart
+instead of merging them: the airport-wide best-case gate uses `min(dh_lo)` over
+*all* approaches, so a ceiling can clear it (via, say, an ILS on the
+wind-unusable end) and still sit below the wind-favoured end's own
+non-precision minimum. Merging the two axes made the evaluator announce "wind
+favours RWY 20; approaches serve 02, **20**" — naming the served, wind-favoured
+runway as if it were unserved (caught in review on #511).
 
 **Circling is flagged, never graded.** `nav.db` carries no circling minima, and
 "circling not authorised" / night / category restrictions are not in the data at
--- a/designs/advisories.md
+++ b/designs/advisories.md
@@ -4,7 +4,7 @@
 
 ## Intent
 
-Provide actionable, severity-graded (GREEN/AMBER/RED) advisories from 21 hazard evaluators (grouped into icing, cloud, precipitation, turbulence, convective, wind, model, airport, feasibility, fronts, and sun categories) along the route. Evaluators analyze existing route analysis data — no additional data fetch. User-tunable parameters allow recalculation without re-running the pipeline. This is a **route-level** system (advisory per route), complementing the per-waypoint `AltitudeAdvisories` in the sounding subpackage.
+Provide actionable, severity-graded (GREEN/AMBER/RED) advisories from 22 hazard evaluators (grouped into icing, cloud, precipitation, turbulence, convective, wind, model, airport, feasibility, fronts, and sun categories) along the route. Evaluators analyze existing route analysis data — no additional data fetch. User-tunable parameters allow recalculation without re-running the pipeline. This is a **route-level** system (advisory per route), complementing the per-waypoint `AltitudeAdvisories` in the sounding subpackage.
 
 ## Architecture
 
@@ -14,6 +14,7 @@ RouteContext (immutable)
   ├── cross_sections: list[RouteCrossSection]  (per-model forecast grids)
   ├── elevation: ElevationProfile | None
   ├── airport_conditions: AirportConditions | None  (dep + arr weather)
+  ├── arrival_approaches: AirportApproaches | None  (destination IAPs, #509)
   ├── sun: RouteSunAnalysis | None  ├── route_fronts: RouteFrontsManifest | None
   ├── cruise_speed_ias_kt, flight_duration_hours  (headwind trip-time inputs)
   ├── models, cruise_altitude_ft, flight_ceiling_ft, total_distance_nm, locale
@@ -38,6 +39,7 @@ Registry → evaluate_all(ctx, enabled_ids?, user_params?, aggregation?)
   ├── @register LLWSEvaluator
   ├── @register VFRFeasibilityEvaluator    # composite go/no-go
   ├── @register IFRFeasibilityEvaluator
+  ├── @register ApproachFeasibilityEvaluator  # arrival: approach vs wind vs ceiling
   ├── @register FrontsEvaluator            # fronts (experimental, gated on artifact)
   └── @register SunEvaluator               # sun (glare + night-proximity + seating note)
       ↓
@@ -85,7 +87,7 @@ Detail text comes from the worst-performing model. Shared classmethods on the mo
 
 `AdvisoryStatus.majority(statuses)` implements the majority logic: count each status (ignoring UNAVAILABLE), find max count, return worst among tied leaders. The registry re-aggregates after each evaluator returns if mode isn't WORST, so evaluator code is unchanged.
 
-## The 21 Evaluators
+## The 22 Evaluators
 
 ### Icing
 
@@ -140,6 +142,7 @@ Detail text comes from the worst-performing model. Shared classmethods on the mo
 |-----------|----------|-------|----------------|
 | `VFRFeasibilityEvaluator` | feasibility | Composite VFR go/no-go combining: airport flight category, en-route cloud clearance (base vs cruise), VMC compliance (BKN/OVC percentage), climb-out/descent corridor decks (BKN/OVC fully below cruise within `terminal_corridor_nm` of either end — OVC→RED, BKN→AMBER, see meteorology-decisions §10), and en-route precipitation (shared `classify_enroute_precip`, capped at AMBER in the composite). Worst of sub-assessments wins | `cloud_clearance_ft`, `imc_pct_amber`, `imc_pct_red`, `terminal_corridor_nm` (5) |
 | `IFRFeasibilityEvaluator` | feasibility | Composite IFR go/no-go combining: airport IFR viability (LIFR→amber, below minimums→red), en-route icing exposure (uses shared `icing_zones_in_altitude_range()` helper over `[0, cruise + icing_altitude_buffer_ft]`, aligned with FIKI advisory), convective risk along route | `min_dep_ceiling_ft`, `min_arr_ceiling_ft`, `icing_pct_amber`, `icing_pct_red`, `icing_altitude_buffer_ft` |
+| `ApproachFeasibilityEvaluator` | flight_rules | **The arrival question no other evaluator owns (#509):** *can I get in, on a runway I can also land on?* Joins the destination's ceiling, the runway the wind favours, and which runways actually have a published approach — a joint function of ceiling **and** wind **and** infrastructure. See the dedicated section below | `tailwind_limit_kt` (10), `crosswind_limit_kt` (20), `circling_ceiling_ft` (1000) |
 
 ### Sun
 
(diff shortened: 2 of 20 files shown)


roznet:
Confirmed and fixed in `a5fc65f`.

Reproduced it exactly as described (with `VOR` on the second end — the DH range quoted for RNP in the review is the non-precision bucket; RNP is 250–600. The defect holds for any two classes whose bands differ):

```
02: ILS (best case 200 ft), 15 kt tailwind  -> RED on wind
20: VOR (best case 400 ft), 15 kt headwind  -> RED on minima  (wind-favoured)
ceiling 250 ft, vis 4000 m
```

Old output — self-contradictory, since 20 is in its own served list:

> `Arr EGKA: IFR — Wind favours RWY 20; approaches serve 02, 20, and ceiling 250ft will not support circling`

New output:

> `Arr EGKA: IFR — RWY 20 is clear on wind, but the forecast is below typical VOR minima (~400ft), and the ceiling will not support circling`

The root cause was merging the wind and minima verdicts into one per-candidate status before reaching the fallback, which threw away the only information that distinguishes the two stories. `_GradedPlan` now carries both axes and the fallback branches on cause: a wind-acceptable end blocked solely by its own approach's minima gets copy naming that runway and its approach class; the genuine "every served end is out on wind" case keeps the misalignment copy.

One correction to the framing: this isn't an unlabelled fourth RED path. It's rule 2 — *below the best case* — applied **per approach** rather than airport-wide. The airport-wide gate uses `min(dh_lo)` over all approaches, so a ceiling can clear it via a low-minima approach sitting on an end the wind has already ruled out, while still being below the wind-favoured end's own best case. That's still a hard fact about that end, so RED is right; only the copy was wrong. The three documented causes now read as four in `designs/advisories.md` with the per-approach case spelled out, and the same uncertainty softening (`circling_supported` / `alignment_unresolved` / `wind_unresolved` → AMBER) applies unchanged.

Test coverage for the gap you identified — you were right that `test_best_usable_end_wins_over_a_worse_served_end` used two identical ILS proxies and never exercised differing bands:

- `test_minima_blocked_served_end_is_not_reported_as_misaligned` — the repro above; asserts RED, that the copy names RWY 20 and VOR, and that the words "favours"/"serve" do **not** appear.
- `test_minima_blocked_softens_to_amber_with_circling_room` — same shape with a 1200 ft ceiling and 1000 m visibility (visibility drives the block), asserting the AMBER + circling wording.
- `test_all_ends_out_on_wind_still_reads_as_misalignment` — guards that the genuine misalignment case keeps its own copy.

Full backend suite: 4308 passed, with the same 16 failures and 6 collection errors as the pre-change baseline (all environmental in this sandbox — subprocess launching, thread-pool timing, OAuth redirect, ECMWF file paths).

---

roznet:
Confirmed and fixed in `0466a30` — and the trace turned up the same defect in the sibling branch, with worse consequences.

**The flagged case**, reproduced as described (ILS on 02 with an out-of-limits tailwind, VOR on the wind-favoured 20, ceiling 250 ft, plus one unresolved-alignment approach):

> AMBER | `RWY 20 is clear on wind, but the forecast is below typical VOR minima (~400ft), and the ceiling will not support circling`

**The case review didn't flag** — same split, in the misalignment branch:

> AMBER | `Wind favours RWY 24; approaches serve 02 — plan for circling`

…at a 600 ft ceiling, when the same call had just evaluated `circling_supported` as false against the 1000 ft `circling_ceiling_ft`. That one isn't an unexplained badge, it's advice the evaluator's own arithmetic contradicts.

**On the suggested fix**: I took the second option, not the first. Driving the grade off `circling_supported` alone would harden genuine uncertainty into RED — an unresolved approach may well serve the wind-favoured end with lower minima, which is exactly the "uncertainty may only soften" rule (design rule 2) that the earlier `lookup_failed` commit on this branch exists to protect. The softening is correct; only the copy was wrong.

Both branches now derive grade and copy from a single `_Softening` discriminator whose value doubles as the message-key suffix (`minima_{circling,uncertain,blocked}` × `misaligned_{…}`), so they cannot be chosen from different tests again. All six outcomes:

| ceiling supports circling | uncertainty present | grade | copy |
|---|---|---|---|
| yes | — | AMBER | "plan for circling" |
| no | yes | AMBER | "an approach here could not be matched to a runway — check the plates" |
| no | no | RED | "the ceiling will not support circling" |

New test `test_uncertainty_softening_never_borrows_the_circling_advice` pins both branches, asserting the word "circling" is absent from the misalignment-uncertain copy and that the hard claim is absent from the minima-uncertain copy.

**On the minor** — you're right, and it was worse than a docstring mismatch: `test_params_are_honoured_on_recalc` promised to check that every declared param reaches the grading code while asserting only `min <= default <= max`, so a param dropped in `_grade_full` would have been silently inert on the settings page with a green test. Split into an honest bounds check (`test_param_defaults_sit_inside_their_declared_bounds`) plus `test_every_declared_param_actually_reaches_the_grading_code`, which moves each of the three params across the threshold it controls and asserts the verdict changes — including checking that the tailwind param flips the *copy* to the circling branch, not just the badge.

Full backend suite: 4310 passed, same 16 failures + 6 collection errors as the pre-change baseline (all environmental in this sandbox).

---

roznet:
All four confirmed and fixed in `26e8cad` — but this is the fourth round of findings in the same ~40 lines, so it replaces the shape rather than patching another cell.

**The pattern.** That block's verdict is a cross product — *what* blocks the arrival × *why* it isn't RED — and the cells were written out by hand. Each round found a different cell rendering the wrong sentence: round 2 the minima cell, round 3 the uncertainty cells, round 4 the fog and no-wind-data cells. Patching cell-by-cell was never going to converge. It's now two enums with the copy composed as `{blocked_<blocker>} — {soften_<softening>}`:

| `_Blocker` | `_Softening` |
|---|---|
| `MINIMA` — wind-acceptable end blocked by its own approach's minima | `CIRCLING` — ceiling **and** visibility support it |
| `MISALIGNED` — every served end is out on wind | `UNRESOLVED` — an approach we couldn't match to a runway |
| `NO_STRAIGHT_IN` — nothing straight-in published at all | `NO_WIND_DATA` — no wind components for the served end |
| | `NONE` — the only RED |

Every cell now has correct text by construction, and a new blocker or reason costs one string instead of a row or column. `test_every_cell_resolves_to_real_text` asserts all cells resolve in en/fr/de/es (no silent fall-through to a raw key), and `test_softening_reasons_do_not_share_wording` asserts no non-circling reason recommends circling.

**Finding 1 (visibility)** — right, and it was reachable in a specific band: above the straight-in floor (`min(vis_lo)`, 550 m with an ILS) but below anything that supports circling. Added `circling_visibility_m` (default 1500 m). Before/after with no ceiling and 900 m visibility:

> ~~AMBER | `Wind favours RWY 24; approaches serve 02 — plan for circling`~~
> RED | `Wind favours RWY 24; approaches serve 02 — conditions will not support circling`

Absent values stay permissive on both terms, so a model that publishes no visibility never costs the pilot a grade (`test_absent_visibility_never_hardens_the_grade`).

**Finding 2 (`NO_WIND_DATA` borrowing the unresolved copy)** — right, and fairly called: that's the same bug class I claimed to have closed one commit earlier. I unified grade and copy, then mapped two causes onto one string. Now its own reason: *"no wind components for the approach-served runway, so this could not be checked"*. You were also right that `test_wind_components_missing_never_drives_red` asserted only the status — it now asserts the text too.

**Finding 3 (the two registries)** — both wired. `ADVISORY_TAG_MAP` → `IMC` (with the web mirror; iOS reads `advisory_tag_map` from the served catalog, so it propagates). `ADVISORY_SITUATION` → the existing `ifr_marginal` cell rather than a new `SITUATION_VOCAB` entry: this only grades at an IFR/LIFR destination, and a new vocab cell would change the coverage-matrix denominator every corpus fixture is scored against (#252/#254) — a bigger call than this PR should make unilaterally.

**Finding 4 (`approaches serve -`)** — fixed by `NO_STRAIGHT_IN` having its own copy: *"No straight-in approach is published here; wind favours RWY 24 — plan for circling"*.

One existing test was rewritten rather than kept green: `test_minima_blocked_softens_to_amber_with_circling_room` built its case from 1000 m visibility, which the new visibility term makes self-contradictory (you cannot circle in 1000 m). It now lowers the circling assumption instead — which is what that param is for.

Suites: backend 4317 passed; the failure set matches the pre-change baseline plus `test_grib_pool::test_two_concurrent_submits_run_in_parallel`, a wall-clock concurrency assertion that flakes under load and passes 3/3 in isolation. Web 556 passed.

---

roznet:
Both confirmed and fixed in `612afea`.

**The eval-workbench flag** — right, and the consequence is the bad one: `rerun.py` never passes `airports_db_path` to `run_advisories_from_pack`, so `_collect_arrival_approaches` returns `None` and every corpus re-run grades `approach_feasibility` UNAVAILABLE. Without the flag that flip lands in the diff as an *unflagged* change, sitting next to genuine regressions — precisely what the set exists to prevent. Added.

While checking it I found a **pre-existing bug in that same set**, which I've noted in the code but deliberately not fixed:

```python
_AIRPORT_CONDITION_ADVISORIES = {"airport_conditions", "density_altitude", "llws"}
```

`"airport_conditions"` matches no registered advisory id — the airport category is `flight_category` / `airport_wind` / `density_altitude` / `llws` — so `flight_category` and `airport_wind` are currently **unflagged** in every diff, and have been. Fixing it changes diff output for advisories this PR doesn't touch, so it's a one-line follow-up for whoever owns the workbench rather than something to bundle here. Happy to open an issue if useful.

**The circling-minima blend** — agreed, and I'd put it slightly more strongly than "worth documenting". `APPROACH_CLASS_PROXY` holds straight-in minima; a circling ILS claiming a 200 ft best case is an optimistic assumption sitting *inside a RED gate*, which is design rule 3 leaking into exactly the place rule 2 is supposed to be strictest. Your trace matches mine — no verdict changes today, because the fallback re-derives the RED independently — but "harmless until someone edits the fallback" is what the next round would have inherited. Circling procedures are now excluded, and a field whose every approach is circling-only skips the gate rather than inventing a minimum it has no data for (the `None` return is live for that shape, not defensive — worth noting since an earlier round flagged a genuinely-dead `None` on this same function).

Two tests: `test_a_circling_ils_does_not_lower_the_best_case_gate` (asserts the gate fires on the VOR's 400 ft, not the circling ILS's 200 ft) and `test_a_circling_only_field_skips_the_gate_rather_than_inventing_one`.

Suites: backend 4319 passed — failure set now matches the pre-change baseline **exactly** (16 failures + 6 collection errors, all environmental in this sandbox; the `test_grib_pool` flake from the last round did not recur). Web 556 passed.

Worth recording for whoever merges: this is the first round with **no findings in the no-usable-straight-in fallback**, which rounds 2–4 all hit. The blocker × softening restructure in `26e8cad` looks to have closed that class rather than moved it.

---

PR #514, opened 2026-07-29
feat(data-sources): timeline view with expected-vs-actual arrival markers

roznet:
## What

Adds a **Table | Timeline** switch to the Data Sources tab, defaulting to the timeline. The existing table stays available and unchanged; the compact table in the guide tab is untouched.

The timeline shows the last 24 h and next 12 h in UTC. Each model run draws from its init to `init + delivery_offset` (hollow ring = the expectation we hold), with the marker store's observed `published_at` as a filled dot and the gap between them drawn as a length plus a labelled delta.

`GET /api/data-sources` already returned everything needed — `cycles`, `horizon_hours`, `delivery_offset_hours`, `latest_init`, `published_at`. This is a second renderer over the same payload, not a new data path, and both views share one `loadCatalog()` result so switching costs no request.

## Why

Expected-vs-actual is the part that earns its keep. It turns offset drift into a standing signal instead of something you find by hand. On the day of writing it showed Open-Meteo ICON arriving ~60 min *early* and Météo-France ARPEGE 46 min *late* against their configured offsets — both invisible in the table.

It also surfaces two things the table structurally cannot: the per-cycle horizon *shape* (short cycles draw thinner, so ECMWF's 168/144 alternation and ICON-EU's 120/78 nowcast rhythm read at a glance), and arrival times against the hours people actually plan in, via the clock rows.

## Included bug fix

First commit is independent and worth reviewing separately: the registry advertised a **90 h** horizon for ECMWF's short cut-off 06/18Z cycles. The delivery manifest says **144 h** and the files on disk agree. The 90 h came from reasoning off where the *hourly* step cadence ends — a cadence boundary, not a horizon — which under-reported 06/18Z coverage by 54 hours everywhere `horizon_for(6)` is consumed, including the user-facing "covers until" column and the source description in the help popup.

`max_horizon` stays 168 h, so 06/18Z still correctly fail to qualify as full-horizon runs and the scheduler's big-run deferral keeps waiting only on 00/12Z. Two tests carried the same assumption and were updated.

## Implementation notes

- **Two lanes.** Delivery offsets are routinely longer than the cycle interval (ECMWF waits 6h40m for a run every 6h), so consecutive runs overlap in time. On one line they fuse into a single continuous band and the per-run rhythm — the whole point of the row — disappears. Alternating lanes keeps every run's start and end readable.
- **DST by construction.** Every local-time conversion goes through `Intl.DateTimeFormat` per instant; no offset is hardcoded. The viewer's own zone leads the clock rows, skipped when its *current offset* matches a reference zone — matched by offset rather than zone id, so Europe/Zurich doesn't draw a second Europe/Paris.
- **Colour encodes role, not model.** Identity is already in the row label; the four roles map onto a palette checked for colour-vision separation against both the light and dark surfaces. Lateness keeps a separate reserved status colour so a late primary source never reads as a different role.
- Legend labels reuse the table's existing `dataSources.role.*` keys — one label set, not two.

## Testing

- **13 new unit tests** covering run enumeration, short-run classification, realised-marker matching (including the GEM case where an init is known but `published_at` is null), and both the European and US DST transitions — which fall on different dates, exactly the case a hardcoded offset gets wrong.
- Full web unit suite passes (569 tests). The 8 file-level failures from `npx vitest run` are pre-existing Playwright specs vitest can't collect.
- 188 Python tests pass across the freshness / catalog / scheduler modules.
- Rendered under Playwright against the live production catalog in **both themes**: tooltip, toggle round-trip, viewer-zone row all confirmed; no page errors; no horizontal page overflow at 390 px (the strip scrolls inside its own container).

## Notes for review

- English-only i18n, matching the existing `dataSources.*` keys — fr/de/es already carry none of them.
- Not included, deliberately: the marker store keeps only the *latest* marker per source, so there is one realised dot per row. A small rolling delivery log would let the timeline draw an observed-lag band and is the natural follow-up.


roznet:
Thanks — both **Important** findings were real and are fixed in `ec359d0b`, along with both **Minor** ones (pushing for the blockers anyway, so the round was already paid for).

## Fixed

**Model grouping** (`data-sources-timeline.ts`) — confirmed against the live catalog order: `gfs` at indices 1 and 4, `ecmwf` at 0 and 5, so variants were never contiguous and the heavy `ds-tl-group` rule marked the wrong boundaries. Rather than add local grouping, `groupByModel` and `orderedModelKeys` are now exported from `data-sources-table.ts` and reused through a new `planRows` — the review was right that the table already solves this and the timeline was carrying a weaker duplicate. Both views now group *and* order families identically.

Side benefit: ECMWF-direct and ECMWF-openmeteo now sit adjacent, which makes their differing cadence (4 cycles vs 2) directly comparable.

**Gridline phase alignment** — confirmed by calculation and in the browser. Anchored to an absolute 3-hourly boundary via a new `alignUp`, as suggested. **The clock rows' hour readouts had the identical defect** (same `Math.ceil(t0 / HOUR_MS)` start, stepped 3h) and were not flagged; fixed with the same anchor so ruler, gridlines and local-hour columns share one set of columns at any load hour.

**Band overscan** — `flush` now clamps to `scale.t1`, so the one-hour overscan that closes a still-open band can't push its right edge past 100%.

**`designs/multi-user-deployment.md:196`** — updated; that was the last stale 90h reference in the tree (re-grepped `designs/` and `src/` to confirm).

## On the "eyeballing wouldn't catch this" point

Correct, and worth recording why: the screenshots I verified against were taken at **12:07Z** — one of the 1-in-3 hours where the grid happens to align. Both bugs are now covered by **7 regression tests**: variant contiguity, exactly one family opener per model, no rows dropped by grouping, role ordering within a family, and grid anchoring at both aligned and unaligned load hours.

Re-verified in the browser with a fixed clock at **13:23Z** (previously broken) and 12:23Z: every hour label has a matching gridline, synoptic lines present in both, zero bands overflowing, no page errors. 576 web unit tests pass; the Python side was untouched this round (188 green on the previous one).

## Deferred

Nothing.

PR #516, opened 2026-07-30
fix(advisories): grade convection once, consume it everywhere

roznet:
## The bug

On [`egtf_…_lfmd-2026-08-01` (b460)](https://weather.flyfun.aero/briefing.html?flight=egtf_gwc_sitet_kovak_domod_trare_mtl_abdil_lfmd-2026-08-01-b460) the briefing showed **IFR Feasibility RED** — "HIGH convective over 95nm/600nm (16%)" — directly beside **Convective Activity AMBER** — "MODERATE+ over 18–24% across models, peak MODERATE". Same route, same soundings, same run.

`ifr_feasibility` carried a second, older derivation of "how bad is the convection?":

| | `convective` (after #442) | `ifr_feasibility` (before) |
|---|---|---|
| risk floor | `min_risk` 2 (LOW) | `convective_min_risk` 3 (MODERATE) |
| DD-trigger amber (§18) | yes | **no** — read the raw active track |
| tops-below-cruise filter | `top_clearance_ft` 2000 | **none**, despite `altitude_dependent=True` |
| → RED | NWP HIGH, or MOD+ ≥ 50 % | HIGH anywhere, or ≥ **10 %** |

#442 reworked the *convective advisory* to grade NWP-native with a DD-trigger AMBER cap; it never touched the other consumers of the convective track. Two formulas over one dataset diverged in both directions:

| model | `convective` | `ifr_feasibility` | why |
|---|---|---|---|
| ECMWF | AMBER (MOD+ 17 %) | **GREEN** | dd_trigger amber invisible to IFR |
| GFS | AMBER (MOD+ 24 %) | **GREEN** | ditto |
| ICON | AMBER (MOD+ 19 %) | AMBER (2 %) | IFR saw only the genuinely-NWP-MODERATE point |
| UKMO | RED (peak HIGH) | RED | no NWP scheme → both fall back to thermo |
| MétéoFrance | RED (peak HIGH) | RED | ditto |

Aggregation is MAJORITY with ties broken to worst, so this **inverted the headline**: convective read amber 3 / red 2 → AMBER; IFR read green 2 / amber 1 / red 2 → a 2–2 tie → RED. The digest then narrated a *"RED IFR Feasibility advisory driven by HIGH convective exposure"* two paragraphs after stating the models' *"own convective schemes continue to show zero convective precipitation, consistent with the cap currently suppressing initiation"* — the loaded-gun false-alarm RED that §18 exists to prevent, re-entering through a composite that had not been told.

## The fix

**One formula.** New `convective_grading.grade_convective_model` holds the #442 grading verbatim. `ConvectiveEvaluator` becomes a thin wrapper over it (locale wording + cross-model aggregate only); `IFRFeasibilityEvaluator` takes its status **verbatim** as the convective axis. Identical by construction, not by calibration — porting the rules into the composite would have re-synchronised two copies that drift again on the next change, which is exactly what happened between §14 and §18.

**Tuning follows.** `evaluate_all` publishes the override map on `RouteContext.advisory_params` (via `dataclasses.replace` on the frozen dataclass — one copy per call, no call-site or evaluator changes). The composite resolves the *convective advisory's* parameters through `resolve_convective_params(ctx)`, falling back to catalog defaults when that advisory is disabled. Raising `affected_pct_red` now moves both cards with one edit; before it moved neither. `convective_min_risk` / `convective_pct_red` are retired.

**Character couples in through EMBEDDED only.** ISOLATED/SCATTERED mean "circumnavigable with see-and-avoid" — a VFR statement that does not transfer to IMC. WIDESPREAD/ORGANIZED are realized-coverage bands and `convective` already reds on MODERATE+ coverage, so escalating on them would price one measurement twice. EMBEDDED is orthogonal — it is about the *deck hiding the cells*, not the convection — and is the one band genuinely worse under IFR than VFR. AMBER→RED only; a GREEN convective axis has no cells for a deck to hide. The escalated red names itself (`ifr.conv_embedded`).

**Prerequisite fixed.** `classify_convective_character` returned ISOLATED (→AMBER) whenever *any* point was `is_convective`, because `realized_pct = 0` falls into the `<= isolated_max_pct` bucket — rendering *"Isolated cells — circumnavigable VFR with see-and-avoid, expect possible diversions **(0nm/600nm (0%))**"* on UKMO and MétéoFrance (no model-native convective scheme, so nothing ever realizes), beside a RED "peak HIGH" on the same model. Zero realized cells now returns NONE (→GREEN), checked *before* the EMBEDDED test — "cells hidden under a deck" presupposes cells.

## Invariant

For every model, IFR's convective axis **is** the convective advisory's status. Exactly two deviations are sanctioned and tested:

1. **Abstention** — the #391 thin-coverage guard may turn a would-be-GREEN axis UNAVAILABLE (contributes nothing to `worst`, never a different grade).
2. **The EMBEDDED escalation**, above.

## Effect on b460

Inferred from the observed per-model statuses run through the real aggregator — the pack isn't available locally, so this is not a replay:

- `ifr_feasibility` **RED → AMBER** (no model is EMBEDDED there, so it tracks convective exactly)
- `convective_character` **AMBER → GREEN** (UKMO/MF had zero realized cells; ICON's real 2 % cell keeps it AMBER)
- `convective` unchanged at AMBER

## Testing

Full suite **4502 passed / 9 skipped**. The pre-existing #442 convective suite and the character classifier tests pass unchanged — the evidence the extractions are behaviour-preserving. New `test_convective_consistency.py`: the invariant parametrized across dd_trigger / NWP-cell / HIGH / tops-below-cruise / quiet, the specific regression, shared tuning, the EMBEDDED escalation firing end-to-end, the bands that must not escalate, GREEN staying GREEN, and the detail naming itself. Plus three zero-realized classifier cases in `test_convective.py`.

## Docs

`meteorology-decisions` **§22** (decision, reasoning, what it supersedes); `advisories.md` IFR + character rows and the module tree.

## Not addressed

A model with **no convective scheme at all** (UKMO, MétéoFrance) falls back to thermo and can therefore produce a RED off DD alone — which is what drove both reds on this flight. §18 promises RED comes only from the NWP track; the scheme-less fallback is a gap in that promise. Recorded in §22 as a related open item — it deserves its own decision since it would change grades on every non-GRIB model.


roznet:
## Review round 1 — addressed

Both findings fixed in `ddb5b298`.

**Important — `advisory_params` unfiltered by `enabled_ids`: fixed.** Confirmed the premise on both halves: `_load_advisory_profile` reads `adv_config["enabled"]` and `adv_config["params"]` independently (`packs.py:3386/3390`), and `renderParamInput` emits the `disabled` attribute for a switched-off advisory's inputs (`settings-main.ts:1296`) — so the UI actively presents those values as inert while the backend would have graded against them. Agreed this is a blocker rather than a docstring inaccuracy: it's a regression against pre-PR behaviour (IFR's convective params came only from `user_params["ifr_feasibility"]`, always live when that row is on), and it is the §22 divergence rebuilt one level up — between what the pilot sees and what the pilot configured.

`evaluate_all` now publishes only the params of advisories actually in play, filtered through the same `_is_enabled` helper the evaluate loop uses so the two rules cannot drift. That is also what makes the `resolve_*` helpers' documented "disabled ⇒ catalog defaults" contract true rather than aspirational — no docstring change needed, the code now does what it always claimed.

Two regression tests: with `convective` disabled, a saved `min_risk: 4` no longer greens out IFR (RED via catalog defaults); with it enabled, a tuned `affected_pct_red: 95` still moves *both* cards from RED to AMBER together, so the filter doesn't over-reach.

**Minor — `graded_risk_by_point`: dropped.** Correct, and a sharper catch than it looks: the character work it was staged for landed in this same PR and `build_character_points` re-derives `is_conv` from the raw track instead, so it was born dead. Removed the field, its population, and the §22 reference to it.

meteorology-decisions §22 now records the enabled-gating rule and its rationale in the decision itself, not only in code comments.

Full suite: 4504 passed / 9 skipped.

roznet:
## Review round 2 — clean, no push

No correctness, architecture or meteorological issues. Both remaining findings are efficiency/tidiness, so I'm **not** pushing again — another push costs a full review round for no behavioural gain. Recommending merge with these folded into a follow-up commit on `main`.

**1. `conv_count` duplicates `conv_grade.affected` (`ifr_feasibility.py:184`)** — confirmed empirically, not just by reading: across all eight grading scenarios (dd_trigger dense/sparse, NWP MODERATE, HIGH dense/sparse, tops-below-cruise, quiet, embedded) the non-`None` `region_cells` count equals `grade.affected` exactly. They're written in the same block after every filter, so the identity holds by construction. Worth dropping the second walk.

**2. `build_character_points` runs twice per model when both advisories are enabled (`ifr_feasibility.py:299`)** — real, and I'd take this one more cautiously. Sharing the result needs a cache keyed on `(ctx, model, params)`, and `RouteContext` is frozen but holds lists, so it isn't hashable; keying on `id(ctx)` is fragile, and threading a computed-once map through `evaluate_all` re-introduces exactly the cross-evaluator coupling §22 just finished removing. At ~63 points × 5 models the cost is negligible. I'd rather leave the duplicate walk than buy a caching mechanism that costs more in coupling than it saves in time — but happy to revisit if profiling ever says otherwise.

CI green (GitGuardian + claude-review), 0 commits behind `main`, CLEAN/MERGEABLE.

PR #518, opened 2026-08-01
HRRR full-sounding upgrade of the gfs slot on CONUS routes (#457)

downle:
Independent implementation of **#457**, built from the pre-#508 base (`42e24645`) without reference to the merged #508 code.

**Status note:** #508 has already merged an implementation of this issue, so this branch conflicts with `main` in the same feature area — offered as an **independent alternative implementation for review/comparison**, not intended to merge as-is. The Files-changed view (three-dot diff from the merge-base) shows exactly this implementation. Built via the superpowers workflow: approved design (`docs/superpowers/specs/`), plan (`docs/superpowers/plans/`), subagent-driven per-task reviews, and a final whole-branch review.

## What it does

When the whole route fits the HRRR CONUS Lambert grid **and** a covering run exists, the `gfs` slot's GRIB enrichment is sourced from **HRRR** (3 km, convection-allowing, radar-assimilating) as a **full sounding replacement** (the ICON/ECMWF pattern, not today's GFS patch); otherwise plain GFS exactly as today. All-or-nothing gate, `WB_HRRR_ENABLED=0` kill switch, whole-slot fallback to plain GFS on total failure, `GFS (HRRR)` badge, `grib_sources["gfs"] = hrrr:noaa` attribution.

- **Fetch (`hrrr_fetch.py`)** — S3 URLs (`hrrr.{YYYYMMDD}/conus/hrrr.t{HH}z.wrfprsf{FF}.grib2[.idx]`, flat layout, 2-digit fhour — verified live 2026-07-31); run selection across hourly cycles (00/06/12/18z → 48h, others → 18h, ~1h publish delay, probes the **last-needed** fhour since publication is progressive); **exact Lambert domain gate** (pyproj projection of route points onto the 1799×1059 grid's x/y axes — no lat/lon bbox approximation); `gfs_idx.py` parametrized on variable sets (GFS defaults byte-identical).
- **Lambert decode (`decode.py`)** — branch in the shared interpolation helpers: datasets with `(y, x)` dims + 2D lat/lon get route points projected to fractional grid indices, then the same vectorised bilinear gather (no scipy griddata). **Wind rotation** grid→earth (`uvRelativeToGrid=1`, verified): α = (λ−LoV)·sin 38.5°, sign validated against a pyproj finite-difference oracle (the initially-drafted sign was caught and corrected by that test). `CIMIXR` decodes as cfgrib `unknown` (NCEP-local) → aliased (safe within the planned sounding blob); `DPT`/`HGT` direct (no Magnus/hypsometric); `VVEL` already Pa/s.
- **Enrichment** — ECMWF-shaped replacement: per-fhour idx → sounding byte-ranges (TMP/DPT/RH/UGRD/VGRD/VVEL/HGT/CLMR/CIMIXR on 40 levels + PRES surface, ~190 MB/fhour envelope) → pool decode → `_replace_pressure_levels_from_grib`; then diagnostics (band covers + ceiling + overall base, ML CAPE/CIN — measured **already negative** → passthrough) and gap-fill surface extras (VIS/GUST/sfc CAPE/CIN). `_enrich_gfs` returns `(ts, source_key)`; `gfs_init_dt` is set only for `gfs:noaa`, so **no GFS averaged-window machinery ever runs on HRRR hours** (all fields instantaneous).
- **Fallback/containment (three review rounds)** — total failure (no idx / zero decoded hours / crash) → whole-slot plain GFS; a crash *after* partial HRRR replacement keeps the partial HRRR pack honestly attributed (never a half-HRRR pack mis-labeled `gfs:noaa`); a diag-pass crash is contained without triggering the sounding fallback.
- **Wiring** — `hrrr` cache dir (6h TTL), `hrrr:noaa` freshness SourceConfig (extended cycles, 48h horizon), badge in the freshness bar + popover (mirroring the `ICON (D2)` mechanic), pyproj promoted to a declared dependency.
- **Docs** — `designs/weather-engine-specs.md` HRRR section + attribution matrices, `designs/fetch.md` fetch section.

## Verification

- **Golden decode on a real byte-ranged wrfprs sample** (`tests/test_hrrr_golden.py`, gitignored, skip-if-absent): projected interpolation sub-kelvin vs nearest-neighbour (|Δ| = 0.078 K at KDEN); projection anchored to the message's own 2D lat/lon coordinate arrays and a forward pyproj check (mutation-tested: wrong radius/sign flips FAIL the tests).
- **~150 HRRR tests** (gate/fallback branching, run selection, Lambert math, rotation closed form, builder, replacement semantics, crash containment, fill, kill switch, registry, badge vitest).
- **Full suite: 4615 passed**, 1 pre-existing failure (`test_google_login_redirects`, environment-dependent, fails on the base commit too — verified).

## Review-triaged follow-ups (recorded, none merge-blocking)

Run-selection garbage-probe on a pathological `last_needed` (→ 404 → correct behavior anyway); domain-gate axes rebuilt per call (could cache); CIN passthrough unguarded if a future cycle ever flips sign (measured 2026-07-31); degraded-path partial writes after a mid-run crash (documented semantics); GNU-only `date`/`stat` in the sample-download script.

🤖 Generated with [Kimi Code](https://www.kimi.com/code) via the superpowers subagent-driven workflow — co-authored on the commits via `Co-authored-by`.


roznet:
## Review — comparison against the merged #508

Thanks for this, and for the honest framing in the description (built from `42e24645` without reference to #508, offered for comparison rather than merge). I've read the whole branch against our merged implementation. Closing it as not-for-merge, but recording the analysis here because there are ideas in it we're going to take.

Short version: **the fetch/decode half is genuinely good** — it independently arrives at the same architecture we did, and in three places it's cleaner than what we merged. What it doesn't do is cross the boundary into the *consumers* of an HRRR-shaped diagnostic, and that's where the blocking findings are.

The structural tell was visible before reading any code: our change touched 9 downstream files (`clouds.py`, `convective.py`, `advisories.py`, `dd_nwp_agreement.py`, `spatial_interpolation.py`, `models/analysis.py`, `fill.py`, `advise.py`, `scheduler.py`). This branch touches none of them.

---

### What this branch got right

Independently reproduced, essentially identically to ours: the 2-digit `wrfprsf{FF}` flat layout, probing the **last-needed** fhour rather than f000 (progressive publication), per-cycle horizons for the hourly cycles, an **exact Lambert projection gate** rather than a lat/lon bounding box, the `CIMIXR`-decodes-as-`unknown` aliasing, direct DPT/HGT/VVEL with no Magnus / hypsometric / −ρgw derivations, full-sounding replacement via the ECMWF flow, and `_enrich_gfs` returning `(ts, source_key)`.

Two that deserve specific credit because they're easy to miss:

- **`gfs_init_dt` is set only for `gfs:noaa`**, so the GFS averaged-window machinery never runs over instantaneous HRRR fields. Same conclusion we reached, arrived at independently.
- **A `WB_HRRR_ENABLED` kill switch was in from the start.** We only added ours in a follow-up commit after the fact.

### Three things this branch does *better* than #508

1. **`modelSlotLabel()` extracted to `web/ts/utils.ts`** with a `{source → badge}` lookup table and a vitest unit test. Ours is a local `sourcedModelLabel()` helper inside `briefing-ui.ts` with an if-chain and no test. Yours is the right shape for when a third variant lands.
2. **The golden decode test anchors to independent oracles** — the message's own 2-D lat/lon coordinate arrays plus a forward pyproj check, mutation-tested so a wrong sphere radius or a flipped sign actually fails. That test caught your own initially-wrong wind-rotation sign, which is exactly what a golden test is for. Ours checks sub-kelvin against a nearest-neighbour readback, which is a weaker oracle.
3. **`scripts/download_hrrr_samples.sh`.** Our golden sample is gitignored and skip-if-absent with no reproducible way to regenerate it — a documented docstring and good intentions. A script is strictly better.

**We will pick up all three**, adapting (2) and (3) to our decode entry points and fixing the GNU-only `date`/`stat` usage you already flagged so it runs on macOS.

Not picked up, for the record: the `"mb"` level wildcard in the idx planner is neat, but our planner parametrises on `variables` + `min_level_hpa` and already matches pressure levels generically (we also skip 50–125 hPa, which you fetch).

---

### Blocking findings — all downstream, all invisible from the fetch layer

`build_hrrr_cloud_diagnostics` here produces the same shape ours does: band covers, ceiling, one overall base mirrored onto `low.base_ft`, ML CAPE/CIN, **no per-band top and no 3-D cloud fraction**. Three consumers break on that shape.

**1. HRRR ends up with no native cloud envelope at all.** `_build_grib_layers` requires `base_ft` *and* `top_ft` per band (`clouds.py:569`). HRRR has base without top, so `has_usable_band` stays False and `build_nwp_cloud_layers` returns `None`. That marks the default `ogimet_nwp` icing method unavailable, gates native SFIP off, and leaves the 3 km microphysics — the entire reason to fetch HRRR — unreachable by the advisory that most needs it. The branch fetches CLMR and CIMIXR on all 40 levels and then nothing reads them for cloud geometry. We fixed it with a third builder, `build_nwp_cloud_layers_from_condensate`: geometry from the condensate, amount from the model's own published band cover, threshold pinned to SFIP's lowest non-zero liquid bin.

**2. `_nwp_cloud_cover_at` returns a confident 0 % over an overcast column.** The gate on your base is `any(layer.cover_pct is not None)` — cover, not geometry. HRRR has cover *without* geometry, so it enters the geometry branch, matches nothing, and falls through to `return 0.0`. We regated it on geometry. Worth noting this was never HRRR-specific — ECMWF's `lcc`/`mcc`/`hcc` hit the identical path, and the test covering it asserted `0.0` while its own name said `uses_icao_bands`; the defect was pinned as the design.

**3. The convective track reads NONE on a convection-allowing model.** This is the worst one. HRRR's `total_cover_pct` / `ceiling_ft` / band covers make `_has_native_cloud_content()` return True, so `assess_convective_nwp` takes the **native** path — which grades on `convective_top_ft` / `convective_cover_pct` / `convective_precip_mm_h`, all `None` for HRRR (no parameterized scheme). It falls to the `else` branch, `_native_convective_risk(None, None)` → `ConvectiveRisk.NONE`, with `method="nwp"`. So every HRRR-sourced US briefing reports the model as *quiet on convection*, with a method string that tells `dd_nwp_agreement` this is a genuine model opinion — which then fires DD-vs-NWP divergence on every convective day. The 180-0 mb MLCAPE you correctly download is used for nothing but CIN suppression.

Fair caveat on this one: `convective_scheme_absent` didn't exist on your base — we added it in review round 1. But that's the point. The field is the fix; the gap is not having discovered the need.

### Correctness gaps inside the fetch layer

**4. `round()` is banker's rounding — half-hour departures silently lose half the native hours.** `_snap_to_hrrr_grid` uses bare `round()`. Running your own `hrrr_window_hours`:

```
dep 09:00 init 00z dur 4h -> [9, 10, 11, 12, 13]
dep 09:30 init 00z dur 4h -> [9, 10, 12, 14]     <- f11 and f13 missing
dep 09:45 init 00z dur 4h -> [9, 10, 11, 12, 13, 14, 15]
```

A :30 departure — a very common GA slot — degrades HRRR's headline hourly resolution to 2-hourly, filled by interpolation. **We shipped the identical bug** in our first commit (`290f4f4d`); review caught it and replaced it with a contiguous `floor(departure)..ceil(end)` range plus a property test.

**5. A partial byte-range download commits a truncated blob to cache.** `fetch_cloud_diag_ranges` silently drops failed spans (`if data is not None`), and `put_cached` fires on any non-empty result — so a truncated ~190 MB sounding serves as a valid cache hit for the whole TTL. We added plan validation before spending the bytes, a raise on a failed span so nothing commits, per-column completeness validation (every level must carry temperature/wind/height/humidity/microphysics), and delete-and-refetch-once when a cached artifact fails validation.

**6. Attribution isn't atomic.** `total_replaced > 0` flips the slot to `hrrr:noaa` and the `GFS (HRRR)` badge. But `fill.py` only rebuilds gap hours strictly *between* two anchors, so hours outside the merged span keep their Open-Meteo/GFS sounding under an HRRR badge — the code comment reasons about the bracketed case and not the unbracketed one. Separately, uncovered route points become `{}` and keep Open-Meteo levels while their neighbours get HRRR: a spatially mixed slot within one hour. We stage every hour — fetch, decode, validate — before writing anything, so one bad hour leaves the GFS pack untouched, and an hour covering only some route points is skipped whole.

**7. Memory.** No range coalescing (315 messages → 315 HTTP range requests per fhour; ours coalesces adjacent messages to 106 with no extra bytes), the full ~190 MB assembled in memory as a `bytearray` and then copied to `bytes`, and then `decode_hrrr_pressure` does `Path(file_path).read_bytes()` and writes it straight back out to a `NamedTemporaryFile` for cfgrib. Ours streams chunks to the cache file via `put_cached_from_chunks` and hands cfgrib the cache path directly. This one is operational, not stylistic — GRIB precache is currently disabled in production after an OOM killed a user briefing, so a transient ~3× on a 190 MB/fhour download matters.

**8. Cache lifecycle.** 6 h TTL is what we shipped first and then raised to 9 h: only 00/06/12/18z reach 48 h, so an extended run's successor is 6 h away and the 12z run gets purged at 18z while the 18z run is still inside its ~1.25 h publication delay. Also: no size cap (`hrrr` absent from `_DEFAULT_CACHE_CAP_GIB`), not in the scheduler's daily retention pass, and `purge_old_runs(data_dir, model="hrrr")` runs on the hot path with no pinning — a concurrent briefing on an older extended cycle can have its directory deleted mid-decode. We added an 8 GiB cap, the retention-pass entry, and a `pin_run_dir` registry that both eviction rules respect.

**9. Grid detection is loose.** `_lambert_dataset_axes` triggers on the mere presence of `y`/`x` dims and **falls back to pinned HRRR CONUS constants** when the GRIB attrs are missing — a different Lambert product (HRRR-Alaska, a future RRFS) would silently decode against the wrong grid. Scan-direction attrs (`iScansNegatively` / `jScansPositively`) are ignored, and winds are rotated unconditionally without checking `GRIB_uvRelativeToGrid`, so an earth-relative product would be double-rotated. Ours requires `GRIB_gridType == "lambert"` plus all eight attrs or returns None, honours scan direction, derives the rotation numerically from the projection itself (finite-difference north vector, so it can't drift from whatever projection the decode used), and warns loudly when the orientation flag is missing rather than silently skipping rotation.

**10.** `ml_cin` positive-value passthrough is unguarded (you list it as a known follow-up). We map `> +5 J/kg` to `None`, so a provider convention flip can't turn a strong cap into "no inhibition".

---

### Bottom line

Merged as-is, a US route would get a genuinely better 3 km sounding, correctly badged — and would simultaneously lose its NWP cloud envelope, report 0 % cloud cover at every altitude, and report no convective risk from a convection-allowing model. Those are precisely the three advisories a US pilot would want HRRR *for*.

The honest framing: #457 describes the work almost entirely in fetch/decode terms — "the genuinely new work" is Lambert interpolation, sounding replacement, wiring. This branch implements that spec faithfully and competently, at roughly the quality of our own first commit, including one of the same bugs. The distance between the two PRs is almost entirely five review rounds, and what those rounds bought was mostly *product* correctness the issue never described: what happens to the advisories when a model turns up with a shape no existing consumer was written for.

Closing as superseded by #508. Picking up the three items above.

roznet:
Closing as superseded by #508 (merged). Full comparison in the review comment above — three ideas from this branch are being applied to main.

PR #526, opened 2026-08-01
fix(verification,profile): convert the last two MSL ceiling reads to AGL

### LINKED ISSUE #445: GRIB decode audit: deferred tail — datum reads + scalar-interp caching
Small deferred tail from the GRIB decode audit (#441, delivered in #443 + #444). None blocking; grouped here so the big audit issue can close.

## 1. Datum: two remaining same-class reads (AGL/MSL)

The ceiling AGL/MSL datum fix (#444) converted the airport panel, verification scoring (flight + standalone), forecast map, and route alternates. Two reads of the same raw MSL ceiling fields were **not** converted:

- **`src/weatherbrief/tasks/standalone_verification.py:1208-1212`** — `cloud_base_delta_ft` / `lcl_delta_ft` compute `snap.cloud_base_ft - obs.ceiling_ft` and `snap.lcl_ft - obs.ceiling_ft`. `cloud_base_ft` is MSL for GFS/ICON (AGL for ECMWF) and `obs.ceiling_ft` (METAR) is AGL, so the delta is biased by field elevation at elevated airports. (`lcl_ft` is already AGL — the Espy surface approximation — so its delta is fine; only `cloud_base_ft` needs conversion.) Convert `cloud_base_ft` to AGL via the shared `to_agl_ceiling` + `get_airport_elevations` before differencing.
- **`src/weatherbrief/api/airport_profile.py:330`** — the airport-profile chart series reads raw `nwp_ceiling_ft` / `sounding_ceiling_ft` (MSL for GFS/ICON) for display. Convert to AGL for consistency with the rest of the datum-aware surfaces.

Both shift verification/display metrics, so pair the change with a before/after diff, as with the #444 corpus replay.

## 2. Efficiency: scalar-diagnostic coordinate caching

The vectorisation in #444 (`_bilinear_grid_weights`) covered the pressure-level and ICON model-level paths. The **scalar/surface** decode paths still call `_interpolate_per_point` (per-variable xarray `.interp`) rather than computing grid indices/weights once per dataset:

- `src/weatherbrief/fetch/grib/decode.py` — `decode_cloud_diag_per_point` (GFS), `decode_ecmwf_surface_per_point`, and the ICON cloud-diag path.

Reuse `_bilinear_grid_weights` (already extracted) to compute corner indices/weights once per dataset and gather all surface variables in one op — same equivalence bar as the ICON vectorisation (numerically identical, regression-tested).

---

Delivered audit: #443 (7 correctness findings + `.idx`/buffer efficiency), #444 (AGL/MSL datum class, ICON model-level vectorisation, ECMWF multi-grid resolved as a no-op). See #441 for the full evidence trail and grounding method.

roznet:
## What & why

The #441 finding-#3 datum sweep (delivered in #444) converted the airport panel, verification scoring, forecast map and route alternates to a common AGL datum — but two reads of the same raw MSL ceiling fields were left behind. Both are fixed here, which closes out the datum class properly.

**1. `_score_cycle` — `cloud_base_delta_ft` was elevation-biased**

`snap.cloud_base_ft` carries the model's datum (MSL for GFS/ICON/HRRR, AGL for ECMWF) and was differenced directly against the AGL METAR ceiling, so the stored delta was off by the field elevation at elevated airports. It now converts through the shared `to_agl_ceiling` + `_nwp_ceiling_is_agl` helpers, reusing the `elev_map` that #444 already loads a few lines up in the same function — no extra lookup or query.

`lcl_delta_ft` on the next line is deliberately left alone: `lcl_ft` is the Espy surface approximation (`_LCL_CONSTANT_FT * (T2m - Td2m)`), a height above the *station*, so it is already AGL and must not have field elevation subtracted a second time. A test pins that asymmetry so it doesn't read as an oversight later.

**2. Airport-profile chart — MSL ceilings plotted against AGL METARs**

`_surface_from_cache` emitted raw `nwp_ceiling_ft` / `sounding_ceiling_ft` for display. It now takes the field elevation the endpoint already resolves via `_resolve_airport_coords` and converts per source: the NWP ceiling on the model's datum, the sounding ceiling always MSL (it's geopotential height for every model, including ECMWF, whose *NWP* ceiling is the AGL one).

Both call sites degrade to the previous datum-naive value when no elevation is available, matching the convention `to_agl_ceiling` already established elsewhere.

**Scope note — item 2 of the issue is intentionally not here.** The issue also proposed reusing `_bilinear_grid_weights` in the scalar/surface decode paths (`decode_cloud_diag_per_point`, `decode_ecmwf_surface_per_point`). That is pure perf with a numerical-equivalence bar, it runs in `decode_worker` subprocesses off the request path, and those paths have churned substantially since the issue was filed (NCEP pressure-band carving, native cover provenance, dual microphysics). Dropped rather than deferred — worth reopening as its own issue only if decode latency shows up in profiling.

## Issue linkage

Closes #445

## Testing / verification

New tests:

- `test_cloud_base_delta_converts_msl_model_to_agl` — GFS cloud base 2800 MSL at a 1500 ft field vs a 3000 ft AGL METAR ceiling gives -1700, not -200; asserts in the same test that `lcl_delta_ft` is unchanged at -1000.
- `test_cloud_base_delta_leaves_ecmwf_agl_alone` — ECMWF cloud base passes through unconverted.
- `TestSurfaceCeilingDatum` (5 cases) — MSL→AGL conversion, ECMWF pass-through, the sounding fallback always being MSL, a literal 0 ft ceiling still not falling through to the sounding fallback, and the no-elevation legacy path.

The pre-existing `test_cloud_base_and_lcl_deltas` still passes unchanged (its fake DB path fails the elevation lookup, exercising the datum-naive fallback) and now carries a comment saying so.

Full suite: 4753 tests, all pass except `tests/test_auth.py::TestLoginRedirect::test_google_login_redirects`, which fails identically on a clean checkout of this branch's base (no Google OAuth config in the sandbox) — verified by stashing the diff and re-running.

Both changes shift stored/displayed metrics. `cloud_base_delta_ft` / `lcl_delta_ft` currently have no reader — no endpoint, no rollup, no UI (the `cloud_base_mae_ft` aggregate is still future work in `designs/future/standalone-verification-plan.md`) — so the change affects newly-written score rows only; existing rows keep their datum-naive values. The profile chart is user-visible and will show lower ceilings at elevated airports, which is the point.

---
_Generated by [Claude Code](https://claude.ai/code/session_0145eeuL4oEXGXbE4mWjGWNh)_

### DIFF
--- a/designs/metar-taf-accuracy.md
+++ b/designs/metar-taf-accuracy.md
@@ -557,6 +557,17 @@ def compute_verification_score(
     )
 ```
 
+**Ceiling datum.** Every ceiling that meets a METAR ceiling in a delta is first
+converted to **AGL**, because the METAR side always is. `model_ceiling` gets
+this from `reconcile_ceiling(..., field_elevation_ft=…, model=…)`; the extra
+`cloud_base_delta_ft` on the standalone path converts `cloud_base_ft` with the
+same `to_agl_ceiling` + `_nwp_ceiling_is_agl` helpers before differencing.
+`lcl_delta_ft` is the exception that looks like a bug: `lcl_ft` is the Espy
+surface approximation, a height above the *station*, so it is already AGL and
+must not have field elevation subtracted. Field elevations come from
+`get_airport_elevations`; when the lookup fails, scoring degrades to the legacy
+datum-naive deltas rather than failing the cycle. (#441 finding #3)
+
 ### TAF → METAR Comparison
 
 TAF fields are already parsed and stored in `verification_observations`. The TAF-derived flight category uses the same `classify_flight_category()` as advisories.
--- a/src/weatherbrief/api/airport_profile.py
+++ b/src/weatherbrief/api/airport_profile.py
@@ -268,13 +268,23 @@ def _surface_from_cache(
     icao: str,
     model: str,
     hours: list[datetime],
+    field_elevation_ft: float | None = None,
 ) -> list[dict[str, Any]]:
     """Read surface fields from airport_forecast_snapshots for each hour.
 
     Picks the most recent snapshot per (icao, model, forecast_hour). Hours
     with no data simply don't appear in the result; the frontend treats
     those as gaps.
+
+    ``ceiling_ft`` is emitted on the AGL datum when ``field_elevation_ft`` is
+    given, matching every other ceiling surface (map, alternates, advisories)
+    and the AGL METAR ceilings the chart plots alongside it. Without an
+    elevation the legacy datum-naive value is returned. (#441 #3)
     """
+    from weatherbrief.analysis.airport_conditions import (
+        _nwp_ceiling_is_agl,
+        to_agl_ceiling,
+    )
     from weatherbrief.db.models import AirportForecastSnapshotRow
 
     if not hours:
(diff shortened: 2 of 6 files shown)


roznet:
Fixed the Important finding in 8e3bdba.

The finding checks out: `VizPoint.soundingCeilingFt` is declared ft MSL, `ceiling-dd` subtracts `terrainElevationFt` to reach AGL, and this PR turned the adapter's `?? surface?.ceiling_ft` fallback into an AGL source — so the field would have carried two different datums depending on which side of the `??` fired. Confirmed not reachable today: `soundingCeilingFt` has exactly two read sites (`route-graph/metrics.ts:154` and the two assignments), the profile panel renders no route graph, and the briefing route graph is fed from `data-extract.ts:437`, which is untouched.

Dropped the fallback rather than converting back to MSL — the AGL→MSL round-trip is lossy (`to_agl_ceiling` clamps at 0, so a below-field ceiling can't be recovered), and the profile's AGL ceiling is already available to callers as `surface.ceiling_ft`. The call site now carries a comment saying why the fallback is deliberately absent, which is the "don't reintroduce the double-subtraction" guard the review asked for.

Nothing else deferred — the review found no other issues.

Verification: `npm test` 581 passed / 36 files, `npm run build` clean, `npx tsc --noEmit` down to 2 pre-existing errors in `ts/eval/label-panel.ts` (unrelated). Python suite unchanged from the PR body: 4716 passed, with `tests/test_auth.py::TestLoginRedirect::test_google_login_redirects` failing identically on a clean checkout of the base (no Google OAuth config in the sandbox).

---

PR #528, opened 2026-08-01
MySQL usage optimization: correctness fixes, index hygiene, session lifetime, query shape

downle:
From a full three-track audit of the MySQL layer (all 84 alembic migrations, every `storage/` module, the `flyfun_common` engine + every `SessionLocal()` call site), with every top finding independently re-verified against the code before fixing. Branch is based on latest `main` (`4abd5b49`) and merges cleanly. Design + plan: `docs/superpowers/` (superpowers workflow, per-task reviews + final whole-branch review).

## Correctness & reachable outages

- **`briefing_packs` UNIQUE `(flight_id, fetch_timestamp)`** — the table had no such constraint; `save_pack_meta` inserted unconditionally while readers use `scalar_one_or_none()`, so one raced duplicate (manual refresh × scheduler auto-refresh) would 500 every subsequent read of that pack. Migration 085 dedupes (keeps newest row per pair, logs counts) and adds the constraint; `save_pack_meta` is now race-safe (savepoint + IntegrityError → update core, nested-retry-free).
- **Airport-profile SSE pool exhaustion** — the endpoint held a pooled session for the stream's whole life while the limiter allows 20 concurrent streams vs a 15-connection pool (arithmetically reachable 500s). Now closes the session before streaming, mirroring the proven `packs.py` pattern.
- **Pooled connections released before pipelines/network** — five sites (`process_auto_refreshes`, `_auto_refresh_one`, `collect_and_store`, `_run_scan_job`, `refresh_resume`) pinned web-pool connections across minutes of GRIB/LLM/network work. Now: read → close → sessionless slow work → fresh session to finalize. Notify-after-commit ordering preserved.
- **Deadlock/lock-wait retry helper** (`db/retry.py`) — there was none anywhere despite ≥6 concurrent writer roles. Retries only 1213/1205 (errno-first with `__context__`/`__cause__` chain-walking, because a 1213 inside a savepoint comes back masked as 1305), jittered backoff, applied to the hot write paths (pack meta, `subscribe_flight`, refresh-job write-through).

## Index hygiene (migration 085)

- **Missing, added**: `ix_briefing_usage_timestamp` (4 range-filtered call paths), `ix_cost_ledger (service, created_at)` (global cost rollups), `ix_verif_scores_obs_time` + `ix_taf_verif_obs_time` (monthly retention range-DELETEs on the largest tables), `uq_users_provider_sub` (the OAuth identity lookup, with a zero-duplicates pre-check).
- **Redundant, dropped (14)**: 13 leftmost-prefix duplicates of UNIQUE/PK columns + the `oauth_refresh_tokens.token_hash` unique+plain duplicate (guarded). Kept deliberately: `ix_verif_scores_lead/_model` (needs `sys.schema_unused_indexes` evidence), FORCE-INDEXed `ix_verif_scores_source_time`, and **`ix_afs_hour_model`** — the final review caught that the map path still filters bare `forecast_hour` (`maps.py`, `map_queries.py`, `cache_builder.py`), so its documented drop premise is false; kept with a comment.
- **Money → DECIMAL**: `cost_ledger.cost`, `donation_ledger.amount/amount_usd/fx_rate/net_usd` → `DECIMAL(12,4)`/`(12,6)` (was single-precision FLOAT). Python consumption verified safe (existing `float()`/pydantic coercions); the proper `Numeric` ORM remap belongs upstream in `flyfun_common`.

## Query shape

- **`latest_pack()`** — hot refresh gates loaded a flight's *entire* pack history (5 `json.loads` + 2 HMAC-SHA256 + 2 fs stats per row) to use only the newest. Now `LIMIT 1`, no HMAC, no artifact resolution on that path.
- **Flights-list aircraft batch prefetch** — one `IN` query instead of per-flight `db.get` (the audit found the identity map never deduped, so even shared aircraft repeated SELECTs).

## Tooling & docs

- `scripts/check_mysql_charset.py` — reports per-table engine/collation, exits non-zero on non-InnoDB/non-utf8mb4, guarded `sys.schema_unused_indexes` dump, clean errors (no false-pass).
- `designs/data-models.md` — FK policy: intentional no-FK set vs survive-deletion candidates vs oauth oversight set, with per-table recommendations (no FKs added here — cascades are product decisions); migration conventions (utf8mb4 on new tables, guarded index ops, dedupe-keeps-newest).

## Verification

- **Full suite: 4705 passed.** Only failures: `test_google_login_redirects` (environment-dependent, fails on base too) and 2 `test_grib_precache` order-dependent flakies (pass standalone; fail identically on base full-suite runs).
- New tests: migration chain + dedupe + idempotency, unique-race, session-scope discrimination (spy on close-before-pipeline/stream), `latest_pack` semantics, prefetch query counting, retry helper (chain-walking, non-retryable negatives, nesting), script smoke.
- Per-task spec+quality reviews (9) + final whole-branch review; three final Important findings fixed pre-merge (the `ix_afs_hour_model` keep, the 1305-masking detection, the nested-retry split).

## Before prod rollout (recorded gates, none merge-blocking)

1. Run migration 085 against a **scratch MySQL** with seeded duplicates (MySQL-only paths: derived-table dedupe vs error 1093, `information_schema` guards, FLOAT→DECIMAL ALTER, 1305-masked savepoint shape) — all reasoned, SQLite-verified, not executed.
2. `scripts/check_mysql_charset.py` against prod (one command — settles the charset question for good).
3. Post-deploy `EXPLAIN` on the map path (kept `ix_afs_hour_model` should serve it) and, after some uptime, `sys.schema_unused_indexes` to judge `ix_verif_scores_lead/_model` and the `token_hash` drift index.
4. Follow-ups (separate issues): `flyfun_common` engine tuning (pool size/timeouts/charset) + `Numeric` money remap + commit-level retry in `get_db`; FK decisions per `designs/data-models.md`; the preserved zero-airports `_record_cycle` never-commits latent bug (flagged); `run_standalone_cycle` in-process mode.

🤖 Generated with [Kimi Code](https://www.kimi.com/code) via the superpowers subagent-driven workflow — co-authored on the commits via `Co-authored-by`.

### DIFF
--- a/alembic/versions/085_mysql_review_fixes.py
+++ b/alembic/versions/085_mysql_review_fixes.py
@@ -30,12 +30,19 @@ Tier-1 item 4, Tier-2 items 5-7):
   indexed and duplicate-identity races impossible. A pre-check refuses with a
   clear error if duplicates exist (expected zero; they would need a manual
   account-merge decision, never a silent dedupe).
-* **Drop 15 redundant indexes** — 13 leftmost-prefix duplicates of wider
-  UNIQUE keys, the non-unique ``oauth_refresh_tokens.token_hash`` duplicate,
-  and ``ix_afs_hour_model`` (081's documented follow-up). Never dropped:
-  ``ix_verif_scores_lead`` / ``ix_verif_scores_model`` (need
-  ``sys.schema_unused_indexes`` evidence) and the FORCE-INDEXed
-  ``ix_verif_scores_source_time``.
+* **Drop 14 redundant indexes** — 13 leftmost-prefix duplicates of wider
+  UNIQUE keys and the non-unique ``oauth_refresh_tokens.token_hash``
+  duplicate. Never dropped: ``ix_verif_scores_lead`` /
+  ``ix_verif_scores_model`` (need ``sys.schema_unused_indexes`` evidence),
+  the FORCE-INDEXed ``ix_verif_scores_source_time``, and
+  ``ix_afs_hour_model`` — dropping that one was considered (081's documented
+  follow-up assumed region-aware serving) and rejected: the map path still
+  filters bare ``forecast_hour`` with no region predicate
+  (``api/maps.py:118-129`` availability scan,
+  ``tasks/map_queries.py:262-266`` per-model init-time probe,
+  ``tasks/cache_builder.py:275-279`` rebuild probe), so the region-leading
+  ``ix_afs_region_hour_model`` cannot serve those queries and the drop would
+  restore full/per-model scans on the hot map path.
 * **Money → DECIMAL** — ``cost_ledger.cost`` and
   ``donation_ledger.amount/amount_usd/net_usd`` to ``DECIMAL(12,4)``,
   ``donation_ledger.fx_rate`` to ``DECIMAL(12,6)``.
@@ -144,11 +151,18 @@ _NEW_INDEXES = [
     ("users", "uq_users_provider_sub", ["provider", "provider_sub"], True),
 ]
 
-# Spec Tier-2 item 5, verbatim: the 13 leftmost-prefix duplicates plus
-# ix_afs_hour_model (081's follow-up). Columns are kept on each row so the
-# downgrade can recreate them. The 15th drop — the non-unique duplicate on
-# oauth_refresh_tokens.token_hash — is handled separately in upgrade()
-# because its guard is uniqueness-aware.
+# Spec Tier-2 item 5: the 13 leftmost-prefix duplicates. Columns are kept on
+# each row so the downgrade can recreate them. The 14th drop — the non-unique
+# duplicate on oauth_refresh_tokens.token_hash — is handled separately in
+# upgrade() because its guard is uniqueness-aware.
+#
+# ix_afs_hour_model is deliberately NOT in this list: the spec's premise for
+# dropping it (081's "drop once serving filters by region" follow-up) is
+# unmet — the map path still filters bare forecast_hour with no region
+# predicate (api/maps.py:118-129, tasks/map_queries.py:262-266,
+# tasks/cache_builder.py:275-279), and the region-leading
+# ix_afs_region_hour_model needs a region predicate to be usable. Dropping it
+# would restore the full/per-model table scans migration 077 fixed (#415).
 _REDUNDANT_INDEXES = [
     ("verification_observations", "ix_verif_obs_icao", ["icao"]),
     ("verification_scores", "ix_verif_scores_icao", ["icao"]),
@@ -167,7 +181,6 @@ _REDUNDANT_INDEXES = [
     ("analytics_event_daily", "ix_aed_day", ["day"]),
     ("analytics_briefing_feature_daily", "ix_abfd_day", ["day"]),
     ("analytics_xsection_config_daily", "ix_axcd_day", ["day"]),
-    ("airport_forecast_snapshots", "ix_afs_hour_model", ["forecast_hour", "model"]),
 ]
 
 # Spec Tier-2 item 7: (table, column, target type, nullable).
@@ -231,11 +244,13 @@ def upgrade() -> None:
     for table, name, columns, unique in _NEW_INDEXES:
         _create_index_if_absent(name, table, columns, unique=unique)
 
-    # --- Drop the 15 redundant indexes (spec Tier-2 item 5) ---
+    # --- Drop the 14 redundant indexes (spec Tier-2 item 5) ---
     # Leftmost-prefix duplicates of wider UNIQUE keys; DROP INDEX is
     # metadata-only on MySQL 8. Kept: ix_verif_scores_lead / _model (awaiting
-    # sys.schema_unused_indexes evidence) and FORCE-INDEXed
-    # ix_verif_scores_source_time.
+    # sys.schema_unused_indexes evidence), FORCE-INDEXed
+    # ix_verif_scores_source_time, and ix_afs_hour_model (drop considered and
+    # rejected — the map path still filters bare forecast_hour, see the note
+    # above _REDUNDANT_INDEXES).
     for table, name, _columns in _REDUNDANT_INDEXES:
         _drop_index_if_present(table, name)
     # oauth_refresh_tokens.token_hash: migration 041 declared the column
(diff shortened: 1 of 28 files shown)


roznet:
Thanks @downle — this was a serious piece of work, and reviewing it was genuinely worth the time. I've gone through every finding and verified each one against production (schema metadata, row counts, and ~29k lines of container + archived logs). I'm going to close this PR rather than merge it, but several findings are being picked up and one of them is the most valuable thing anyone has surfaced about this codebase in a while. Follow-ups are tracked in #532.

## What we're taking

**The airport-profile SSE session leak.** This was the one genuinely reachable bug in the PR and the analysis was exactly right: `_MAX_GLOBAL = 20` concurrent streams against a pool that tops out at 15 (`pool_size=5 + max_overflow=10`, since the engine sets no pool config), with `Depends(get_db)` teardown deferred until the `StreamingResponse` finishes. Arithmetically reachable, and your fix correctly mirrors the pattern `packs.py` already uses. Being picked up as-is.

**Dropping `ix_verif_scores_icao` and `ix_vds_date_model`.** Already done — set `INVISIBLE` as a first phase, with the real `DROP` to follow after an observation window. Your leftmost-prefix analysis was correct on all 13, and it landed on something real: `verification_scores` is 8.8M rows carrying **2052 MB of index against 968 MB of data**. The other 11 are correct too, but on tables small enough that the benefit isn't measurable.

**The FK policy audit** (`designs/data-models.md`). Auditing every FK-shaped column across all 84 migrations and sorting them into intentional / undocumented / oversight is work we would not otherwise have done, and it's what surfaced the item below. Being lifted into the design docs on its own.

**The account-deletion gap.** Your FK table flags in one line that `delete_account` never sweeps the OAuth tables, and that this is a security issue rather than hygiene. That's correct, and on verification it's somewhat worse than the entry suggests. It's the single most valuable finding in this PR and it's the one thing here that isn't really about MySQL at all. Being handled upstream in flyfun-common and tracked privately, since it's live and both repos are public.

**Engine pool configuration.** Listed in your PR body as a deferred follow-up rather than a finding, but you're right that it's unset: no `pool_pre_ping`, no `pool_recycle`, against MySQL's 8-hour `wait_timeout`. Queued upstream.

## What we're declining, and why

Mostly because production evidence doesn't support the premise. Some numbers, since they may be useful to you:

**The deadlock/lock-wait retry helper.** 28,775 lines of container and archived predeploy logs contain **zero** 1213s, zero 1205s, zero "MySQL server has gone away", zero QueuePool timeouts. This is the largest single piece of new machinery in the PR — 189 lines plus 356 lines of tests — defending against something never observed. It also plants `session.rollback()` inside `save_pack_meta`, `subscribe_flight` and `_best_effort`; that discards *all* uncommitted work on the session. The module docstring's "retry loops must never nest" doctrine shows you saw the hazard, but it's now a sharp edge in paths reachable from request handlers.

**`ix_verif_scores_obs_time` and `ix_taf_verif_obs_time`.** These are justified as serving the retention range-DELETEs on the largest tables, but raw retention is disabled in production — `VERIFICATION_RAW_RETENTION_DAYS` is unset, so it defaults to 9999 and those DELETEs never execute. Separately, #527 (in flight, verification data tiering) removes both justifications: its prune deliberately adds a `source` predicate so the *existing* `ix_verif_scores_source_time` serves it — its own comment reads "nothing on this table leads with observation_time" — and its monthly rollup reads `verification_daily_stats` instead of raw scores. So the index would add ~200 MB to the very table #527 exists to shrink.

**The `briefing_packs` UNIQUE constraint**, as a correctness fix. The premise is a raced duplicate `(flight_id, fetch_timestamp)` breaking `scalar_one_or_none()`. But that column is `datetime(6)` in production and each refresh takes `datetime.now()`, so a collision requires two refreshes at the identical microsecond. Prod has zero duplicates across 3,005 rows and zero `MultipleResultsFound` in the logs — the dedupe `DELETE` in migration 085 would run against nothing. Worth noting the finding *would* have been real on a plain `DATETIME` column, since MySQL truncates to whole seconds — that's a trap documented in our own `CLAUDE.md`. It's kept in #532 as optional belt-and-braces, not as a fix.

**The money FLOAT → DECIMAL migration.** Diagnosis is right — those really are 4-byte floats and that's the wrong type for money. But as you note yourself, the `Numeric` remap belongs upstream in flyfun_common. Doing the column half first, from this repo's chain, would leave DECIMAL columns mapped as `Float` and drift the schema from the package that owns those models. Queued upstream in the right order instead.

**The session-lifetime rework** across the five sites. This is the most invasive change here and the pool-exhaustion framing undersells it — the scheduler loop is sequential, so it only ever holds 1 of 15 connections. The stronger argument, which the PR doesn't make, is InnoDB history-list growth from a REPEATABLE READ view held open across minutes of GRIB work. For what it's worth I did check the object-crossing risk and the splits are technically sound (`icao_to_flights` is `dict[str, set[str]]`, `_row_to_flight` returns a plain dataclass, so no `DetachedInstanceError`). But it changes two public signatures, knowingly preserves a never-commits bug, and conflicts with #527 on `scheduler.py`.

**`scripts/check_mysql_charset.py`.** Every table in production is already InnoDB + `utf8mb4_unicode_ci`. The script would confirm a settled question.

## Mechanical blockers

Independent of merit, this couldn't have merged as-is:

- **Alembic revision collision.** `085_mysql_review_fixes.py` declares `revision="085", down_revision="084"` — `main` already has `085_briefing_usage_cache_tokens.py` with exactly that pair. Two heads. And 086 is taken by #527, so this would have needed 087.
- Based on `4abd5b49`, 18 commits behind, with conflicts in `airport_profile.py`.
- Overlaps #527 on `db/models.py`, `scheduler.py` and `verify/__main__.py`, and #527 substantially rewrites the verification task layer this audit reasons about — including adding three tables (`verification_global_daily_stats`, `verification_activity_daily`, `taf_verification_daily`) the index pass never saw.

## The honest summary

The static analysis is strong — the SSE arithmetic, the leftmost-prefix work, and the observation that the weak-ref identity map doesn't dedup `db.get` across a list are all correct and none of them are obvious. The gap is that almost nothing was checked against the running system. Live column types, whether the optimized code paths are even enabled, and whether the failure modes being defended against have ever occurred would each have redirected significant effort. Your own "before prod rollout" gates say as much — the MySQL-only paths were "reasoned, SQLite-verified, not executed" — and that's precisely where the findings diverged from reality.

Worth saying plainly: the best thing in this PR is a single row in a documentation table, not any of the code. Thank you for it, and for the audit that produced it.

Closing in favour of #532.

Issue #411, opened 2026-07-13
SLD detection is dormant end-to-end — validate against a real winter freezing-rain case


roznet:
**Do not action this in summer — it needs a real winter freezing-rain case to validate against. Filed now so the evidence isn't lost.**

Found while adding `reason_code` to `fiki_icing` (#PR 5fe6b571): I wrote a RED `sld` reason code, then asked whether the branch was reachable. It isn't. Pulling that thread showed the whole SLD feature is dormant end-to-end — **three broken links in a row**, each of which hides the next.

## The chain

**Link 1 — coalescence is disabled (correctly).**
`analysis/sounding/sld.py:266-274`. Mechanism 2 (collision-coalescence in deep warm-top cloud) is commented out, with the reason stated in-code:

> *"Without droplet size data, deep warm-top clouds cannot be reliably distinguished from ordinary icing clouds."*

That looks like the right call and I'm not proposing we revert it. But it matters because **it was the only mechanism ever producing SLD in our stored packs** (see Evidence).

**Link 2 — warm-nose has never fired.**
`detect_warm_nose` (`analysis/sounding/precipitation.py:174`) is the only live mechanism. Across **5,554 full-profile soundings** it produced **zero** warm noses — despite **147** of them having the sub-freezing surface wet-bulb the gate requires. So the gate got its chance 147 times and declined every one.

That may be entirely correct — a cold surface with cold air all the way up is just a normal cold day, and freezing rain needs the specific warm-nose-over-cold-surface structure, which is genuinely rare. **We cannot tell correct from under-sensitive without a known FZRA case.** That is what this issue is waiting for.

One place I'd look first: the detector requires **≥2 levels with Tw ≥ 0** inside the nose (`precipitation.py:227`). On a pressure-level grid (~25–50 hPa ≈ 700–1,500 ft apart) a thin warm nose could be resolved by a single level and be missed. Unproven — but it is the obvious sensitivity knob.

**Link 3 — `IcingZone.sld_risk` is never populated.** *(a real bug, but currently inert)*
`IcingZone.sld_risk` is declared (`models/analysis.py:600`, `bool = False`) and **neither construction site sets it** — not `analysis/sounding/icing.py:340`, not `tasks/advise.py:173`. So `fiki_icing`'s `if zone.sld_risk` (`fiki_icing.py:58`) reads `False` unconditionally and its RED SLD branch cannot fire.

[...]

PR #531, opened 2026-08-02
feat(icon): fetch D2 qr/qs/qg + EU convective single-level, use the new species

### LINKED ISSUE #530: ICON-D2/EU: fetch qr/qs/qg + tke (and EU lpi_con_max/cape_con), and use the new species for precip phase & icing
## Why

An audit of DWD's open-data feed (verified live 2026-07-30) found several variables we have never fetched that are directly relevant to **icing** and **convection**. All of them are published on the `regular-lat-lon` grid our existing decoder already reads, under the same `.../grib/{cycle}/{var}/` layout — so this is a variable-list change, **not** a new source integration.

This is deliberately sequenced **before** the Météo-France/AROME work in **#529**: ICON-D2's extra hydrometeor species are near-free, and D2 already fully covers our core corridors (London–Paris–Cannes measured 61/61 route points inside the D2 domain).

## Verified availability

Directory listings under `https://opendata.dwd.de/weather/nwp/icon-{eu,d2}/grib/00/`:

| model | variable | grid | notes |
|---|---|---|---|
| ICON-D2 | `qr`, `qs`, `qg` | `regular-lat-lon` model-level | file count **3185**, identical to `qc` which we already fetch |
| ICON-D2 | `tke` | `regular-lat-lon` model-level | 3234 files |
| ICON-EU | `tke` | `regular-lat-lon` model-level | 6975 files |
| ICON-EU | `lpi_con_max`, `cape_con` | `regular-lat-lon` single-level | |

ICON-EU does **not** publish `qr`/`qs`/`qg` — its moisture set is `qc`, `qi`, `qv` only. Any per-variant variable list must reflect that.

Measured file sizes (run 00Z, step 003):

| file | size |
|---|---|
| ICON-D2 `qg` L40 | **2.4 KB** |
| ICON-D2 `qr` L40 | 18.7 KB |
| ICON-D2 `qc` L40 *(already fetched)* | 13.6 KB |
| ICON-EU `qc` L60 *(already fetched)* | 88.7 KB |
| ICON-EU `qi` L60 *(already fetched)* | 30.5 KB |
| ICON-EU `tke` L60 | **903 KB** |

D2's three extra species together cost roughly 2.5× a single `qc` file — about **21 MB** across 50 levels × 12 forecast hours. ICON-EU `tke` is the outlier at ~903 KB per level-hour (≈36 MB per forecast hour across 40 levels), because turbulence compresses badly; treat it as opt-in, not default.

## Changes

**1. Per-variant model-level variable list.**
`ICON_EU_VARIABLES` (`src/weatherbrief/fetch/grib/icon_eu_fetch.py:357`) is currently shared by both variants. Add the list to `IconVariant` (or otherwise make it per-variant) so D2 gets `qr`, `qs`, `qg` and EU does not. D2 uses `per_level_cache=True`, so each new species is cached per `(variable, level)` and a partial download stays detectable — no change to that machinery.

**2. Decode mapping.**
`_VAR_MAP` in `src/weatherbrief/fetch/grib/decode.py:32-42` — add:
- `qr` → `rain_water_kg_kg`
- `qs` → `snow_water_kg_kg`
- `qg` → `graupel_water_kg_kg`

Use these exact field names: Météo-France AROME publishes the same five ICE3 species as `crwc`/`cswc`/graupel, so the AROME integration in **#529** feeds identical fields and every consumer below stays model-agnostic.

**3. Data model.**
`PressureLevelData` (`src/weatherbrief/models/analysis.py:453`) — add `rain_water_kg_kg`, `snow_water_kg_kg`, `graupel_water_kg_kg`, all `Optional[float] = None`, alongside the existing `cloud_liquid_water_kg_kg` / `ice_mixing_ratio_kg_kg` (:464-465).

**4. Spatial interpolation.**
`src/weatherbrief/analysis/spatial_interpolation.py:316-361` interpolates condensate between route points. Extend to the three new fields following the existing `cloud_liquid_water_kg_kg` pattern exactly (including the `None`-guard on both endpoints — do not interpolate when either side is missing).

**5. Precipitation phase — highest-value consumer.**
`_compute_ice_fraction` (`src/weatherbrief/analysis/sounding/precipitation.py:136-146`) currently partitions `clwmr` vs `icmr` — i.e. **cloud** condensate — and uses that to classify *precipitation* phase. That is a proxy: cloud droplets and cloud ice are not what is falling.

With `qr`/`qs`/`qg` we can compute a genuine precipitation partition: liquid `qr` vs frozen `(qs + qg)`. Prefer it when present; fall back to the existing cloud-based ice fraction, then to wet-bulb (`_level_phase_from_wet_bulb`), preserving the current three-tier fallback shape in `_classify_levels` (:148-172). This is a **classification accuracy** improvement, not a grade change.

**6. Icing — supercooled rain.**
`qr > 0` at `T < 0 °C` is supercooled *rain*, physically distinct from supercooled cloud droplets and considerably more hazardous (freezing rain / large-droplet regime). Today nothing can distinguish the two: `analysis/sounding/__init__.py:101-114` and `clouds.py:590,651` only see cloud liquid/ice.

- Expose a derived per-level flag (e.g. `supercooled_rain` on `DerivedLevel`) computed from `rain_water_kg_kg` and `temperature_c`.
- Let icing severity read it.
- **Required as part of this issue:** add a numbered entry to `designs/meteorology-decisions.md` recording the threshold chosen (what `qr` value counts as non-trivial), the reasoning, and rejected alternatives. Per `.claude/CLAUDE.md` that log must be updated whenever a meteorology choice changes — a new input to icing grading qualifies.
- Validate against a real winter freezing-rain case before enabling the grade effect; coordinate with **#411**, which exists precisely to validate SLD detection end-to-end.

**7. ICON-EU convective single-level.**
Add `lpi_con_max` and `cape_con` to `ICON_EU_CLOUD_DIAG_VARIABLES` (`icon_eu_fetch.py:153`). We currently get lightning potential only from D2 (`lpi_max` in `ICON_D2_EXPLICIT_CONV_VARIABLES`, :207), whose domain is 43.18–58.08 N / −3.94–20.34 E. `lpi_con_max` extends a lightning signal across the full ICON-EU domain (29.5–70.5 N / −23.5–62.5 E), covering routes D2 cannot see. `cape_con` complements the `cape_ml` we already fetch.

**8. TKE — opt-in.**
Add `tke` for D2 (cheap) and ICON-EU (expensive) behind an explicit config flag, defaulting **off** for EU. We have no TKE consumer today, so this is data-availability groundwork; do not add a turbulence-grading consumer in this issue.

## Out of scope

- **Do not** wire `IcingZone.sld_risk` or `sounding.sld_zones` into advisories. That is deliberately dormant — see the extended comment at `src/weatherbrief/analysis/advisories/_helpers.py:387-401` and issue **#411**. Do not "clean up" the dormant clauses; the safe default documented there (SLD survives the filter) must be preserved. New species may improve the SLD detector's *inputs*, but advisory wiring stays out.
- `sdi_2` (supercell detection index) and `tcond10_mx` on D2 are recorded here as future candidates only. `tcond10_mx` is already flagged in the #468 discussion as the intended replacement corroborator for the dropped `grau_gsp` — leave it to that thread.
- No changes to ICON-EU moisture variables (it has no `qr`/`qs`/`qg`).

## Acceptance criteria

- [ ] A D2-enriched briefing yields non-null `rain_water_kg_kg` / `snow_water_kg_kg` / `graupel_water_kg_kg` at levels where precipitation exists; values are plausible (`qg` is frequently 0 — that is correct, not a bug).
- [ ] An ICON-EU-only briefing is unchanged: new fields stay `None`, no download-size change, no regressions.
- [ ] Precipitation phase uses the `qr` vs `qs+qg` partition when available, with unit tests covering snow / rain / mixed and the fallback chain when species are absent.
- [ ] Supercooled-rain flag is produced for a constructed profile with `qr > 0` below 0 °C; `designs/meteorology-decisions.md` entry added.
- [ ] Download-volume regression test: the D2 addition stays under ~25 MB for a 12-hour window.
- [ ] ICON-EU `tke` remains off by default; enabling it is a deliberate config change.
- [ ] `designs/` docs updated where the variable inventory is described (see `weather-engine-specs.md`, `analysis-metrics.md`).

## Notes for the implementer

- Read `designs/meteorology-decisions.md` and `designs/weather-engine-specs.md` before starting (project convention: design docs first, then code).
- The `IconVariant` dataclass (`icon_eu_fetch.py:45-137`) is the single source of truth for per-variant behaviour — extend it rather than adding branches at call sites.
- D2 model levels are 16–65 and are **not** comparable to EU's 35–74 numbering; see the note at `icon_eu_fetch.py:250-257`.

roznet:
## What & why

ICON-D2 publishes three hydrometeor species we have never fetched — `qr` (rain water), `qs` (snow), `qg` (graupel) — on the same `regular-lat-lon` model-level layout our decoder already reads. ICON-EU publishes **none** of them, so the model-level variable list becomes **per-variant** (`IconVariant.model_level_variables`) rather than one shared tuple. Cost: ~21 MB over 50 levels × 12 forecast hours, against a D2 sounding that already runs ~494 MB/forecast-hour.

Two things they fix:

**1. Precipitation phase was reading the wrong condensate.** `_compute_ice_fraction` partitioned `clwmr` vs `icmr` — the *cloud* species, suspended, not falling — and used that to classify *precipitation* phase. It now uses the genuine partition, liquid `qr` vs frozen `qs + qg`, wherever a model supplies it, with the cloud proxy and then wet-bulb preserved as tiers 2 and 3. Severity thresholds are unchanged; only *which quantity* is thresholded changed. Classification accuracy, not a grade change.

**2. Supercooled rain was invisible.** `qr > 0` at `T < 0 °C` is freezing rain / the large-droplet regime — physically distinct from the supercooled cloud droplets the icing methods already grade, and considerably worse. `DerivedLevel.supercooled_rain` and `IcingZone.supercooled_rain` now carry it.

**The flag is reported, deliberately NOT graded.** It does not enter `risk`, `_enhance_severity`, or `is_hazardous`. The detector has never been run against a real winter freezing-rain case, and wiring an unvalidated signal into icing severity would move grades on every ICON-D2 winter briefing at once with nothing to check them against. Same dormant-contract shape as `IcingZone.sld_risk`; #411 is the validation that gates the upgrade.

### Drift prevention

The five hydrometeor mixing ratios are enumerated once in `CONDENSATE_LEVEL_FIELDS`. The GRIB field-patch merge, the **full-sounding-replacement passthrough** (the path ICON actually takes — `_convert_raw_sounding`, which would otherwise have decoded and interpolated the new species then dropped them on the floor), and both interpolation axes all derive their list from it. Same single-source-of-truth arrangement `NWP_CLOUD_DIAG_SCALARS` uses after #485.

### Also landed

- **ICON-EU `lpi_con_max` + `cape_con`** → `NWPCloudDiagnostics`. Lightning potential across the whole EU domain, where D2's `lpi_max` sees only 43.18–58.08 °N / −3.94–20.34 °E, plus the convection scheme's own CAPE beside the diagnostic `cape_ml`. **Data availability only — no grader consumes either.** Cloud-diag cache key bumped `V2 → V3` so warm caches re-fetch the wider blob. `lpi_con_max` is an interval MAX, so it is classified in `NWP_CLOUD_DIAG_RATE_SCALARS` (covering-interval hold), not as an instantaneous scalar.
  - ⚠️ Their cfgrib shortNames could **not** be verified against a live message from this environment (no feed/eccodes access). DWD does rename things — `RAIN_CON` arrives as `crr` — so the product names plus one near variant are mapped, with no bare `cape`/`lpi` fallback that could collide with `cape_ml`. If DWD delivers them under another name the fields stay `None` (ordinary missing-data semantics); nothing else regresses. Worth confirming against a real message before anything grades on them.
- **`tke` — opt-in, off.** Available on both variants, fetched only when `WB_ICON_TKE` names one (`icon-d2`, `icon-d2,icon-eu`, or `all`). No TKE consumer exists and ICON-EU's is ~903 KB per level-hour (~36 MB per forecast hour). Deliberately outside `model_level_variables` so a failed optional download cannot trip the #478 "incomplete column → skip the hour" guard on the sounding.
- **Doc correction:** the ICON-EU variable reference listed `QR`/`QS` as "Available" on ICON-EU. They are not (verified 2026-07-30) — corrected, with `QG` and `TKE` added.

Out of scope as specified: no `IcingZone.sld_risk` / `sounding.sld_zones` advisory wiring, no `sdi_2` / `tcond10_mx`.

## Issue linkage

Closes #530

## Testing / verification

`4783 passed, 18 skipped` — full suite, Python 3.12. (`test_auth.py::TestLoginRedirect::test_google_login_redirects` fails on a clean checkout of `main` in this container too — no Google OAuth config — and is deselected.)

Two new files, 76 tests:

- **`tests/test_icon_hydrometeors.py`** — per-variant lists (D2 gets the three, EU must not); every fetched variable has a decode mapping; interpolation bounds clamp the species non-negative; **download-volume regression** (measured 00z/step-003 file sizes × 50 levels × 12 h stays under 25 MB, plus a guard that the estimate prices the *actual* added list); `tke` off by default and never in the required decode set; `lpi_con_max`/`cape_con` decode + `−999.9` sentinel drop + rate-scalar classification; and one test per carrier — merge, **full sounding replacement**, time-axis forward-fill, time-axis interp, spatial interp — including that spatial interp never invents a one-sided species.
- **`tests/test_supercooled_rain.py`** — threshold boundaries (trace rejected, freezing drizzle admitted, exactly 0 °C not supercooled), missing inputs read as "not detected" rather than "none here", enrichment on a constructed profile, zone roll-up, and explicit pins that the flag moves **neither** `risk` **nor** `is_hazardous`.
- **`tests/test_precipitation.py`** — snow / rain / mixed on the new partition, graupel counted as frozen, a single missing species treated as zero, `None` for both "nothing reported" and "nothing falling", the full three-tier fallback chain, and that the zone mean is graded on the same tier as its levels.

Guardrails that fired and were updated rather than worked around: the `_interp_levels_at` completeness audit (#485), the cloud-diag cache-key version pins, and the ICON-D2 per-level tests that were asserting against ICON-EU's variable list.

Not verifiable here: no network access to `opendata.dwd.de` and no `cfgrib`, so nothing was run against real GRIB. A D2-enriched briefing producing non-null `rain_water_kg_kg` at precipitating levels still needs a live check.

---
_Generated by [Claude Code](https://claude.ai/code/session_01RSNLq54a9w9nCSAdLpuQjF)_


roznet:
Merged. The one Minor from the review — `_interpolate_cloud_water` not carrying `rain_water_g_kg`/`supercooled_rain` onto interpolated levels — is addressed directly on main in `817d9c7b`.

Added as its **own pass with its own bracket index** rather than folding qr into the existing one: sharing the index would let a rain-only level become a bracket endpoint for cloud water and, having no CLW of its own, suppress an interpolation that used to happen — a silent behaviour change shipped under the banner of a dormant fix. The CLW/ICMR pass is otherwise bit-identical; only the bracket search is extracted to `_bracketing_pressures` and shared.

`supercooled_rain` is **re-derived** from the interpolated qr against the level's own temperature, never interpolated — it's a threshold test on two quantities, so blending the neighbours' booleans would invent freezing rain at a level above zero. A warm nose between two freezing layers is exactly the freezing-rain profile, so that's the case pinned by test.

7 new tests in `tests/test_supercooled_rain.py::TestRainInterpolation`, including the no-extrapolation guard and the rain-only-level-doesn't-narrow-the-CLW-bracket case.

PR #534, opened 2026-08-04
Fix CAT risk calibration: altitude-dependent thresholds & mixed-layer suppression

### LINKED ISSUE #533: Richardson CAT diagnostic has no altitude floor: severe CAT flagged at 2,500 ft in the daytime boundary layer
Pilot debrief on flight `egny_egkb-2026-08-03-0d8b` (dep 2026-08-03 15:00Z, cruise 2,500 ft), `flown` / `{"TURB": "better"}`:

> For the last week, ICON has said that there will be severe CAT at 2,500 feet. This is clearly never going to happen, and in fact, CAT is only modelled from 5,000 feet up. So something has gone wrong.

Confirmed and root-caused. The advisory was RED "Severe CAT" on **all 7 packs** for this flight back to 2026-07-29, on an anticyclonic CAVOK August afternoon, while GFS/ECMWF mostly weren't:

| pack | gfs | ecmwf | icon |
|---|---|---|---|
| 07-29 | green | green | **red 35%** |
| 08-01 | amber | green | **red 47%** |
| 08-02 10Z | amber | green | **red 29%** |
| 08-02 23Z | amber | green | **red 41%** |
| 08-03 05Z | amber | amber | **red 35%** |
| 08-03 12Z | amber | red | **red 29%** |

## The premise that broke

**The Richardson CAT diagnostic has no altitude floor at all.** The reporter's expectation that "CAT is only modelled from 5,000 ft up" is not implemented anywhere — not in `_build_cat_layers`, not in `TurbulenceEvaluator`, not in the frontend `cat-bands` layer. The only low-level guard is `analysis/sounding/vertical_motion.py:207`: skip *negative* Ri at the two lowest levels. Small-**positive** Ri at 2,500 ft passes straight through and becomes "Severe CAT".

## What was actually severe in ICON

Not instability — **shear**. At the layer containing 2,500 ft, ICON's N² is normal and comparable to the other models; its S² is 5–40x larger. At route point 3 (30 nm), 950→925 hPa:

- **ICON**: 087/20.4 → 106/11.0 kt — speed nearly halves plus a 19° veer over 750 ft (~10 kt/1000 ft)
- **GFS**: 095/13.2 → 102/12.9 kt — essentially nothing (~0.5 kt/1000 ft)
- **ECMWF**: 082/19.5 → 092/16.1 kt — mild

ICON is resolving a sharp low-level decoupling in the anticyclonic easterly that GFS smooths away entirely. ICON is most likely *right* about the shear existing — it is the higher-resolution model. The defect is in what we do with it, not in the data.

## Findings

1. **Ri thresholds calibrated for the upper troposphere are applied unchanged in the boundary layer.** `vertical_motion.py:27–34` loosens SEVERE from the classical 0.25 to **0.5**, justified by "model vertical resolution (25–50 hPa between levels) is too coarse to resolve the thin shear layers (100–300 m)". But at 950–900 hPa, 25 hPa ≈ 230 m — already inside that band. The correction is needed at FL300 (25 hPa ≈ 800 m), not at 2,500 ft. Down low it inflates every tier by one.

2. **Help text and code disagree by 2x.** `web/ts/data/metrics-catalog.json:649` tells the pilot "Below about 0.25 expect clear-air turbulence". The code fires SEVERE at Ri < 0.5.

3. **The meteorology-decisions §8(c) boundary-layer guard misses the case it was written for.** Its stated rationale — "the routine daytime superadiabatic surface layer (thermals) — flagging it as CAT would paint summer-afternoon noise at the bottom of every cross-section" — is exactly this scenario. But the guard is negative-Ri-only and `idx <= 1`. On this day the mixed layer is 3,000–4,000 ft deep (lapse rate 2.99–3.06 °C/1000 ft at points 9–13, i.e. *exactly* dry-adiabatic), spanning levels 0–5, with small-positive Ri. It sails past the guard.

4. **One severe layer at one point forces RED.** `analysis/advisories/turbulence.py:119–120,171` — `has_severe` anywhere in the column → RED regardless of route percentage. A single BL shear layer at 1 of 17 points reds the whole advisory, which is why it never cleared across the week.

The digest noticed and hedged around it: *"At 2,500 ft this is low-level CAT, not mountain-wave or convective turbulence"* and *"some mechanical/shear-induced roughness at low level is plausible even if severe CAT does not materialise."*

## Measured effect of candidate fixes

Re-running `analyze_sounding` over the pack's own data, all 17 route points at 15Z:

| variant | icon SEVERE @cruise | gfs | ecmwf |
|---|---|---|---|
| current (0.5/1.0/2.0) | **8/17** | 3/17 | 0/17 |
| classical (0.25/0.5/1.0) | 2/17 | 2/17 | 0/17 |
| classical + skip well-mixed BL layers (lapse ≥ 2.7 °C/1000 ft) | 0 @cruise (3/17 in column) | 0/17 | 0/17 |

Classical thresholds alone collapse the ICON-vs-GFS disagreement (8→2 vs 3→2). That is the strongest signal that the 0.5 threshold is the primary culprit rather than ICON's data.

## Proposed fix

1. **Altitude-dependent Ri thresholds** — classical 0.25/0.5/1.0 below ~10,000 ft where model layer thickness is already at the shear-layer scale; keep the loosened 0.5/1.0/2.0 above, where the resolution argument genuinely holds.
2. **Extend the §8(c) guard to positive Ri inside the well-mixed layer** — detect the mixed-layer top (lapse rate at/near dry-adiabatic from the surface) and suppress CAT layers below it, rather than the current fixed `idx <= 1`.
3. **Reconsider severe-anywhere → RED for boundary-layer layers** — at minimum, a BL-origin severe layer should not bypass the route-percentage gate.

Preferring (1)+(2) over a hard 5,000 ft floor: a floor would also blind us to genuine low-level wind shear on a frontal day, which is a real GA hazard we want to keep flagging.

Follow-ups: update `metrics-catalog.json` so the documented threshold matches the code, and add a `meteorology-decisions.md` entry recording the altitude-dependent calibration.

Investigation only — no code changed.

roznet:
## What & why

Fixes three separate issues causing false "severe CAT" reports at low altitudes on stable summer afternoons (e.g., EGNY→EGKB case where ICON claimed severe CAT at 2,500 ft for a week straight while GFS/ECMWF were green).

### (a) Richardson number thresholds are now altitude-dependent

The original flat 0.5/1.0/2.0 calibration was justified by "model vertical resolution (25–50 hPa) is too coarse to resolve thin shear layers (100–300 m)". That argument is sound **and altitude-dependent**: 25 hPa spans ~230 m at 950 hPa but ~800 m at 300 hPa. Down low, model levels are already at the shear-layer scale; the correction there simply inflated every tier by one.

Reverts to classical Miles-Howard tiers (0.25/0.5/1.0) and applies a multiplier that is 1.0 at/below 10,000 ft, ramps linearly, and reaches 2.0 at/above 20,000 ft — reproducing the old calibration exactly where the resolution argument holds. A ramp rather than a step prevents adjacent levels either side of a fixed altitude from classifying the same Ri differently.

### (b) Surface well-mixed layer is now suppressed entirely

The §8(c) guard excluded *negative* Ri at the two lowest levels (daytime superadiabatic surface layer). It missed the EGNY→EGKB case: the mixed layer was 3,000–4,000 ft deep (lapse rate exactly dry-adiabatic) with small-**positive** Ri.

`mixed_layer_top_index()` now walks up from the surface while each layer's lapse rate is ≥ 8.8 °C/km (≈ dry-adiabatic), capped at 10,000 ft deep. Every level inside that layer is excluded from CAT **regardless of Ri sign** — shear there is convective-BL roughness, not KH instability. The top is published as `mixed_layer_top_ft` so suppression is inspectable rather than silent.

Suppression stops at the mixed-layer top (no hard 5,000 ft floor) to preserve genuine low-level wind shear on frontal days.

### (c) Boundary-layer severe no longer bypasses the route-percentage gate

`TurbulenceEvaluator` previously graded RED on `has_severe` anywhere at cruise, which is why a single BL shear layer at 1 of 17 points reddened the advisory for a week. That bypass is written for free-atmosphere severe CAT, a route-wide hazard; low-level shear is local and short-lived.

CAT layers lying wholly within the boundary layer are tagged `boundary_layer`. A severe one at cruise is floored at AMBER and graded by route percentage; free-atmosphere severe keeps the bypass unchanged.

## Issue linkage

Closes #533

## Testing / verification

- Added 13 new unit tests covering altitude-dependent thresholds, mixed-layer detection, and boundary-layer tagging.
- Updated 11 existing CAT-layer tests: their Ri values are halved so each test keeps asserting the same severity structure it was written for (e.g., a test that expected SEVERE at Ri=0.5 now uses Ri=0.25).
- Verified against the EGNY→EGKB pack: re-running `analyze_sounding` over the pack's own data shows ICON's SEVERE count at cruise collapses from 8/17 to 0 with these changes, while GFS/ECMWF remain unchanged.
- Existing tests pass; no regressions in upper-level CAT detection (the ×2 ramp leaves FL300 jet-flank cases untouched).

https://claude.ai/code/session_01TA6hzwrWiFRudXeByg1zH4


roznet:
Review round 1 processed. Branch rebased onto `main` (1 commit behind, clean) and pushed as `8afe43b`.

## Critical — `bl_top_ft` max() vs documented either/or

**Correct catch, resolved the other way round: the code was right, the prose was wrong.**

The contradiction was real — `max(ml_top_ft, surface + 5000)` in code against an either/or in both the inline comment and §24(c). But the max() is the intended behaviour, and the review's own framing points at why:

- The two readings **only diverge when a mixed layer is actually detected**, i.e. a convective daytime BL. A frontal low-level-wind-shear day — the hazard the review is protecting — has no well-mixed surface layer, so `ml_top_ft` is `None` and the AGL fallback applies. Both readings agree there.
- Where they do diverge, a shear sheet between the mixed-layer top and 5,000 ft AGL is most often that layer's own entrainment/capping shear, so the boundary-layer tag is the right reading.
- Using only the detected top would re-open #533 from the other side: with a shallow mixed layer (say 1,200 ft), a severe sheet at 2,000 ft would read as free atmosphere and RED the whole route off one point of seventeen.

One correction to the finding's consequence: the tag is **not** a mute, which is what makes the generous ceiling safe. A BL-tagged severe layer is still floored at AMBER, still REDs past the route-percentage threshold, and still paints its band in the cross-section. It loses the severe-anywhere bypass, not its voice.

So: the inline comment and §24(c) now state max() explicitly and record this reasoning, rather than the code changing to match prose that was the imprecise half.

**The coverage gap was real and is closed.** `test_shear_above_mixed_layer_top_still_flagged` did only assert `risk`/`base_ft`, as noted. Two tests now pin the semantics directly — `test_bl_ceiling_is_the_higher_of_mixed_layer_top_and_agl_fallback` (3,900 ft sheet above a 2,500 ft mixed layer → `boundary_layer is True`) and `test_layer_above_the_agl_fallback_is_free_atmosphere` (6,200 ft → `False`, keeps the RED bypass).

## Minor — duplicate `mixed_layer_top_index` walk

Fixed. Resolved once in `assess_vertical_motion` and passed into `_build_cat_layers`; the function signature now takes `ml_top_idx` explicitly. Only one caller, verified.

## Minor — iOS Swift DTO parity — deferred, not fixed

`RouteAnalysesResponse.swift:117` is a deliberate minimal subset that predates this PR: it already omits `richardson_number`, `base_pressure_hpa`, `e_shear_layers` and `convective_contamination`. No iOS code branches on CAT severity today, so mirroring two fields nothing consumes would be scope creep on a calibration fix. Decoding is unaffected. Worth a `/sync-ios-web` pass if and when iOS grows a cross-section that distinguishes boundary-layer from free-atmosphere shear.

## Verification

Full suite: 4820 passed. The one failure, `test_auth.py::TestLoginRedirect::test_google_login_redirects`, reproduces on a clean tree — pre-existing and unrelated.

---

roznet:
Review round 2 processed. No push — nothing in this round is a correctness blocker, and the round itself confirms the core logic traces clean. Both checks green (`claude-review`, GitGuardian).

## Important — E-Shear sibling module

**Checked. It does not need this fix, and the reason is structural rather than a threshold coincidence.**

The Ri false-positive has a specific engine: `Ri = N²/S²`, and inside a well-mixed layer N² → 0 *by definition* — that is what "well mixed" means. So Ri collapses toward zero for **any** non-zero shear, however weak, and the tier table then reads that as severity. The mixed layer manufactures the signal.

E-Shear has no stability term at all — `E = (5·HWS + VWS² + 42)/4` is absolute shear magnitude. There is no vanishing denominator, so weak shear cannot be amplified into severity. It is immune to this mechanism by construction, not by calibration.

Confirmed numerically against the exact profile from #533 (ICON, 950→925 hPa, 087/20.4 → 106/11.0 kt — the shear sheet that produced the week of RED):

| diagnostic | verdict on that sheet |
|---|---|
| Richardson (before this PR) | **SEVERE** |
| Richardson (after this PR) | suppressed — inside the mixed layer |
| **E-Shear (unchanged)** | **LIGHT** |

That profile is VWS ≈ 14.2 kt/1000 ft → E ≈ 61. E-Shear needs VWS ≈ **24.5 kt/1000 ft** to reach SEVERE with no horizontal shear — a genuine low-level jet, not summer-afternoon mixing. Reaching SEVERE at the observed VWS would additionally require HWS ≈ 80 kt/100 nm, a frontal-scale horizontal gradient absent on an anticyclonic CAVOK day.

Two further points on exposure:

- **E-Shear feeds no advisory grade.** `e_shear_layers` is written in `tasks/analyze.py:441` and read only by `data-extract.ts` / `airport-profile-adapter.ts`. Nothing under `analysis/advisories/` references it — verified. So the "one point of seventeen REDs the route for a week" mechanism cannot occur through E-Shear at all; the review's concern there is limited to a rendered band.
- The shared code is the *grouping* helpers (`_group_e_shear_levels` / `_split_e_shear_by_severity`), which are adjacency and severity-splitting only. The defect was in classification and suppression, upstream of both.

Applying mixed-layer suppression to E-Shear would arguably be a regression: a 25 kt/1000 ft shear inside the mixed layer is a real hazard, and E-Shear — unlike Ri — only fires there when the shear is genuinely that strong. I'll record this in §24 rather than change E-Shear.

## Minor — altitude ramp is model-uniform

Fair, and the honest answer is **no, UKMO and Météo-France were not spot-checked** — the ramp was derived against ICON vs GFS/ECMWF on the one route. I won't claim otherwise.

The tradeoff is worth stating precisely though: for a coarser model the classical tiers *under*-detect (thicker layers → Ri biased higher → fewer flags), so a model-uniform ramp errs toward missing rather than toward the false alarms this PR removes. That makes it a safe default but a real refinement — making the ramp a function of each model's actual pressure-level spacing rather than altitude alone is a design change beyond a calibration fix. Follow-up, not a blocker.

## Minor — iOS Swift DTO parity (raised in both rounds)

Position unchanged from round 1: `RouteAnalysesResponse.swift:117` is a deliberate minimal subset predating this PR (it already omits `richardson_number`, `base_pressure_hpa`, `e_shear_layers`, `convective_contamination`), no iOS code branches on CAT severity, and decoding is unaffected. Tracking rather than fixing here.

## Deferred to a direct-to-main commit

Both doc items below are worth recording; neither justifies another full review round on a green PR:

1. §24 — the E-Shear analysis above, as an explicit "sibling module checked, not affected, here's why".
2. §24 "Real-world validation needed" — that the ramp is calibrated against ICON/GFS/ECMWF only, with UKMO/Météo-France unverified.

---

roznet:
Post-review round processed (independent re-check of the implementation against the saved EGNY→EGKB pack). Pushed as `543596cf`.

## Critical — CAT layer geometry (found by the external review, confirmed on the pack)

`richardson_number` on a level describes the layer **below** it, but `_build_single_cat_layer()` built layers from the flagged levels' own altitudes. Every CAT band was bottom-clipped by one layer, and a lone qualifying level emitted a **zero-thickness layer** at the shear layer's upper bound. Verified on the pack: ICON point 8 emitted severe `2,523–2,523 ft` for a shear layer physically spanning 1,765–2,523 ft — missing the 2,500 ft cruise by 23 ft. 3 of ICON's 4 severe layers were zero-thickness; point 3's moderate band started 40 ft above cruise. The bug pre-dates this PR (the old doubled tiers merged levels into thick bands often enough to conceal it); this PR's finer severity splitting exposed it.

**Consequence: the PR's headline verification ("8/17 → 0 severe at cruise, advisory clears") was substantially an artifact.** With corrected geometry alone, the pack re-REDs at 11/17 any-risk coverage — reproducing the #533 complaint from the other direction.

Fixed: layers span base-level → top level (base extension capped by the grouping's own 100 hPa adjacency criterion, so sparse columns can't paint monster bands), threshold scale read at the layer midpoint. Regression test pins the point-8 scenario.

## Important — mixed-layer detector replaced with θv parcel criterion

The per-layer lapse walk truncated at a single sub-8.8 °C/km kink: on this pack it read ~1,000 ft tops at ICON points 3–8 while adjacent points detected 7,400 ft. Now primary detection is the standard θv-parcel criterion (+0.5 K Holtslag excess, plain-math θv — MetPy scalar calls avoided for standalone's 56k-soundings/single-core path; verified within 18 mK of MetPy). The lapse walk survives only as the no-temperature fallback for older packs. Honest caveat, recorded in §24: on ICON's own profiles for points 3–8, θv *agrees* with the shallow top — ICON genuinely models a shallow layer capped by a sheared weakly-stable band there, which is why the gate change below matters as much as the detector.

## Important — RED-via-coverage now requires moderate-or-worse

`pct_above_threshold` counted light identically to severe, so light chop over half the route graded RED. AMBER still keys on any-risk coverage; the RED tier requires moderate+ over 50%. Severe handling unchanged (free-atmosphere bypass; BL amber-floor). Strong updrafts count as moderate-tier, preserving their old ability to RED.

## Honest pack outcome (replaces the §24(a) table)

| variant | icon | gfs | ecmwf |
|---|---|---|---|
| pre-#533 | **RED** ("severe CAT at 2,500 ft", 7 days) | amber | green |
| #534 as first written | green "smooth ride" (geometry artifact) | amber | green |
| this round | **AMBER** (mod+ 8/17, worst BL-severe) | AMBER (5/17 light) | GREEN |

AMBER-with-detail is the defensible read: ICON resolves real low-level shear the other models smooth away; the week of RED was the defect, not the shear. §24 rewritten accordingly — including the previously-deferred round-2 items (E-Shear sibling analysis, ramp-calibration caveat) and a new follow-up pointer: Δz-based threshold scaling now falls out nearly free of the geometry fix.

## Verification

- Full suite: **4838 passed, 9 skipped, 0 failed.**
- 11 new tests: point-8 zero-thickness regression, sparse-column guard, θv-vs-lapse kink behaviour, θv stable-surface/fallback cases, light-vs-moderate coverage gate, updated geometry expectations throughout.
- Pack re-run script results above generated from the PR code path (`assess_vertical_motion` over the pack's own derived levels, all 17 points × 3 models).

---

roznet:
Review round 4 processed. Pushed as `b841d783`.

## Important — waypoint advisory bypasses the gate: **accepted, fixed by qualification, not suppression**

Correct catch, and the best sibling-hunt of the four rounds — `_cat_turbulence_advisory` is user-facing (per-waypoint drill-down table via `renderAdvisoriesTable`) and had no knowledge of the `boundary_layer` tag. Two qualifiers on scope: it reads *post-suppression* layers, so mixed-layer suppression already reaches this surface; and point-level surfaces showing severity the route grade doesn't is established behaviour (§24 kept the cross-section ribbon on severe-anywhere deliberately). So the defect was the **wording**, not the existence of the advisory: "CAT turbulence SEVERE 1765–2523ft (low Richardson number)" for a BL-tagged layer is the exact #533 framing — and meteorologically wrong too, since capping-shear at the BL top isn't clear-air turbulence.

Fix: when every contributing moderate+ layer is BL-tagged, the reason reads `Low-level wind shear SEVERE 1765-2523ft (boundary layer, low Richardson number)`; one free-atmosphere layer in the mix keeps the CAT wording. The layer still shows — drill-down keeps full depth. Three tests pin the split (BL-only, free-atmosphere, mixed).

## Minor — digest pack-dump: fixed

`(BL)` suffix on boundary-layer-tagged CAT lines in `digest/text.py`, as suggested. (Confirmed the LLM prompt path is unaffected — `prompt_builder` uses the graded manifest.)

## Verification

`test_altitude_advisories.py` + `test_vertical_motion.py` + `test_evaluators.py` + `test_digest_assertions.py`: 189 passed, 0 failed against the branch.

---

PR #537, opened 2026-08-08
feat(flights): opt-in soonest-first ordering for upcoming flights (#536)

### LINKED ISSUE #536: Flight list ordering preference: soonest-first for upcoming flights (web + iOS + Siri)
## Problem

On the Flights page, saved flights are ordered newest-departure-first. For **upcoming** flights that means the flight furthest in the future sits at the top, and the flight you're about to fly is buried at the bottom of the Future section. A user reported this: they expect the next flight at the top.

Some users prefer the current behaviour, so this must be an **opt-in account preference, not a default change**.

Scope is deliberately narrow: **only the Future section reorders.** Recent and Past stay most-recent-first.

The ordering is `departure_time DESC` applied in two independent places:

- `src/weatherbrief/storage/flights.py:509` — `list_flights_with_role` orders desc; `api/flights.py:669-684` buckets into future/recent/past preserving that order.
- Web (`web/ts/managers/flights-ui.ts:274-288`) re-buckets by `section` but **preserves server order**, so it inherits whatever the server does.
- iOS (`app/flyfun-weather/flyfun-weather/Views/Flights/FlightListView.swift:739-741`) **ignores** server order and sorts desc itself, per section.

There is also an existing inconsistency to fix as part of this work — see Part 2.

---

## Preference contract

Stored in `app_prefs_json` (JSON blob — **no DB migration required**), exactly like `units_region`:

```
flight_order: "furthest_first" | "soonest_first"      default "furthest_first"
```

`"furthest_first"` reproduces today's behaviour exactly. Naming note: the reporter called these "recent first / chronological", but the sort key is *departure* time, not creation time — "recent" reads as creation order and would mislead. Label by what ends up at the top.

### Settings copy (use verbatim)

> **Upcoming flights**
> ○ **Furthest first** — the flight departing last is at the top. Newly added flights, usually the furthest ahead, appear first.
> ○ **Soonest first** — the flight departing next is at the top, and the list reads forward in time.
>
> *Applies to upcoming flights only. Past flights always stay most-recent-first, so with "Soonest first" the list runs forward through your upcoming flights, then backward through your past ones.*

That last sentence is required, not optional — it discloses that the reading direction reverses at the Future→Recent boundary, which is the one genuinely awkward consequence of this feature.

---

## Part 1 — Backend (`src/weatherbrief/api/preferences.py`, `api/flights.py`)

1. Add `flight_order: str` to `PreferencesResponse` (near `units_region`, ~line 85) and `flight_order: Literal["furthest_first", "soonest_first"] | None = None` to `PreferencesUpdate` (~line 131).
2. Parse it in `_parse_service_toggles` (~line 236): `"flight_order": data.get("flight_order", "furthest_first"),`.
3. Persist it in `update_preferences` (~line 390), following the `units_region` pattern.
4. Add `load_flight_order(db, user_id) -> str` as a sibling of `load_units_region` (~line 543).
5. In `list_all_flights` (`api/flights.py:628`), sort the **future** flights ascending by `departure_time` when the preference is `"soonest_first"`. The `other` list holds future + recent — sort only the future members, leave recent alone.

### :warning: Do NOT add a `Depends()` parameter to `list_all_flights`

This is the single most important instruction in this issue.

`api/agent.py:175` (ChatGPT connector) calls `flights_api.list_all_flights(...)` **directly as a plain Python function**, not through FastAPI DI, passing `response`/`past_limit`/`past_offset`/`user_id`/`db` explicitly. If you add:

```python
    flight_order: str = Depends(load_flight_order),   # ← WRONG
```

…the HTTP route still works, but the direct caller doesn't pass the argument, so Python uses the default — and that default is a `fastapi.params.Depends` **instance**, not a string. `flight_order == "soonest_first"` is then silently `False` forever. No exception, no type error, no static-checker warning (the annotation says `str`). The website would work while the ChatGPT connector quietly kept the old order.

**Instead, load it in the function body** — `db` and `user_id` are already in scope:

```python
order = load_flight_order(db, user_id)
```

No signature change, no callers to update. See `.claude/CLAUDE.md` → "Changing Function Signatures".

### Agent/MCP surfaces inherit this — intended

`api/agent.py` and `src/weatherbrief/mcp/server.py::list_flights` both walk `list_all_flights` output in order, so they pick up the preference for free. This is the desired outcome: order doesn't semantically matter to an LLM, but a chatbot printing the list verbatim will then match what the user sees in the app. No separate work needed.

---

## Part 2 — Make the future/past boundary duration-aware

Today the server and the web UI disagree about when a flight stops being "upcoming":

- Server `_classify_section` (`api/flights.py:544`): `flight.departure_time >= now`
- Web `isFlightPast` (`web/ts/utils.ts:186-193`): `start + durationHours > now`

The web version is correct and is what we want everywhere: a flight that departed 30 minutes ago on a 3-hour trip is still in progress, not past. Move the server to the same predicate.

Requirements:

- **`_compute_recent_section` must use the identical predicate.** It independently tests `f.departure_time < now` (`api/flights.py:577`). If only `_classify_section` moves, an airborne flight can be `future` by one function and a `recent` candidate by the other. Extract one shared helper (e.g. `_flight_has_ended(flight, now)`) and call it from both.
- **Guard zero/None duration.** `Flight.flight_duration_hours` defaults to `0.0` (`src/weatherbrief/models/storage.py:45`) and zero-duration flights are a real case (the web add-flight flow has a `confirmZeroDuration` dialog). Zero duration must behave exactly as today: past the instant it departs.
- **Expected side effect:** an in-progress flight moves Past → Future, so it leaves the paginated past list and joins the always-full future list, and `X-Past-Total` drops by one at that moment. This is correct, not a pagination bug. Under "soonest first" the in-progress flight lands at the very top of the list, which is where you want it while flying.
- **iOS duplicates this boundary twice** and must move too: `Models/API/FlightResponse.swift` `flightSection` (~line 98) and the `groupedFlights` legacy fallback (`FlightListView.swift:755-764`), both currently `dep >= now`. These only fire against a legacy server that omits `section`, but the comment at `FlightResponse.swift:95` explicitly promises they "mirror exactly" the server bucketing, so the promise must be kept.

---

## Part 3 — Web

**No change needed in `flights-ui.ts`** — it already preserves server order within each section, so the server-side sort is sufficient.

Settings UI only:

- `web/settings.html:111` — add a `<select id="input-flight-order">` alongside the Units select in the same `.form-row`. Neighbouring options in that block are plain hardcoded English, so no i18n keys are required.
- `web/ts/settings-main.ts:840` — populate the select from `prefs.flight_order` on load (mirror the `input-units-region` block).
- `web/ts/settings-main.ts:1702` — add `flight_order` to the `accountUpdate` object.
- `web/ts/adapters/preferences-adapter.ts` — add `flight_order: string` to `PreferencesResponse` (~line 51) and `flight_order?: string` to `PreferencesUpdate` (~line 84).

The flights page already fetches `/user/preferences` (`web/ts/flights-main.ts:723`), so no new round-trip is introduced — but nothing on the flights page needs to read it, since the server sorts.

---

## Part 4 — iOS list

`FlightListView.groupedFlights` (`FlightListView.swift:735-743`) currently sorts every section descending, overriding the server. Change so the **Future** group follows the preference while Recent and Past stay descending.

Read the preference from `AppState.userPreferences.preferences`.

### :warning: `PreferencesResponse` field must be Optional

`Models/API/PreferencesResponse.swift` is cached to `UserDefaults` and decoded at launch (`Services/UserPreferencesStore.swift:18-23`). Adding a **non-optional** field makes that cached decode fail, silently falling back to `.empty` — which visibly resets the user's push/email notification toggles until the next network refresh.

Add it exactly like the notify fields: `let flightOrder: String?` plus a computed accessor defaulting to `.furthestFirst`, and include it in the `.empty` initialiser.

Suggest a small `enum FlightOrder: String { case furthestFirst = "furthest_first", soonestFirst = "soonest_first" }` for use across the list, settings and Siri.

### iOS settings

Add a Picker in `Views/SettingsView.swift` (follow the "Briefing updates" Picker at line 53, which is server-backed, rather than the local-only "Auto-Download" one at line 86). Wire it through a new `updateFlightOrder(_:using:)` on `UserPreferencesStore` using the existing private `put(_:field:client:)` helper (line 70) — that already PUTs the single key and adopts the returned full prefs as the new cache.

Include a condensed version of the settings copy as footer text.

**Accepted limitation (decided, do not "fix"):** if the preference is changed on the web while iOS is offline, iOS keeps the stale order until its next prefs refresh. Flipping this setting is rare; the round-trip isn't worth engineering around.

---

## Part 5 — Siri / App Intents

`FlightResolver.orderedForSuggestions` (`AppIntents/FlightResolver.swift:33-42`) is the display list — it backs Shortcuts suggestions (`FlightEntityQuery.suggestedEntities`) and the overview intent. It currently hardcodes upcoming-soonest-first, which today disagrees with the list view. It should follow the preference.

Add an `order:` parameter rather than reaching for global state:

```swift
nonisolated static func orderedForSuggestions(
    _ flights: [FlightResponse],
    order: FlightOrder = .furthestFirst,
    now: Date = Date()
) -> [FlightResponse]
```

Only the `upcoming` half flips; the `past` half stays most-recent-first.

### :warning: `FlightResolver.nextFlight` must NOT follow the preference

`nextFlight` (line 25) answers "my next flight" — semantically the soonest departure, regardless of how the list is drawn. Making it follow a display preference would be a real bug ("my next flight" returning the furthest-away one).

### Isolation wrinkle

`orderedForSuggestions` is deliberately `nonisolated` and pure so the non-MainActor test target can call it, but `UserPreferencesStore` is `@MainActor`. Do **not** force an actor hop into `suggestedEntities()`. Add a `nonisolated` static accessor that decodes the same `cachedUserPreferences` `UserDefaults` key the store reads at launch, and pass the value in as the `order:` argument. This keeps the function pure and testable.

Note: Shortcuts caches suggested entities, so a preference flip may not reorder the Shortcuts picker until iOS refreshes it. In-app behaviour is immediate. Acceptable.

---

## Tests

- `tests/test_preferences.py` — round-trip the new key; assert the default is `"furthest_first"` when absent from `app_prefs_json`.
- `tests/test_flights_api_sections.py` — future section ascending under `"soonest_first"`, unchanged under default; **Recent and Past unchanged under both**; past pagination (`past_limit`/`past_offset`, `X-Past-Total`) unaffected by the preference.
- New/extended: an in-progress flight (departed, still within `flight_duration_hours`) classifies as `future`; a zero-duration flight classifies as `past` immediately after departure; `_classify_section` and `_compute_recent_section` agree for an in-progress flight.
- `app/flyfun-weather/flyfun-weatherTests/FlightGroupingTests.swift` — Future ordering under both preference values; Recent/Past descending in both.
- New Swift test for `FlightResolver.orderedForSuggestions(order:)` under both values, plus a regression test that `nextFlight` returns the soonest under both.

---

## Out of scope

- Past / Recent ordering — stays most-recent-first under all settings. Do not touch the `order_by(...desc())` in `storage/flights.py`; past pagination is offset-based over that order and reordering it produces duplicate and skipped rows across pages.
- `_compute_recent_section`'s internal `candidates.sort(..., reverse=True)` (`api/flights.py:580`) — that is *selection* logic for the 2-flight debrief nudge, not display order. Leave it (other than the shared-predicate change in Part 2).
- Changing the default for existing or new users.
- Any DB migration — this is a JSON key in `app_prefs_json`.

---

## File reference

| Area | Files |
|---|---|
| Backend prefs | `src/weatherbrief/api/preferences.py` (~85, ~131, ~236, ~390, ~543) |
| Backend list + boundary | `src/weatherbrief/api/flights.py` (528-591, 628-691) |
| Web settings | `web/settings.html:111`, `web/ts/settings-main.ts:840,1702`, `web/ts/adapters/preferences-adapter.ts:41,74` |
| iOS models/store | `Models/API/PreferencesResponse.swift`, `Services/UserPreferencesStore.swift`, `Models/API/FlightResponse.swift:~98` |
| iOS list + settings | `Views/Flights/FlightListView.swift:735-764`, `Views/SettingsView.swift` |
| Siri | `AppIntents/FlightResolver.swift:25-42`, `AppIntents/FlightEntityQuery.swift:35` |
| Tests | `tests/test_flights_api_sections.py`, `tests/test_preferences.py`, `flyfun-weatherTests/FlightGroupingTests.swift` |

Design docs worth reading first: `debrief.md` (owns the three-section list), `ios-app-intents.md` (resolver design).

roznet:
## What & why

Saved flights are ordered newest-departure-first, so in the **Future** section the flight furthest ahead sits at the top and the one you're about to fly is buried at the bottom. Some users prefer that, so this ships as an **opt-in account preference**, not a default change — and only the Future section reorders.

`flight_order` lives in `app_prefs_json` (**no migration**), default `"furthest_first"` = today's behaviour exactly.

**Part 1 — backend.** `flight_order` added to the preferences response/update, parsed, persisted, and read back via `load_flight_order`. `list_all_flights` sorts only the *future* members of the future+recent list ascending under `"soonest_first"`; Recent entries keep both their positions and their order (the two sections can interleave in the incoming `departure_time desc` order once an in-progress flight is Future, so the reorder permutes within a subset rather than partitioning).

The preference is loaded **in the function body, not via `Depends()`** — `api/agent.py` calls `list_all_flights` directly as a plain Python function, where a `Depends` default would arrive as a `fastapi.params.Depends` instance and silently compare unequal to every valid value. No signature change, no callers to update. Agent/MCP surfaces inherit the ordering for free, which is the intended outcome.

An unrecognized stored value falls back to the default via a shared `_normalize_flight_order`, used by both the GET response and `load_flight_order`, so what a client is told and what the list actually does can't diverge.

**Part 2 — duration-aware future/past boundary.** The server used `departure_time >= now` while the web used `start + duration > now`. The web version is correct and is now used everywhere: a flight that departed 30 minutes ago on a 3-hour trip is in progress, not past. `_classify_section` and `_compute_recent_section` both go through a single `_flight_has_ended(flight, now)` helper, so an airborne flight can't be `future` to one and a `recent` candidate to the other. Zero/None duration behaves exactly as before — past the instant it departs.

Expected side effect: an in-progress flight moves Past → Future, leaving the paginated past list for the always-full future list, so `X-Past-Total` drops by one at that moment. Correct, not a pagination bug. Under "soonest first" it lands at the very top, which is where you want it while flying.

**Part 3 — web.** Settings select + `PreferencesUpdate`/`PreferencesResponse` types. `flights-ui.ts` needed no change: it already preserves server order within each section.

**Part 4 — iOS.** `FlightListView.groupedFlights(_:order:now:)` applies the preference to Future only; Recent/Past stay descending. Both duplicated copies of the boundary (`FlightResponse.resolvedSection(now:)` and the legacy `groupedFlights` fallback) move to the shared duration-aware `hasEnded(now:)`, keeping the "mirrors the server exactly" promise. Settings gains a server-backed Picker wired through `UserPreferencesStore.updateFlightOrder(_:using:)`. `PreferencesResponse.flightOrder` is **Optional** — a non-optional field would fail the cached-prefs decode at launch and visibly reset the notification toggles.

**Part 5 — Siri / App Intents.** `FlightResolver.orderedForSuggestions(_:order:now:)` takes the order as a parameter rather than reaching for global state, so it stays `nonisolated` and pure for the test target; `FlightEntityQuery.suggestedEntities()` passes `UserPreferencesStore.cachedFlightOrder()`, a `nonisolated` accessor over the same `cachedUserPreferences` blob, avoiding a main-actor hop.

The settled rule, arrived at over two review rounds and now written into `ios-app-intents.md`:

> **Entity pickers follow the display preference; spoken "next/upcoming" answers stay chronological.**

So the Shortcuts picker follows `flight_order`, while `nextFlight` and `IntentDialogs.overviewSummary` do not — the latter because it truncates to five, where a furthest-first preference would read out the five most *distant* flights and omit the one departing tomorrow.

All three intent surfaces share one duration-aware `FlightResolver.isUpcoming(_:now:)` delegating to `hasEnded(now:)`, so they agree with the list and the server about when a flight stops being upcoming. Without that, an in-progress flight sat at the top of Future in the app while Siri reported no upcoming flights and `nextFlight` opened the *next scheduled* briefing instead of the one you were flying.

Design docs updated: `debrief.md` (section assignment + ordering), `ios-app-intents.md` (display ordering vs. "my next flight", and the shared upcoming boundary).

### Worth a reviewer's eye

A flight whose `departureTime` won't parse now counts as **upcoming** on the intent surfaces rather than past — a consequence of `hasEnded` returning `false` for an unparseable date. That's the right side to land on: `resolvedSection` and the list's legacy bucketing already treat such a flight as Future, so this closes that gap rather than opening a new one.

## Issue linkage

Closes #536

## Testing / verification

- `tests/test_preferences.py` — default is `"furthest_first"` when absent from `app_prefs_json`; round-trip; invalid value 422s; doesn't disturb other prefs; an unknown stored value is coerced identically by `load_flight_order` and the GET response.
- `tests/test_flights_api_sections.py` — future ascending under `"soonest_first"`, unchanged under the default; Recent and Past identical under both; past pagination (`past_limit`/`past_offset`/`X-Past-Total`) unaffected; in-progress flight leads under soonest-first. New `TestDurationAwareBoundary`: in-progress → `future`, zero-duration → past on departure, departure-exactly-now unchanged, and `_classify_section`/`_compute_recent_section` agree for an airborne flight.
- Full Python suite: **4826 passed**, 1 pre-existing unrelated failure (`test_auth.py::TestLoginRedirect::test_google_login_redirects`, which also fails on a clean tree in this environment).
- Web: `npx tsc --noEmit` clean apart from two pre-existing errors in `ts/eval/label-panel.ts`.
- Swift tests written — `FlightGroupingTests` (Future ordering under both values, Recent/Past descending under both, legacy-fallback ordering, in-progress/zero-duration bucketing, `hasEnded` boundary) and `AppIntentsResolverTests` (`orderedForSuggestions` under both values, `isUpcoming` duration-awareness, `nextFlight` preferring the in-progress flight and ignoring the preference, overview truncation staying chronological) — but **not compiled or run**: no Swift toolchain in this environment. **They need an Xcode run before merge.**

https://claude.ai/code/session_018ANhAoH9izJu5rdaYuvtDT

### DIFF
--- a/app/flyfun-weather/flyfun-weather/AppIntents/FlightEntityQuery.swift
+++ b/app/flyfun-weather/flyfun-weather/AppIntents/FlightEntityQuery.swift
@@ -32,10 +32,16 @@ struct FlightEntityQuery: EntityStringQuery {
     }
 
     /// Suggestions shown in the Shortcuts parameter picker — upcoming flights
-    /// first (soonest at top), then the rest most-recent-first.
+    /// first (in the user's chosen order), then the rest most-recent-first.
+    ///
+    /// The preference is read from the cached-prefs blob via a `nonisolated`
+    /// accessor rather than the `@MainActor` store, so this stays off the main
+    /// actor. Shortcuts caches suggested entities, so a preference flip may not
+    /// reorder its picker until iOS refreshes them; in-app order is immediate.
     func suggestedEntities() async throws -> [FlightEntity] {
         guard let flights = try? await Self.loadFlights() else { return [] }
-        return FlightResolver.orderedForSuggestions(flights).map(FlightEntity.init)
+        let order = UserPreferencesStore.cachedFlightOrder()
+        return FlightResolver.orderedForSuggestions(flights, order: order).map(FlightEntity.init)
     }
 
     /// Cache-first flight load, shared by every query path. Does *not* gate on
--- a/app/flyfun-weather/flyfun-weather/AppIntents/FlightResolver.swift
+++ b/app/flyfun-weather/flyfun-weather/AppIntents/FlightResolver.swift
@@ -32,10 +32,14 @@ enum FlightResolver {
             .min(by: { ($0.departureDate ?? .distantFuture) < ($1.departureDate ?? .distantFuture) })
     }
 
-    /// Upcoming flights in the user's chosen order, then past flights
-    /// most-recent-first — the order used for Shortcuts suggestions and for the
-    /// "overview" intent. This is the *display* list, so it follows the
-    /// `flight_order` preference and matches what the flight list shows.
+    /// Upcoming flights in the requested order, then past flights
+    /// most-recent-first — the order used for the Shortcuts entity picker.
+    ///
+    /// That picker is a *display* list, so its caller passes the user's
+    /// `flight_order` preference and it matches what the flight list shows. The
+    /// overview intent calls this too but pins `.soonestFirst`: it truncates to
+    /// five and is answering "what's coming up", so it must not read out the
+    /// five most distant flights.
     ///
     /// `order` is a parameter rather than a read of global state so the function
     /// stays pure and callable from the nonisolated test target; callers pass
--- a/app/flyfun-weather/flyfun-weather/AppIntents/FlightResolver.swift
+++ b/app/flyfun-weather/flyfun-weather/AppIntents/FlightResolver.swift
@@ -21,14 +21,33 @@ enum FlightResolver {
     // the nonisolated test target — and any actor context — can call them
     // synchronously. They touch no shared state, only their arguments.
 
-    /// The soonest upcoming flight (departure ≥ now), or nil if none upcoming.
+    /// Whether a flight still lies ahead of `now` — the one "upcoming" test for
+    /// every intent surface.
+    ///
+    /// Duration-aware, so it agrees with `FlightResponse.resolvedSection`,
+    /// `FlightListView.groupedFlights` and the server's `_flight_has_ended`
+    /// (#536): a flight that departed 30 minutes ago on a 3-hour trip is in
+    /// progress, and the app's list shows it at the top of Future. Gating these
+    /// surfaces on bare `departureDate >= now` instead would make Siri report no
+    /// upcoming flights while the list showed one — the exact disagreement the
+    /// shared boundary exists to prevent.
+    ///
+    /// A flight whose `departureTime` won't parse counts as upcoming, matching
+    /// what `resolvedSection` and the list's legacy bucketing do with it.
+    nonisolated static func isUpcoming(_ flight: FlightResponse, now: Date) -> Bool {
+        !flight.hasEnded(now: now)
+    }
+
+    /// The soonest upcoming flight, or nil if none upcoming. An in-progress
+    /// flight is the soonest of all — which is right: it's the flight whose
+    /// briefing you want when you ask for "my next flight" mid-trip.
     ///
     /// Deliberately does NOT follow the `flight_order` display preference: "my
     /// next flight" means the soonest departure however the list happens to be
     /// drawn, and returning the furthest-away one would be a real bug.
     nonisolated static func nextFlight(in flights: [FlightResponse], now: Date = Date()) -> FlightResponse? {
         flights
-            .filter { ($0.departureDate ?? .distantPast) >= now }
+            .filter { isUpcoming($0, now: now) }
             .min(by: { ($0.departureDate ?? .distantFuture) < ($1.departureDate ?? .distantFuture) })
     }
 
@@ -53,10 +72,10 @@ enum FlightResolver {
             ? { ($0.departureDate ?? .distantFuture) < ($1.departureDate ?? .distantFuture) }
             : { ($0.departureDate ?? .distantPast) > ($1.departureDate ?? .distantPast) }
         let upcoming = flights
-            .filter { ($0.departureDate ?? .distantPast) >= now }
+            .filter { isUpcoming($0, now: now) }
             .sorted(by: upcomingSorted)
         let past = flights
-            .filter { ($0.departureDate ?? .distantPast) < now }
+            .filter { !isUpcoming($0, now: now) }
             .sorted { ($0.departureDate ?? .distantPast) > ($1.departureDate ?? .distantPast) }
         return upcoming + past
     }
--- a/app/flyfun-weather/flyfun-weather/AppIntents/FlightResolver.swift
+++ b/app/flyfun-weather/flyfun-weather/AppIntents/FlightResolver.swift
@@ -22,18 +22,35 @@ enum FlightResolver {
     // synchronously. They touch no shared state, only their arguments.
 
     /// The soonest upcoming flight (departure ≥ now), or nil if none upcoming.
+    ///
+    /// Deliberately does NOT follow the `flight_order` display preference: "my
+    /// next flight" means the soonest departure however the list happens to be
+    /// drawn, and returning the furthest-away one would be a real bug.
     nonisolated static func nextFlight(in flights: [FlightResponse], now: Date = Date()) -> FlightResponse? {
         flights
             .filter { ($0.departureDate ?? .distantPast) >= now }
             .min(by: { ($0.departureDate ?? .distantFuture) < ($1.departureDate ?? .distantFuture) })
     }
 
-    /// Upcoming flights soonest-first, then past flights most-recent-first — the
-    /// order used for Shortcuts suggestions and for the "overview" intent.
-    nonisolated static func orderedForSuggestions(_ flights: [FlightResponse], now: Date = Date()) -> [FlightResponse] {
+    /// Upcoming flights in the user's chosen order, then past flights
+    /// most-recent-first — the order used for Shortcuts suggestions and for the
+    /// "overview" intent. This is the *display* list, so it follows the
+    /// `flight_order` preference and matches what the flight list shows.
+    ///
+    /// `order` is a parameter rather than a read of global state so the function
+    /// stays pure and callable from the nonisolated test target; callers pass
+    /// `UserPreferencesStore.cachedFlightOrder()`.
+    nonisolated static func orderedForSuggestions(
+        _ flights: [FlightResponse],
+        order: FlightOrder = .furthestFirst,
+        now: Date = Date()
+    ) -> [FlightResponse] {
+        let upcomingSorted: (FlightResponse, FlightResponse) -> Bool = order == .soonestFirst
+            ? { ($0.departureDate ?? .distantFuture) < ($1.departureDate ?? .distantFuture) }
+            : { ($0.departureDate ?? .distantPast) > ($1.departureDate ?? .distantPast) }
         let upcoming = flights
             .filter { ($0.departureDate ?? .distantPast) >= now }
-            .sorted { ($0.departureDate ?? .distantFuture) < ($1.departureDate ?? .distantFuture) }
+            .sorted(by: upcomingSorted)
         let past = flights
             .filter { ($0.departureDate ?? .distantPast) < now }
             .sorted { ($0.departureDate ?? .distantPast) > ($1.departureDate ?? .distantPast) }
--- a/app/flyfun-weather/flyfun-weather/AppIntents/FlightsOverviewIntent.swift
+++ b/app/flyfun-weather/flyfun-weather/AppIntents/FlightsOverviewIntent.swift
@@ -14,7 +14,8 @@ struct FlightsOverviewIntent: AppIntent {
         let repo = IntentSupport.makeRepository()
         do {
             let flights = try await repo.flights()
-            return .result(dialog: "\(IntentDialogs.overviewSummary(flights))")
+            let order = UserPreferencesStore.cachedFlightOrder()
+            return .result(dialog: "\(IntentDialogs.overviewSummary(flights, order: order))")
         } catch APIError.unauthorized {
             // Token expired and the silent refresh failed (Decision 4).
             return .result(dialog: "\(IntentSupport.signedOutSpokenLine)")
--- a/app/flyfun-weather/flyfun-weather/AppIntents/FlightsOverviewIntent.swift
+++ b/app/flyfun-weather/flyfun-weather/AppIntents/FlightsOverviewIntent.swift
@@ -14,8 +14,7 @@ struct FlightsOverviewIntent: AppIntent {
         let repo = IntentSupport.makeRepository()
         do {
             let flights = try await repo.flights()
-            let order = UserPreferencesStore.cachedFlightOrder()
-            return .result(dialog: "\(IntentDialogs.overviewSummary(flights, order: order))")
+            return .result(dialog: "\(IntentDialogs.overviewSummary(flights))")
         } catch APIError.unauthorized {
             // Token expired and the silent refresh failed (Decision 4).
             return .result(dialog: "\(IntentSupport.signedOutSpokenLine)")
--- a/app/flyfun-weather/flyfun-weather/AppIntents/IntentDialogs.swift
+++ b/app/flyfun-weather/flyfun-weather/AppIntents/IntentDialogs.swift
@@ -44,7 +44,7 @@ enum IntentDialogs {
     /// "next/upcoming" answers stay chronological.
     static func overviewSummary(_ flights: [FlightResponse], now: Date = Date()) -> String {
         let upcoming = FlightResolver.orderedForSuggestions(flights, order: .soonestFirst, now: now)
-            .filter { ($0.departureDate ?? .distantPast) >= now }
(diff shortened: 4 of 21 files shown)


roznet:
Both findings addressed in `68457c0`.

**Important — `overviewSummary` truncation.** Agreed, and taken as a blocker: the five-flight cut meant that under the default `"furthest_first"` the spoken overview would read out the five most *distant* flights and omit the one departing tomorrow — a regression for every user who never touches the setting, not just for those who opt in.

Fixed by pinning the overview to soonest-first, on exactly the precedent you cite: **entity pickers are display lists and follow the preference; spoken "next/upcoming" answers stay chronological**, same reason `FlightResolver.nextFlight` is preference-independent. `FlightEntityQuery.suggestedEntities()` still passes the user's preference through, so the Shortcuts picker matches the app's list. The principle is now written down at both `FlightResolver.orderedForSuggestions` and `IntentDialogs.overviewSummary`, and pinned by a new test asserting the truncation keeps the soonest five whatever the preference.

**Minor — GET response vs. `load_flight_order` divergence.** Batched in since the push was already paid for. Both now go through one `_normalize_flight_order` helper, so a hand-edited or future-version blob can't make the reported preference disagree with the ordering actually applied. `test_load_flight_order_ignores_unknown_stored_value` now asserts the GET response too. Left the `units_region` precedent alone — out of scope for this PR.

Python suite still green on the touched files (64 passed). The Swift changes remain uncompiled — no Swift toolchain in this environment — so the App Intents tests still need an Xcode run before merge.

---

roznet:
Confirmed and fixed in `b642b4f`. Good catch — this one was a divergence *this PR introduced*, not a pre-existing wart: before Part 2, `resolvedSection` also used `dep >= now`, so every surface agreed an in-progress flight was past. Moving the list and the server without moving the resolver is what split them.

The user-visible symptom was worse than ordering: with an in-progress flight as your only upcoming one, the list showed it at the top of Future while Siri said *"you have no upcoming flights"*, and `nextFlight` (which backs the no-parameter `OpenBriefingIntent`) skipped the flight you were actually on to open next week's briefing instead.

All three now go through one `FlightResolver.isUpcoming(_:now:)` delegating to `FlightResponse.hasEnded(now:)` — the same predicate as `resolvedSection`, `groupedFlights` and the server's `_flight_has_ended`.

One behaviour change worth naming explicitly: a flight whose `departureTime` won't parse now counts as **upcoming** rather than past. That's a consequence of `hasEnded` returning `false` for an unparseable date, and it's the right side to land on — `resolvedSection` and the list's legacy bucketing already treat such a flight as Future, so this closes that gap too rather than opening a new one.

Tests added: `isUpcoming` duration-awareness (including zero-duration), `nextFlight` returning the in-progress flight over the next scheduled one, the in-progress flight leading the suggestions list under both preference values, and the overview no longer reporting "no upcoming flights" for an airborne flight. `FlightResolver.upcoming(_:now:calendar:)` (the tier-2 candidate set) is left alone — it's deliberately calendar-day scoped, not a future/past boundary.

Design doc updated; it also still described the overview as preference-driven, which was stale after the round-1 fix.

This is round 2 of the review loop, so I'm handing back to @roznet rather than iterating further. Still outstanding: **the Swift changes have never been compiled** — there's no Swift toolchain in this environment — so the App Intents and grouping tests need an Xcode run before merge.

---

roznet:
Not acting on this one — the second filter isn't redundant, and dropping it would be a bug.

`orderedForSuggestions` returns `upcoming + past`, not just the upcoming prefix (`FlightResolver.swift:80`) — it's the Shortcuts entity picker's list, which deliberately includes history so you can pick a past flight. The `.filter { isUpcoming }` in `overviewSummary` is what strips that past suffix. Remove it and the spoken overview announces every past flight as upcoming: `"You have 12 upcoming flights: …"` for a user with one flight next week.

The existing `overview: none upcoming` test pins exactly this — a past-only list must yield `"You have no upcoming flights."`, which only holds because of that filter.

What's true is that the call site reads as though it were re-filtering an already-filtered list, which is presumably how it scanned as redundant. That's a comment-level clarity issue, not a code change, and it's the only thing outstanding from this round — no blockers. Recommend merging (`--rebase`), and I'm happy to fold a one-line clarifying comment into a direct commit on main afterwards, or drop it if you'd rather not churn the file.

Still outstanding from earlier rounds: the Swift changes have never been compiled — no Swift toolchain in this environment — so the App Intents and grouping tests need an Xcode run before merge.

---

roznet:
Merged, after an Xcode run to close the outstanding gate: the Swift tests that had never been compiled were built and run on iPhone 17 Pro — **362 passed, 0 failed** for the `flyfun-weatherTests` target. Confirmed the new suites actually executed (`FlightGrouping`, and the resolver's `isUpcoming`/`nextFlight`/ordering cases) rather than silently matching nothing.

The last Minor — the `isUpcoming` filter at the `overviewSummary` call site reading as redundant — is folded into main as a comment in `817d9c7b`, recording that `orderedForSuggestions` returns `upcoming + past` so the filter is what strips the past suffix. No behaviour change.

PR #538, opened 2026-08-08
Couple the prompt-cache TTL to its write multiplier, and surface cache_saving_usd

### LINKED ISSUE #535: Prompt cache costing: couple TTL to its write multiplier, and surface cache_saving_usd
Two follow-ups on the digest prompt-cache work (#524, briefer_v3 `09d02f02`). Both are about the *cost* side, not the caching mechanism, which is now confirmed working on prod.

## Context

Measured on prod 2026-08-04 (first full day on briefer_v3): all 15 post-deploy digests share a single 4152-token prefix across all three guidance presets and 11 users. A `balanced`-written entry was read back by a `conservative` call — the collapse works. Hit rate 66.7% on that thin day; replaying Aug 3's 44 arrivals under v3 gives 84.1%.

Break-even hit rate is `(w - 1) / (w - 0.1)` for write multiplier `w`: **52.6%** on the 1h tier (2x), **21.7%** on 5m (1.25x). Prod sits comfortably above the 1h threshold.

## A. Couple the cache TTL to its write multiplier

Today these are two independent knobs that must agree, and nothing enforces it:

- TTL is a per-call-site parameter — `_system_content(..., ttl: str = "1h")`, `src/weatherbrief/digest/llm_digest.py:163`, applied at `:207`.
- The price of a write is a cost-config field — `CostConfig.cache_write_multiplier = 2.0`, `src/weatherbrief/costs.py:23`, consumed in `compute_cost()` at `:101`.

Anthropic bills a cache write at 2x on the 1h tier and 1.25x on 5m. Change the TTL without changing the multiplier and **every digest mis-prices its writes by 60%**, silently — no test fails, no log line. `designs/cost-attribution-design.md:124` documents the hazard but nothing enforces it.

This is not hypothetical. PR #524's own closing advice was "if the hit rate lands materially below 84%, the 5-minute TTL becomes the better bet — a one-line change." That one-line change is exactly the trap. (For the record that advice is also wrong on the merits now: replaying real arrival gaps at 5m gives a 20.0% hit rate against a 21.7% break-even, i.e. it bills *more* than not caching at all.)

**The crux:** cost is computed well after the call, at `src/weatherbrief/api/packs.py:1425`, from a `briefing_usage` row that records `llm_cache_write_tokens` but **not** which TTL produced them. So the pricing path currently has no way to know what it should charge. Options, roughly in increasing cost:

1. Pin the TTL as a single named constant and derive the multiplier from it, with a guard test asserting the two agree. Cheapest; keeps the eval's legitimate 5m override as the only divergence.
2. Have `_system_content` surface the effective multiplier to the caller so the write price travels with the write.
3. Persist the TTL (or the effective multiplier) on `briefing_usage`. Most correct, needs a migration.

Option 1 is probably enough — the constraint is that a future TTL change must be *unable* to land without the multiplier following.

**Gotcha for whoever picks this up:** the active prod `cost_config` row carries no cache keys at all, so both multipliers currently come from the `CostConfig` dataclass defaults (2.0 / 0.1). Any fix must not assume the DB row has them.

**Acceptance:** changing the TTL without the matching multiplier fails a test.

## B. Expose `cache_saving_usd` on `CostBreakdown`

The admin cost page can show what a briefing cost but not what caching saved, so the benefit is invisible in the product even though we now have the data.

`CostBreakdown` (`src/weatherbrief/costs.py:50`) and `breakdown_to_dict()` (`:259`) both need the field. Definition: what the cached tokens *would* have cost at 1x, minus what they actually cost at their multipliers —

```
saving = (cache_read + cache_write) * rate - (cache_read * 0.1 + cache_write * 2.0) * rate
```

For scale, that came to ~$0.05/day at 15 digests and ~$0.33/day on Aug 3's 44.

`detail_json` is schemaless, so this is purely additive — no migration. Rendering side is `src/weatherbrief/api/admin.py:398` and the admin cost view in `web/ts/`.

**Acceptance:** historical ledger rows written before the field existed still render (absent, not zero-that-looks-real, and never a hard error).

roznet:
## What & why

Two cost-side follow-ups on the digest prompt-cache work.

### A. The TTL and its write multiplier can no longer disagree

Cost is computed long after the digest call, at `_charge_briefing_cost()`, from a `briefing_usage` row that records *how many* cache-write tokens there were but **not which TTL produced them**. So nothing downstream could detect a TTL change: switching 1h → 5m without moving `cache_write_multiplier` mis-priced every write by 60%, silently — no failing test, no log line.

Fixed by derivation rather than documentation:

```python
costs.DIGEST_CACHE_TTL = "1h"                              # the TTL the charging path writes with
costs.CACHE_WRITE_MULTIPLIERS = {"5m": 1.25, "1h": 2.0}    # the only price table
CostConfig.cache_write_multiplier = CACHE_WRITE_MULTIPLIERS[DIGEST_CACHE_TTL]
_system_content(..., ttl: str = DIGEST_CACHE_TTL)
```

- Move the constant and the write price follows automatically.
- An unpriced TTL (`"30m"`) raises `ValueError` at the call site instead of billing at the wrong tier — a new Anthropic tier has to be priced first.
- The eval keeps its legitimate `"5m"` override; it writes no ledger row.
- The rate-card row can still override the multiplier (the DB wins over the derived default), so `update_cost_config` logs a warning when a submitted value disagrees with the TTL — a warning, not a 422, since the operator may be modelling an announced price change. It *does* now 422 a non-numeric multiplier, which previously stored cleanly and only blew up later inside the charging path, where exceptions are swallowed.
- Per the issue's gotcha, nothing assumes the active `cost_config` row carries the cache keys — prod's doesn't, and both multipliers come from the dataclass defaults.

### B. `cache_saving_usd` on `CostBreakdown` and the program report

What the cached tokens would have cost at 1x, minus what they actually billed at their multipliers. Reporting-only: it is already inside `token_cost_usd` / `variable_token_usd`, never joins a subtotal, and is not floored — a pure cache write is a real 2x surcharge and reads negative.

`detail_json` is schemaless, so this is purely additive: **no migration**. Rows charged before the field existed have no key, and every reader keeps *absent* distinct from a measured `0.0` — the program report carries `None`, the per-user aggregate omits the key, and both frontends render nothing rather than "caching saved $0.00", which would be a lie about untracked history.

Surfaced as a "Cache saved" card + note sentence on the admin Cost tab, and a line under the bar plus a breakdown cell on the per-user cost page.

### Drive-by

`scripts/run_digest_eval.py` still called `_system_content` with the pre-`head`/`tail` signature and had been raising `TypeError` on its cached path since briefer_v3. It lives in `scripts/`, outside the import graph pytest walks, so nothing caught it. Fixed and covered with the LLM mocked out.

## Issue linkage

Closes #535

## Testing / verification

- `tests/test_costs.py::TestCacheTtlPricingCoupling` — the rate-card default is derived from `DIGEST_CACHE_TTL`, the price table is pinned against Anthropic's published premiums, the digest writes at the TTL the rate card prices, and an unpriced TTL raises. `test_llm_digest.py` keeps a literal `"1h"` assertion as a tripwire so a TTL change stays a deliberate act.
- `tests/test_costs.py::TestPromptCachePricing` — the saving reconciles against a real uncached run (not just its own formula), a pure write is negative, and no total moves.
- `tests/test_cost_report_api.py::TestCacheSavingReporting` — legacy-shaped ledger rows (no key) report absent rather than zero, mixed windows sum only the rows that carry it, negatives pass through, and the per-user aggregate behaves the same. `TestConfigCacheMultiplierGuard` covers the warn/422 paths.
- `tests/test_llm_digest.py::TestEvalCacheCallSite` — the eval's cached call site matches the signature and sends a 5m breakpoint.
- Full suite: 4828 passed. `tests/test_auth.py::TestLoginRedirect::test_google_login_redirects` fails identically on the unmodified base branch (missing OAuth env in this container) — unrelated.
- `npx tsc --noEmit` clean apart from two pre-existing errors in `ts/eval/label-panel.ts`; `build:admin` and `build:user-costs` both build.
- Design docs updated: `cost-attribution-design.md` (the hazard section now describes the enforcement + the new field's absent-vs-zero contract) and `digest.md`.

https://claude.ai/code/session_01VBhRxWApGZpSCkReqRdCkF

---
_Generated by [Claude Code](https://claude.ai/code/session_01VBhRxWApGZpSCkReqRdCkF)_

### DIFF
--- a/designs/cost-attribution-design.md
+++ b/designs/cost-attribution-design.md
@@ -122,12 +122,67 @@ no migration. The serialized `CostConfig` keys: `token_cost_per_1k_input`,
 
 The two cache multipliers scale `token_cost_per_1k_input` for Anthropic prompt
 caching: a cache write bills at 2x on the 1-hour TTL the digest uses (1.25x on
-the 5-minute tier), a cache read at 0.1x. They are **not** independent of the
-TTL chosen in `digest/llm_digest.py:_system_content` — change one without the
-other and every digest is mis-priced. Note the admin cost-config form in
-`web/ts/admin-cost-view.ts` enumerates its fields explicitly and does not yet
-render these two, though `update_cost_config` already accepts them (it derives
-its allowed keys from the `CostConfig` dataclass).
+the 5-minute tier), a cache read at 0.1x.
+
+### TTL ↔ write multiplier are coupled by derivation
+
+A cache write's price depends on the TTL it was written with, and **nothing
+downstream records which TTL produced a write** — cost is computed later, in
+`_charge_briefing_cost()`, from a `briefing_usage` row that has only the token
+count. So a TTL change that the multiplier doesn't follow mis-prices every
+digest's writes by 60%, silently: no test fails, no log line.
+
+The defence is derivation, not documentation (issue #535):
+
+```
+costs.DIGEST_CACHE_TTL = "1h"                 # the TTL the charging path writes with
+costs.CACHE_WRITE_MULTIPLIERS = {"5m": 1.25, "1h": 2.0}   # the only price table
+CostConfig.cache_write_multiplier = CACHE_WRITE_MULTIPLIERS[DIGEST_CACHE_TTL]
+_system_content(..., ttl: str = DIGEST_CACHE_TTL)         # single call site default
+```
+
+- Move `DIGEST_CACHE_TTL` and the write price follows automatically.
+- An unpriced TTL (`"30m"`) raises `ValueError` at the call site rather than
+  billing at the wrong tier — a new Anthropic tier has to be priced first.
+- `tests/test_costs.py::TestCacheTtlPricingCoupling` pins the table against
+  Anthropic's published premiums and asserts the digest writes at the TTL the
+  rate card prices. `test_llm_digest.py` keeps a literal `"1h"` assertion as a
+  tripwire so a TTL change is a deliberate act.
+- The eval (`scripts/run_digest_eval.py`) overrides to `"5m"` — legitimate only
+  because a burst never expires *and* an eval writes no ledger row.
+- For the record, 5m is not the cheaper bet in prod: replayed arrival gaps give
+  a 20.0% hit rate against a 21.7% break-even, i.e. it bills more than not
+  caching at all. Break-even is `(w - 1) / (w - 0.1)`.
+
+A rate-card row can still override `cache_write_multiplier` (the DB wins over
+the derived default). `update_cost_config` logs a warning when a submitted
+value disagrees with `DIGEST_CACHE_TTL` rather than rejecting it, since the
+operator may be modelling a price change. Note the active prod row carries no
+cache keys at all, so both multipliers come from the dataclass defaults.
+
+Note the admin cost-config form in `web/ts/admin-cost-view.ts` enumerates its
+fields explicitly and does not yet render these two, though
+`update_cost_config` already accepts them (it derives its allowed keys from the
+`CostConfig` dataclass).
+
+### `cache_saving_usd` — what caching bought
+
+`CostBreakdown.cache_saving_usd` (and `ProgramCostReport.cache_saving_usd`)
+report the cached tokens priced at 1x minus what they actually billed:
+
+```python
+saving = (cache_read + cache_write) * rate
+       - (cache_read * read_mult + cache_write * write_mult) * rate
+```
+
+Reporting only — it is *already inside* `token_cost_usd` / `variable_token_usd`
+and must never be added to a subtotal. Negative on a pure cache write (2x is a
+real surcharge); the value is not floored. `detail_json` is schemaless, so this
+was purely additive — **no migration**. Rows charged before the field existed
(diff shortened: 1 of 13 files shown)


roznet:
The last Minor is addressed on main in `306f6897`.

Deduped one level further down than the review called for. The two `signedUsd4` copies weren't textually identical — admin's non-negative branch called its local `usd4(n)`, the user page inlined `\$${n.toFixed(4)}` — and chasing that down showed `usd4` itself duplicated between `admin-cost-view.ts` and `cost-summary-main.ts`. Moving only `signedUsd4` would have left the same smell one level lower, so both now live in `web/ts/utils.ts` and all three cost pages import them.

Purely mechanical, and verified rather than assumed: `usd4(n)` is exactly `\$${n.toFixed(4)}`, and old-admin / old-user / new agree on 15 cases including `-0`, `NaN` and `±Infinity`.

`usd` (2dp) is deliberately left local — a dozen call sites plus inlined `.toFixed(2)` siblings make that a real refactor, not a dedup.

